PR #3, opened 2017-08-03
Add more debug info to openapi generator

gmarek:
@mbohlool PTAL

cc @sttts

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -328,6 +328,7 @@ func (g openAPITypeWriter) generateMembers(t *types.Type, required []string) ([]
 			required = append(required, name)
 		}
 		if err = g.generateProperty(&m, t); err != nil {
+			glog.Errorf("Error when generating: %v, %v\n", name, m)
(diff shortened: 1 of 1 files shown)


pkg/generators/openapi.go
```diff
@@ -328,6 +328,7 @@ func (g openAPITypeWriter) generateMembers(t *types.Type, required []string) ([]
 			required = append(required, name)
 		}
 		if err = g.generateProperty(&m, t); err != nil {
+			glog.Errorf("Error when generating: %v, %v\n", name, m)
```
sttts: isn't the error printed by the caller of `generateMembers`? Or could it? Printing and returning the same error always looks odd.

gmarek: Returned error doesn't contain information on for what it failed. We can add it there, but this was just simpler.

sttts: As it's only the generator, I am fine with this.

mbohlool:
LGTM.

PR #4, opened 2017-08-21
Improve performance of CloneSpec

mbohlool:
CloneSpec should check for conflicts first and clone only if there is a conflict and a need to rename the spec. This PR addresses that (as well as adding gitignore file).

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -209,58 +209,84 @@ func renameDefinition(s *spec.Swagger, old, new string) {
 	delete(s.Definitions, old)
 }
 
-// Copy paths and definitions from source to dest, rename definitions if needed.
+// MergeSpecsOrFailOnConflict is differ from MergeSpecs as it fails if there is
+// a definition conflict.
+func MergeSpecsOrFailOnConflict(dest, source *spec.Swagger) error {
+	return mergeSpecs(dest, source, false)
+}
+
+// MergeSpecs copies paths and definitions from source to dest, rename definitions if needed.
 // dest will be mutated, and source will not be changed.
 func MergeSpecs(dest, source *spec.Swagger) error {
-	sourceCopy, err := CloneSpec(source)
-	if err != nil {
-		return err
+	return mergeSpecs(dest, source, true)
+}
+
+func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts bool) (err error) {
+	// Check for model conflicts
+	conflicts := false
+	for k, v := range source.Definitions {
+		v2, found := dest.Definitions[k]
+		if found && !reflect.DeepEqual(v, v2) {
+			if !renameModelConflicts {
+				return fmt.Errorf("model name conflict in merging OpenAPI spec: %s", k)
+			}
+			conflicts = true
+			break
+		}
 	}
-	for k, v := range sourceCopy.Paths.Paths {
+
+	if conflicts {
+		source, err = CloneSpec(source)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -209,58 +209,84 @@ func renameDefinition(s *spec.Swagger, old, new string) {
 	delete(s.Definitions, old)
 }
 
-// Copy paths and definitions from source to dest, rename definitions if needed.
+// SafeMergeSpecs is differ from MergeSpecs as it assumers dest and source does not
+// have any name conflicts. In case of a conflict, merging will fail.
+func SafeMergeSpecs(dest, source *spec.Swagger) error {
+	return mergeSpecs(dest, source, false)
```
sttts: I would call this `unsafe`, not `safe`.

pkg/aggregator/aggregator.go
```diff
@@ -209,58 +209,84 @@ func renameDefinition(s *spec.Swagger, old, new string) {
...
+}
+
+func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts bool) (err error) {
+	// Check for model conflicts
+	conflicts := false
+	for k, v := range source.Definitions {
+		v2, found := dest.Definitions[k]
+		if found && !reflect.DeepEqual(v, v2) {
+			if !renameModelConflicts {
+				return fmt.Errorf("Model name conflict in merging OpenAPI spec: %s", k)
```
sttts: lower-case errors

pkg/aggregator/aggregator.go
```diff
@@ -209,58 +209,84 @@ func renameDefinition(s *spec.Swagger, old, new string) {
...
-			i := 2
-			newName := fmt.Sprintf("%s_v%d", k, i)
-			_, foundInSource := sourceCopy.Definitions[newName]
-			for usedNames[newName] || foundInSource {
-				i += 1
-				newName = fmt.Sprintf("%s_v%d", k, i)
-				_, foundInSource = sourceCopy.Definitions[newName]
+
+	if conflicts {
+		source, err = CloneSpec(source)
```
sttts: ic, so we do not clone without conflict :+1:

pkg/aggregator/aggregator.go
sttts: In fact, it's neither safe nor unsafe. In case of conflict we return an error, right? So let's just call it `MergeSpecsOrFailOnConflict`.

mbohlool: How about `MergeSpecsWithNoConflicts`

pkg/aggregator/aggregator.go
mbohlool: fixed.

mbohlool:
PTAL

pkg/aggregator/aggregator.go
sttts: it's not that there are no conflicts. Only the resolution is different (error).

pkg/aggregator/aggregator.go
```diff
@@ -209,58 +209,84 @@ func renameDefinition(s *spec.Swagger, old, new string) {
 	delete(s.Definitions, old)
 }
 
-// Copy paths and definitions from source to dest, rename definitions if needed.
+// SafeMergeSpecs is differ from MergeSpecs as it assumers dest and source does not
```
sttts: outdated godocs

sttts:
Godoc is out of date. Feel free to LGTM/merge after fixing that.

PR #5, opened 2017-08-25
Add MergeSpecsIgnorePathConflict

mbohlool:
MergeSpecsIgnorePathConflict is a new merge method that ignores path conflicts. It is useful if we want the destination has priority over source in serving paths.

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -209,19 +209,43 @@ func renameDefinition(s *spec.Swagger, old, new string) {
 	delete(s.Definitions, old)
 }
 
-// MergeSpecsOrFailOnConflict is differ from MergeSpecs as it fails if there is
+// MergeSpecsIgnorePathConflict is the same as MergeSpecs except it will ignore any path
+// conflicts by keeping the paths of destination. It will rename definition conflicts.
+func MergeSpecsIgnorePathConflict(dest, source *spec.Swagger) error {
+	return mergeSpecs(dest, source, true, true)
+}
+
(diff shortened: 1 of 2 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -209,19 +209,43 @@ func renameDefinition(s *spec.Swagger, old, new string) {
 	delete(s.Definitions, old)
 }
 
-// MergeSpecsOrFailOnConflict is differ from MergeSpecs as it fails if there is
+// MergeSpecsIgnorePathConflict is the same as MergeSpecs except it will ignore any path conflicts.
```
sttts: ignoring by what? by keeping the entry for dest or source?

pkg/aggregator/aggregator.go
```diff
@@ -209,19 +209,43 @@ func renameDefinition(s *spec.Swagger, old, new string) {
...
 
-func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts bool) (err error) {
+func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConflicts bool) (err error) {
+
+	specCloned := false
+	if ignorePathConflicts {
+		keepPaths := []string{}
+		for k := range source.Paths.Paths {
+			if _, found := dest.Paths.Paths[k]; !found {
+				keepPaths = append(keepPaths, k)
```
sttts: so dest wins, right? Should be in the comment above.

sttts:
Just a comment about godoc. After fixing that, lgtm.

pkg/aggregator/aggregator.go
mbohlool: fixed.

PR #8, opened 2017-08-29
Bugfix in MergeSpecsIgnorePathConflict

mbohlool:
We had a bug in MergeSpecsIgnorePathConflict that would fail aggregation if all paths were conflicting. This PR fixes that.

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -154,8 +154,28 @@ func (s *referenceWalker) Start() {
 	}
 }
 
-// FilterSpecByPaths remove unnecessary paths and unused definitions.
+// usedDefinitionForSpec returns a map with all used definition in the provided spec as keys and true as values.
+func usedDefinitionForSpec(sp *spec.Swagger) map[string]bool {
+	usedDefinitions := map[string]bool{}
+	walkOnAllReferences(func(ref spec.Ref) spec.Ref {
+		if refStr := ref.String(); refStr != "" && strings.HasPrefix(refStr, definitionPrefix) {
+			usedDefinitions[refStr[len(definitionPrefix):]] = true
+		}
+		return ref
+	}, sp)
+	return usedDefinitions
+}
+
+// FilterSpecByPaths removes unnecessary paths and definitions used by those paths.
+// i.e. if a Path removed by this function, all definition used by it and not used
+// anywhere else will also be removed.
 func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
+	// Walk all references to find all used definitions. This function
+	// want to only deal with unused definitions resulted from filtering paths.
+	// Thus a definition will be removed only if it has been used before but
+	// it is unused because of a path prune.
+	initialUsedDefinitions := usedDefinitionForSpec(sp)
+
 	// First remove unwanted paths
 	prefixes := util.NewTrie(keepPathPrefixes)
 	orgPaths := sp.Paths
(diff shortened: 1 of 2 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -156,6 +156,22 @@ func (s *referenceWalker) Start() {
...
 // FilterSpecByPaths remove unnecessary paths and unused definitions.
 func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
+	// Walk all references to find all unused definitions. This function
+	// want to only deal with unused definitions resulted from filtering paths.
+	initialUnusedDefinitions := map[string]bool{}
+	for k := range sp.Definitions {
+		initialUnusedDefinitions[k] = true
+	}
+	walkOnAllReferences(func(ref spec.Ref) spec.Ref {
+		if ref.String() != "" {
```
sttts: `if rs := ref.String(); rs != "" {` and then reuse this.

pkg/aggregator/aggregator.go
```diff
@@ -156,6 +156,22 @@ func (s *referenceWalker) Start() {
...
+			refStr := ref.String()
+			if strings.HasPrefix(refStr, definitionPrefix) {
+				delete(initialUnusedDefinitions, refStr[len(definitionPrefix):])
```
sttts: turn this around: initialUnusedDefinitions -> initiallyUsedDefinitions. Then you can get rid of the first loop.

pkg/aggregator/aggregator.go
```diff
@@ -156,6 +156,22 @@ func (s *referenceWalker) Start() {
 
 // FilterSpecByPaths remove unnecessary paths and unused definitions.
 func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
+	// Walk all references to find all unused definitions. This function
```
sttts: Shouldn't we remove all unused definitions first and then keep the algorithm as it was? Much simpler.

Also `FilterSpecByPaths` should be called `FilterSpecByPathsAndPrune`.

mbohlool: Actually I did this because we should NOT remove unused definitions. When we rename definitions in one release, that is, kind of, backward incompatible change. To make sure we would keep backward compatibility we add definitions with a deprecation description and reference them to the new name. This way all tooling can keep the backward compatibility for a release. These deprecation definitions, however, are unused in the Spec (All paths refer to the new names). If we don't do this here, all of those definitions will be removed from the spec. A good example is https://github.com/kubernetes/kubernetes/pull/51191.

pkg/aggregator/aggregator.go
mbohlool: fixed

pkg/aggregator/aggregator.go
mbohlool: Fixed.

pkg/aggregator/aggregator.go
```diff
@@ -154,8 +154,28 @@ func (s *referenceWalker) Start() {
 	}
 }
 
+// usedDefinitionForSpec returns a map with all used definition in the provided spec as keys and true as values.
+func usedDefinitionForSpec(sp *spec.Swagger) map[string]bool {
+	usedDefinitions := map[string]bool{}
+	walkOnAllReferences(func(ref spec.Ref) spec.Ref {
+		if refStr := ref.String(); refStr != "" {
+			if strings.HasPrefix(refStr, definitionPrefix) {
```
sttts: nit: merge both if clauses with &&

pkg/aggregator/aggregator.go
```diff
@@ -154,8 +154,28 @@ func (s *referenceWalker) Start() {
...
+			}
+		}
+		return ref
+	}, sp)
+	return usedDefinitions
+}
+
 // FilterSpecByPaths remove unnecessary paths and unused definitions.
 func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
+	// Walk all references to find all used definitions. This function
```
sttts: this is a strange semantics. I had to read this sentence multiple time to get it. So essentially we are garbage collecting definitions previously used by filtered-out paths? Can you write something like that in the godoc?

pkg/aggregator/aggregator.go
```diff
@@ -228,16 +238,22 @@ func MergeSpecs(dest, source *spec.Swagger) error {
...
-
 	specCloned := false
 	if ignorePathConflicts {
 		keepPaths := []string{}
+		hasConflictingPath := false
 		for k := range source.Paths.Paths {
 			if _, found := dest.Paths.Paths[k]; !found {
 				keepPaths = append(keepPaths, k)
+			} else {
+				hasConflictingPath = true
```
sttts: isn't `hasConflictingPath` and `len(keepPaths) != 0` equal? Don't understand why we need these two way to express that.

mbohlool: If there is no common paths, `len(keepPaths) != 0` but `hasConflictingPath` will be false.In the case of no conflicting path, we want to avoid an unnecessary Clone.

pkg/aggregator/aggregator.go
mbohlool: Done.

pkg/aggregator/aggregator.go
```diff
@@ -154,8 +154,28 @@ func (s *referenceWalker) Start() {
...
+	walkOnAllReferences(func(ref spec.Ref) spec.Ref {
+		if refStr := ref.String(); refStr != "" && strings.HasPrefix(refStr, definitionPrefix) {
+			usedDefinitions[refStr[len(definitionPrefix):]] = true
+		}
+		return ref
+	}, sp)
+	return usedDefinitions
+}
+
+// FilterSpecByPaths remove unnecessary paths and definitions used by those paths.
```
sttts: removes

pkg/aggregator/aggregator.go
sttts: misread. It's logically different. thanks.

sttts:
Lgtl. please squash.

PR #9, opened 2017-08-30
Move parsing code from kubernetes repo to this repo

mengqiy:
Comparing to https://github.com/kubernetes/kubernetes/tree/master/pkg/kubectl/cmd/util/openapi
- validation is not moved since it depends on https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/util/openapi/validation/BUILD#L40
- openapi_cache.go & openapi_cache_test.go will be deleted in https://github.com/kubernetes/kubernetes/pull/50404/files#diff-924015a54cbffd96c80a6551d2bd05ee
- openapi_getter.go & openapi_getter_test.go depends on client-go. Maybe should not be moved.


mengqiy:
`swagger.json` is so big. It is used in `pkg/util/parsing/openapi_test.go`.
I can truncate it to keep only the part we need for testing if you want.

mengqiy:
After discussing with @mbohlool and @caesarxuchao , we think the code in this PR should be moved into apimachinery repo.
Closing

mengqiy:
Updated the PR to no depend on any k8s repo.

mengqiy:
It seem this PR doesn't pick my new commits. Will open a new one.

PR #11, opened 2017-09-05
initialise schemaTypeFormatMap once at package level

jim-minter:
initialise schemaTypeFormatMap once at package level instead of every time GetOpenAPITypeFormat is called


mbohlool:
LGTM. Thanks.

PR #12, opened 2017-10-16
proto: Index by model name rather than GVK

apelisse:
Remove completely the notion of Group-version-kind from this package,
as we'll move it where it's actually relevant. This removes the
parsing/constructing which was kind of out of place.

Also allow to return the full list of models. The code in kubernetes
will get all these models, and then index by gvk.


apelisse:
First step of https://github.com/kubernetes/kubernetes/issues/51823

mengqiy:
/lgtm Thanks!
Waiting for @mbohlool's approval.

pkg/util/proto/document.go
```diff
@@ -58,23 +59,21 @@ func VendorExtensionToMap(e []*openapi_v2.NamedAny) map[string]interface{} {
 // Definitions is an implementation of `Resources`. It looks for
 // resources in an openapi Schema.
 type Definitions struct {
-	models    map[string]Schema
-	resources map[string]string
+	resources map[string]Schema
```
mbohlool: so my understanding is you removed "resources" and then renamed "models" to "resources". Why the later? I mean why not keep the name as models?

apelisse:
I don't have a strong opinion tbh. Either name works fine :-)

PR #14, opened 2017-10-21
Enable generation to a vendored output path

marun:
The PR adds a way to generate OpenAPI spec in an output path in the `vendor/` folder by using the argument to ``--output-package`` verbatim.  Previously, the argument to ``--output-package`` was canonicalized to an importable package name (e.g. ``vendor/k8s.io/kubernetes`` => ``k8s.io/kubernetes``) which would result in the file being generated outside of the vendoring repo (e.g. ``$GOPATH/src/k8s.io/kubernetes`` instead of ``$GOPATH/src/k8s.io/federation/vendor/k8s.io/kubernetes``). 

This is in support of generating openapi code for vendored kube by federation as per https://github.com/kubernetes/kubernetes/issues/54335.

cc: @mbohlool

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -118,17 +118,32 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 
 `)...)
 
-	if err := context.AddDir(arguments.OutputPackagePath); err != nil {
+	outputPath := arguments.OutputPackagePath
+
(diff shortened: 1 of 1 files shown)


mbohlool:
Can you add a little explanation to the PR? Is this correct:?

```
The PR adds a way to generate OpenAPI spec in an output path in `vendor/` folder.
```

marun:
@mbohlool Done.

PR #15, opened 2017-10-27
proto: Add model validation

apelisse:
This new code allows someone to validate that a given object (as a
`map[string]interface{}`) is valid against a specific model.

There are no test in this code, since the old test are depending on
apimachinery code that we can't import here. A nice follow-up would be
to add small unit-tests to validate each individual part.


mengqiy:
Already reviewed code.
LGTM

/assign @mbohlool

mbohlool:
Only checked imports for violation and went with "Already reviewed code" for the rest.

PR #16, opened 2017-10-30
Add util function to convert package/type path to OpenAPI name

roycaihw:
Add a function to convert the package/type path (e.g. parsed by gengo) into openapi name.

Examples:
Input:  k8s.io/api/core/v1.Pod
Output: io.k8s.api.core.v1.Pod

Input:  k8s.io/api/core/v1
Output: io.k8s.api.core.v1


roycaihw:
cc @mbohlool

PR #18, opened 2017-11-02
New OpenAPI endpoint

roycaihw:
OpenAPI endpoint that includes spec version in name, but hides all other details in the header of the request and response. 

The design doc and deprecation process are tracked at: https://docs.google.com/document/d/19lEqE9lc4yHJ3WJAJxS_G7TcORIJXGHyq3wpwcH28nU. 

The endpoints in [generic apiserver](https://github.com/kubernetes/kubernetes/blob/443908193d564736d02efdca4c9ba25caf1e96fb/staging/src/k8s.io/apiserver/pkg/server/routes/openapi.go#L35) and [aggregator](https://github.com/kubernetes/kubernetes/blob/443908193d564736d02efdca4c9ba25caf1e96fb/staging/src/k8s.io/kube-aggregator/pkg/controllers/openapi/aggregator.go#L112-L113) will be deprecated.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -22,18 +22,19 @@ import (
 	"crypto/sha512"
 	"encoding/json"
 	"fmt"
-	"gopkg.in/yaml.v2"
 	"mime"
 	"net/http"
 	"strings"
 	"sync"
 	"time"
 
+	yaml "gopkg.in/yaml.v2"
+
 	"github.com/NYTimes/gziphandler"
-	"github.com/emicklei/go-restful"
+	restful "github.com/emicklei/go-restful"
 	"github.com/go-openapi/spec"
 	"github.com/golang/protobuf/proto"
-	"github.com/googleapis/gnostic/OpenAPIv2"
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
 	"github.com/googleapis/gnostic/compiler"
 
 	"k8s.io/kube-openapi/pkg/builder"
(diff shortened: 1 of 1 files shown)


roycaihw:
cc @mbohlool

pkg/handler/handler.go
```diff
@@ -202,3 +211,53 @@ func toGzip(data []byte) []byte {
...
+// RegisterOpenAPIVersionedService registers a handler to provides access to provided swagger spec.
+func RegisterOpenAPIVersionedService(openapiSpec *spec.Swagger, servePath string, handler common.PathHandler) (*OpenAPIService, error) {
+	o := OpenAPIService{}
+	if err := o.UpdateSpec(openapiSpec); err != nil {
+		return nil, err
+	}
+
+	files := map[string]func() ([]byte, string, time.Time){
+		"application/json":                                           o.getSwaggerBytes,
+		"application/com.github.proto-openapi.spec.v2@v1.0+protobuf": o.getSwaggerPbBytes,
```
mbohlool: How the user show now this exact string for proto? we can document it but it would be nice if we can return list of supported types with the response, probably in the same Accept header. Also with that, we can support HEAD call too so a user just do the HEAD call and get the list and do the right call with the right type.

roycaihw: I'm afraid there is no standardized way to do that kind of negotiation in HTTP. The closest thing I guess is `HTTP 300 Multiple Choices`, but it uses response body, URL to distinguish types, and there is no standardized way of choosing one of the responses on client side. We should probably document it.

mbohlool: Documenting it looks good for now.

mbohlool:
@roycaihw Thanks for this PR. Please coordinate with kubectl team to use this endpoint instead of deprecated one. We need a plan to deprecate the old endpoints.

PR #21, opened 2017-11-11
generators: let types provide defs without importing kube-openapi

ericchiang:
This change lets types override their generated spec definition
with methods that don't import kube-openapi. Doing this will let
client-go remove its openapi imports, eliminating ~10 transitive
dependencies for Kubernetes clients.

To override the generated spec definition, a type can now implement
the following two methods:

	func (_ MicroTime) OpenAPISchemaType() []string {
		return []string{"string"}
	}

	func (_ MicroTime) OpenAPISchemaFormat() string {
		return "date-time"
	}

POC can be found here: https://github.com/ericchiang/kubernetes/commit/b0282c59f05df721e769fb9d34ab52fe9a8db4a5

cc @sttts @mbohlool

### DIFF
--- a/pkg/generators/README
+++ b/pkg/generators/README
@@ -11,5 +11,36 @@ escape or quote the value string. Extensions can be used to pass more informatio
 documentation generators. For example a type might have a friendly name to be displayed in documentation or
 being used in a client's fluent interface.
 
+# Custom OpenAPI type definitions
+
+Custom types which otherwise don't map directly to OpenAPI can override their
+OpenAPI definition by implementing a function named "OpenAPIDefinition" with
+the following signature:
+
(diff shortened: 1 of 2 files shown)


sttts:
👍  👍  :+1:

sttts:
Am not deep in the kube-openapi code. @mbohlool can you take a look?

apelisse:
/unassign

I have no knowledge of this part of the code. @mbohlool

mbohlool:
Look good to me, we probably should document this (and the other customization method) here: https://github.com/kubernetes/kube-openapi/tree/master/pkg/generators.

ericchiang:
@mbohlool thanks, docs added.

mbohlool:
thanks @ericchiang LGTM

PR #22, opened 2017-11-17
Aggregator now re-uses the model definition that it had renamed previously

luksa:
Before this fix, the aggregator would create multiple identical
copies (but with a different name) of the same object type.

This fixes https://github.com/kubernetes/kubernetes/issues/55941

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -291,15 +291,29 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 			from, to string
 		}
 		renames := []Rename{}
+
+	OUTERLOOP:
 		for k, v := range source.Definitions {
 			if usedNames[k] {
 				v2, found := dest.Definitions[k]
-				// Reuse model iff they are exactly the same.
+				// Reuse model if they are exactly the same.
 				if found && reflect.DeepEqual(v, v2) {
 					continue
 				}
-				i := 2
-				newName := fmt.Sprintf("%s_v%d", k, i)
+
+				// Reuse previously renamed model if one exists
+				var newName string
+				i := 1
+				for found {
+					i++
+					newName = fmt.Sprintf("%s_v%d", k, i)
+					v2, found = dest.Definitions[newName]
+					if reflect.DeepEqual(v, v2) {
+						renames = append(renames, Rename{from: k, to: newName})
+						continue OUTERLOOP
+					}
+				}
+
 				_, foundInSource := source.Definitions[newName]
 				for usedNames[newName] || foundInSource {
 					i++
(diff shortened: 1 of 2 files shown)


apelisse:
I have no clue about that

/assign @mbohlool 
/cc @cheftako

pmorie:
@cheftako, do you have some review bandwidth for this PR?

pkg/aggregator/aggregator.go
```diff
@@ -291,15 +291,29 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
 				if found && reflect.DeepEqual(v, v2) {
 					continue
 				}
-				i := 2
-				newName := fmt.Sprintf("%s_v%d", k, i)
+
+				// Reuse previously renamed model if one exists
+				var newName string
+				i := 1
+				for found {
```
cheftako: This is a forever loop correct? So if reflect.DeepEqual never evaluates to true, you will be stuck in this loop?

pkg/aggregator/aggregator.go
```diff
@@ -291,15 +291,29 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
-				// Reuse model iff they are exactly the same.
+				// Reuse model if they are exactly the same.
 				if found && reflect.DeepEqual(v, v2) {
 					continue
 				}
-				i := 2
-				newName := fmt.Sprintf("%s_v%d", k, i)
+
+				// Reuse previously renamed model if one exists
+				var newName string
```
mbohlool: nit: newName only used in the loop. I would use := on assignment instead of defining it here.

pkg/aggregator/aggregator.go
mbohlool: Ultimately we get to a point that newName will not be found in the destination and loop will stop.

pkg/aggregator/aggregator.go
```diff
@@ -291,15 +291,29 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
+				i := 1
+				for found {
+					i++
+					newName = fmt.Sprintf("%s_v%d", k, i)
+					v2, found = dest.Definitions[newName]
+					if reflect.DeepEqual(v, v2) {
```
mbohlool: better be `found && reflect.DeepEqual(v, v2)`

pkg/aggregator/aggregator.go
mbohlool: should you set initial value of found to `true` here?

pkg/aggregator/aggregator.go
luksa: No, `newName` is used further down in the next loop.

pkg/aggregator/aggregator.go
luksa: Fixed. Thanks.

pkg/aggregator/aggregator.go
luksa: It's always `true` anyway, because for it to be `false`, `usedNames` would have to contain `k`, whereas `dest.Definitions` wouldn't. And that's never the case, because `usedNames` is initialized with all the keys from `dest.Definitions`.

PR #23, opened 2017-11-20
Fix mapping and kind validation errors when array is passed

rutsky:
"Kind" is map with specific structure.


apelisse:
/lgtm

Thanks for fixing my mistakes :-)

PR #24, opened 2017-11-27
Fix schema merging, so references are updated properly

luksa:
When merging two schemas with conflicting definitions of the same object,
the aggregator would rename one of them, but wouldn't update all the
references ($ref), because it was updating copies due to golang's
pass-by-value.

Fixes https://github.com/kubernetes/kubernetes/issues/56417

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -58,8 +58,10 @@ func (s *referenceWalker) walkRef(ref spec.Ref) spec.Ref {
 	// We do not support external references yet.
 	if !s.alreadyVisited[refStr] && strings.HasPrefix(refStr, definitionPrefix) {
 		s.alreadyVisited[refStr] = true
-		def := s.root.Definitions[refStr[len(definitionPrefix):]]
+		k := refStr[len(definitionPrefix):]
+		def := s.root.Definitions[k]
 		s.walkSchema(&def)
+		s.root.Definitions[k] = def
 	}
 	return s.walkRefCallback(ref)
 }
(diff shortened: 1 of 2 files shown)


apelisse:
/assign @mbohlool 
/cc @cheftako 

Same, I don't know aggregator.

pmorie:
@cheftako do you have some review bandwidth for this PR?

pkg/aggregator/aggregator.go
```diff
@@ -69,23 +71,26 @@ func (s *referenceWalker) walkSchema(schema *spec.Schema) {
...
 	}
-	for _, v := range schema.Properties {
+	for k, v := range schema.Properties {
 		s.walkSchema(&v)
+		schema.Properties[k] = v
 	}
-	for _, v := range schema.PatternProperties {
+	for k, v := range schema.PatternProperties {
 		s.walkSchema(&v)
+		schema.PatternProperties[k] = v
```
mbohlool: Can we do it consistently. e.g. for cases like this (and above):

```
for k, _ := range schema.PatternProperties {
  s.walkSchema(&schema.PatternProperties[k])
}
```

mbohlool:
one comment, otherwise looks good. Sorry for the delay.

pkg/aggregator/aggregator.go
luksa: We can't, because `Schema.PatternProperties` is a map and you can't get a pointer to a map entry. 

I could change the rest to make them consistent with all the above, but that would just add unnecessary lines of code and would also introduce copying of the array values, which is unnecessary.

mbohlool: make sense.

PR #25, opened 2017-11-27
Handle "any value" openapi definitions properly

luksa:
This will allow us to support the k8s RawExtension object

### DIFF
--- a/pkg/util/proto/openapi.go
+++ b/pkg/util/proto/openapi.go
@@ -55,6 +55,7 @@ type SchemaVisitor interface {
 	VisitMap(*Map)
 	VisitPrimitive(*Primitive)
 	VisitKind(*Kind)
+	VisitArbitraryValue(*ArbitraryValue)
 	VisitReference(Reference)
 }
 
@@ -242,19 +243,19 @@ func (p *Primitive) GetName() string {
 	return fmt.Sprintf("%s (%s)", p.Type, p.Format)
 }
 
-// AnyValue is an arbitrary value (object, array or primitive)
-type AnyValue struct {
+// ArbitraryValue is a value of any type (primitive, object or array)
+type ArbitraryValue struct {
 	BaseSchema
 }
 
-var _ Schema = &AnyValue{}
+var _ Schema = &ArbitraryValue{}
 
-func (m *AnyValue) Accept(v SchemaVisitor) {
-	// A AnyValue accepts any type of value, so it's not necessary to call the SchemaVisitor
+func (a *ArbitraryValue) Accept(v SchemaVisitor) {
+	v.VisitArbitraryValue(a)
 }
 
-func (m *AnyValue) GetName() string {
-	return "Any value"
+func (a *ArbitraryValue) GetName() string {
+	return "Arbitrary value (primitive, object or array)"
 }
 
 // Reference implementation depends on the type of document.
--- a/pkg/util/proto/openapi.go
+++ b/pkg/util/proto/openapi.go
@@ -55,7 +55,7 @@ type SchemaVisitor interface {
 	VisitMap(*Map)
 	VisitPrimitive(*Primitive)
 	VisitKind(*Kind)
-	VisitArbitraryValue(*ArbitraryValue)
+	VisitArbitrary(*Arbitrary)
 	VisitReference(Reference)
 }
 
@@ -243,18 +243,18 @@ func (p *Primitive) GetName() string {
 	return fmt.Sprintf("%s (%s)", p.Type, p.Format)
 }
 
-// ArbitraryValue is a value of any type (primitive, object or array)
-type ArbitraryValue struct {
+// Arbitrary is a value of any type (primitive, object or array)
+type Arbitrary struct {
 	BaseSchema
 }
 
-var _ Schema = &ArbitraryValue{}
+var _ Schema = &Arbitrary{}
 
-func (a *ArbitraryValue) Accept(v SchemaVisitor) {
-	v.VisitArbitraryValue(a)
+func (a *Arbitrary) Accept(v SchemaVisitor) {
+	v.VisitArbitrary(a)
 }
 
-func (a *ArbitraryValue) GetName() string {
+func (a *Arbitrary) GetName() string {
 	return "Arbitrary value (primitive, object or array)"
 }
 
--- a/pkg/util/proto/openapi.go
+++ b/pkg/util/proto/openapi.go
@@ -242,6 +242,21 @@ func (p *Primitive) GetName() string {
 	return fmt.Sprintf("%s (%s)", p.Type, p.Format)
 }
 
+// AnyValue is an arbitrary value (object, array or primitive)
+type AnyValue struct {
+	BaseSchema
+}
+
+var _ Schema = &AnyValue{}
+
+func (m *AnyValue) Accept(v SchemaVisitor) {
+	// A AnyValue accepts any type of value, so it's not necessary to call the SchemaVisitor
+}
+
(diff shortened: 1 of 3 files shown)


luksa:
Fixes https://github.com/kubernetes/kubernetes/issues/55890

pkg/util/proto/openapi.go
```diff
@@ -242,6 +242,21 @@ func (p *Primitive) GetName() string {
 	return fmt.Sprintf("%s (%s)", p.Type, p.Format)
 }
 
+// AnyValue is an arbitrary value (object, array or primitive)
```
apelisse: How is one supposed to use that object? It doesn't have any subtype? any data? What abstraction is it mapping to? Can you point me to the specific OpenAPI specification section describing "any"?

pkg/util/proto/openapi.go
```diff
@@ -242,6 +242,21 @@ func (p *Primitive) GetName() string {
...
+	BaseSchema
+}
+
+var _ Schema = &AnyValue{}
+
+func (m *AnyValue) Accept(v SchemaVisitor) {
+	// A AnyValue accepts any type of value, so it's not necessary to call the SchemaVisitor
+}
+
+func (m *AnyValue) GetName() string {
```
apelisse: The lack of better description is a tell that this is missing something.

pkg/util/proto/openapi.go
apelisse: We'll also probably need to change the SchemaVisitor to add that.

apelisse:
OK, I've read the other pull-request that you've created. I might have misunderstood the intent here. Looking at it again.

apelisse:
OK I think this is it: https://tools.ietf.org/html/draft-zyp-json-schema-04#section-5.1

pkg/util/proto/openapi.go
apelisse: Can we figure out a better name? My initial understanding was that this was "AnyOf" (from openapi 3.0).

pkg/util/proto/openapi.go
```diff
@@ -242,6 +242,21 @@ func (p *Primitive) GetName() string {
...
 
+// AnyValue is an arbitrary value (object, array or primitive)
+type AnyValue struct {
+	BaseSchema
+}
+
+var _ Schema = &AnyValue{}
+
+func (m *AnyValue) Accept(v SchemaVisitor) {
+	// A AnyValue accepts any type of value, so it's not necessary to call the SchemaVisitor
```
apelisse: I think that decision should be taken by each individual visitor, what do you think?

apelisse: @pwittrock Do you have an opinion?

pwittrock: Yes, it should call a schema visitor.  The visitor may want to do something.

luksa: OK, that was my initial implementation. Then, since all the visitors' methods were empty, I thought I could remove it, since all visitors should always accept any value. 

I'll add the methods again and also rename this to `ArbitraryValue` or whatever you guys think is a better name.

pkg/util/proto/openapi.go
```diff
@@ -55,6 +55,7 @@ type SchemaVisitor interface {
 	VisitMap(*Map)
 	VisitPrimitive(*Primitive)
 	VisitKind(*Kind)
+	VisitArbitraryValue(*ArbitraryValue)
```
apelisse: I'm happy you thought of that word `Arbitrary`. Can I nit and ask you to just keep `Arbitrary` without the `Value`?

apelisse:
That looks good to me. I have no right here, but lgtm. @mbohlool

ericchiang:
On master I'm getting this when trying to generate the openapi schema with this change.
```
$ ./hack/update-openapi-spec.sh
make: Entering directory '/home/eric/src/k8s.io/kubernetes'                                                                                                                                                        
make[1]: Entering directory '/home/eric/src/k8s.io/kubernetes' 
# ...
    cmd/kube-apiserver                              
# k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/strategicpatch                                   
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/meta.go:166:17: cannot use kindItem (type *kindItem) as type "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to s.Schema
.Accept:                                            
        *kindItem does not implement "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)                                                                   
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/meta.go:181:17: cannot use sliceItem (type *sliceItem) as type "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to s.Sche
ma.Accept:                                          
        *sliceItem does not implement "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)                                                                  
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/types.go:55:5: cannot use kindItem literal (type *kindItem) as type LookupPatchItem in assignment:                                                              
        *kindItem does not implement LookupPatchItem (missing VisitArbitrary method)                     
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/types.go:79:28: cannot use item (type *kindItem) as type "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to schema.SubSc
hema().Accept:                                      
        *kindItem does not implement "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)                                                                   
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/types.go:118:5: cannot use sliceItem literal (type *sliceItem) as type LookupPatchItem in assignment:                                                           
        *sliceItem does not implement LookupPatchItem (missing VisitArbitrary method)                    
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/types.go:146:28: cannot use item (type *sliceItem) as type "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to schema.Sub
Schema().Accept:                                    
        *sliceItem does not implement "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)                                                                  
vendor/k8s.io/apimachinery/pkg/util/strategicpatch/types.go:169:18: cannot use item (type *sliceItem) as type "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to subschema.
Accept:                                             
        *sliceItem does not implement "k8s.io/kubernetes/vendor/k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)    
```

ericchiang:
Actually sorry false alarm. I don't think this is related to this change. sorry for the noise.

pkg/util/proto/openapi.go
```diff
@@ -55,6 +55,7 @@ type SchemaVisitor interface {
 	VisitMap(*Map)
 	VisitPrimitive(*Primitive)
 	VisitKind(*Kind)
+	VisitArbitrary(*Arbitrary)
```
ericchiang: Expanding this interface causes a ton of code to break in kubernetes/kubernetes...

https://github.com/kubernetes/kubernetes/pull/57059

Issue #27, opened 2017-12-11
changes around VisitArbitary rename broke building an initializer


xinyanmsft:
we have an Go app that run as an Kubernetes initializer. It imports the following:

    "k8s.io/client-go/tools/clientcmd"

    "github.com/ghodss/yaml"

    "k8s.io/api/apps/v1beta1"
    corev1 "k8s.io/api/core/v1"
    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
    "k8s.io/apimachinery/pkg/fields"
    "k8s.io/apimachinery/pkg/runtime"
    "k8s.io/apimachinery/pkg/types"
    "k8s.io/apimachinery/pkg/util/strategicpatch"
    "k8s.io/apimachinery/pkg/watch"
    "k8s.io/client-go/kubernetes"
    "k8s.io/client-go/rest"
    "k8s.io/client-go/tools/cache"

And then invokes 2 two merge this way: 
patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, v1beta1.Deployment{})

This works fine up till recently, where building the code failed with the following error:

# k8s.io/apimachinery/pkg/util/strategicpatch
../k8s.io/apimachinery/pkg/util/strategicpatch/meta.go:166:17: cannot use kindItem (type *kindItem) as type "k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to s.Schema.Accept:
        *kindItem does not implement "k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)
../k8s.io/apimachinery/pkg/util/strategicpatch/meta.go:181:17: cannot use sliceItem (type *sliceItem) as type "k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to s.Schema.Accept:
        *sliceItem does not implement "k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor (missing VisitArbitrary method)
../k8s.io/apimachinery/pkg/util/strategicpatch/types.go:55:5: cannot use kindItem literal (type *kindItem) as type LookupPatchItem in assignment:
        *kindItem does not implement LookupPatchItem (missing VisitArbitrary method)
../k8s.io/apimachinery/pkg/util/strategicpatch/types.go:79:28: cannot use item (type *kindItem) as type "k8s.io/kube-openapi/pkg/util/proto".SchemaVisitor in argument to schema.SubSchema().Accept:
[...]

PR #28, opened 2017-12-12
remove VisitArbitrary from SchemaVisitor, make different interface

ericchiang:
Lots of things in k8s.io/kubernetes implement SchemaVisitor, and
adding a new interface method breaks that.[1] Instead of adding
VisitArbitrary directly to that interface introduce a new
SchemaVisitorArbitrary that types can optionally implement to visit
an arbitrary type.

/assign @luksa
/assign @apelisse

Modifies change in https://github.com/kubernetes/kube-openapi/pull/25

Fixes https://github.com/kubernetes/kube-openapi/issues/27

[1] https://github.com/kubernetes/kubernetes/pull/57059#issuecomment-350896736

### DIFF
--- a/pkg/util/proto/openapi.go
+++ b/pkg/util/proto/openapi.go
@@ -55,8 +55,15 @@ type SchemaVisitor interface {
 	VisitMap(*Map)
 	VisitPrimitive(*Primitive)
 	VisitKind(*Kind)
-	VisitArbitrary(*Arbitrary)
 	VisitReference(Reference)
+}
+
+// SchemaVisitorArbitrary is an additional visitor interface which handles
+// arbitrary types. For backwards compatability, it's a separate interface
+// which is checked for at runtime.
+type SchemaVisitorArbitrary interface {
+	SchemaVisitor
+	VisitArbitrary(*Arbitrary)
 }
 
 // Schema is the base definition of an openapi type.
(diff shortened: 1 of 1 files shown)


apelisse:
I'm really sad and unexcited about this change, but that might be the least bad thing to do .. :-/
We're lucky that the empty implementation happens to do something that works here, we might not be lucky next time. Also, I don't think updating kubernetes would be so bad (even though I can't disagree that it's used in many places, so kind of tedious).

/lgtm

luksa:
Note that I've already implemented the method in all implementations of `SchemaVisitor` in Kubernetes core (see https://github.com/kubernetes/kubernetes/pull/56434). And I'm wondering if having it in a separate interface is even ok. If people implement a `SchemaVisitor` and forget to implement the `VisitArbitrary` method, the `Arbitrary` objects obviously won't get visited. I feel like devs should be forced to implement the method (to reduce the chance of them forgetting to do that).

apelisse:
@luksa, could you potentially split your PR in two? One piece that is non-controversial (updating kube-openapi and the `SchemaVisitor`), and one that changes the RawExtension? So that we can get the `SchemaVisitor` merged asap? I definitely agree with you that it's better to keep one visitor that does everything.

ericchiang:
FYI this is broke client-go too https://github.com/kubernetes/client-go/issues/343#issue-280846569

mbohlool:
> FYI this is broke client-go too kubernetes/client-go#343 (comment)

Should we revert any change to fix client-go and others? I am not sure about details of this change, but would like to revert any change resulted in client-go failure.

mbohlool:
I am going to merge this. Hopefully that fixes client-go.

ericchiang:
@mbohlool it breaks client-go and HEAD of this repo (which we only support best effort today). e.g. it breaks `go get k8s.io/client-go/...`

apelisse:
@lukas let's get your PR merged, and then we'll revert that one.

PR #34, opened 2018-01-29
generators: add support for nested lists

jessesuen:
This change makes it possible to generate OpenAPI models when a field is a list-of-lists. 

Consider the following type:

```golang
// GeoJSON
// +k8s:openapi-gen=true
type GeoJSON struct {
    Name        string    `json:"name"`
    Coordinates [][]int64 `json:"coordinates"`
}
```

When the openapi codegen is run against it, it currently results in the following error:
```
ERROR: logging before flag.Parse: E0128 23:09:41.045732   55143 openapi.go:338] Error when generating: coordinates, Coordinates [][]int64
ERROR: logging before flag.Parse: F0128 23:09:41.045761   55143 main.go:44] Error: Failed executing generator: some packages had errors:
slice Element kind Slice is not supported in [][]int64
```

Nested lists has been supported since Swagger 2.0. See:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#itemsObject

image

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -634,6 +634,8 @@ func (g openAPITypeWriter) generateSliceProperty(t *types.Type) error {
 		return fmt.Errorf("please add type %v to getOpenAPITypeFormat function", elemType)
 	case types.Struct:
 		g.generateReferenceProperty(elemType)
+	case types.Slice, types.Array:
+		g.generateSliceProperty(elemType)
 	default:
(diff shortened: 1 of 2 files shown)


jessesuen:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.

Ok it's been signed.

mbohlool:
Look good to me. Thanks.

PR #36, opened 2018-02-07
Endpoint properly handles Accept header

roycaihw:
Sort Accept header with q parameter

**NOTE:** I manually changed `Godep.json`. `goautoneg` is used in kubernetes already.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"bitbucket.org/ww/goautoneg"
+
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/NYTimes/gziphandler"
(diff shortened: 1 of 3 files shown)


roycaihw:
cc @liggitt

pkg/handler/handler.go
```diff
@@ -219,33 +221,41 @@ func RegisterOpenAPIVersionedService(openapiSpec *spec.Swagger, servePath string
...
 			w.Header().Add("Vary", "Accept")
-
-			getDataAndETag = files[decipherableFormat]
-			if getDataAndETag == nil {
-				// Return 406 for not acceptable format
-				w.WriteHeader(406)
-				return
+			for _, clause := range clauses {
+				for _, accepts := range accepted {
+					if clause.Type == accepts.Type && clause.SubType == accepts.SubType ||
```
liggitt: nit, might be more readable unnested like this:

```go
if clause.Type != accepts.Type && clause.Type != "*" {
  continue
}
if clause.SubType != accepts.SubType && clause.SubType != "*" {
  continue
}

data, etag, lastModified := accepts.GetDataAndETag()
w.Header().Set("Etag", etag)
// ServeContent will take care of caching using eTag.
http.ServeContent(w, r, servePath, lastModified, bytes.NewReader(data))
```

liggitt:
just noticed there are no tests in this package... probably worthwhile to test here before doing a vendor cycle into kube

pkg/handler/handler.go
roycaihw: updated

roycaihw:
@mbohlool @liggitt Could we merge this one before the code slush, to avoid changing the deprecation schedule?

liggitt:
code change LGTM
it's a little concerning that vendoring has to be updated manually in this repo

mbohlool:
@liggitt I agree, we need to port some of the main repo tools to this repo or develop a set of generic tool from them for any repo like this.

mbohlool:
this code looks good to me too.

PR #37, opened 2018-02-12
add support for openapiv3 json output format

munnerz:
This PR implements initial support for generating an OpenAPI v3 schema from a resource type.

Paired with a change to the kubernetes repo to openapi-gen we can use this to generate openapiv3 schemas for validating CRDs.

I've used the gengo `parser` package to parse the types, and have basically just reimplemented the existing `openAPITypeWriter` but constructing `github.com/go-openapi/spec` (i.e. the ones output by the existing openapi-gen) types and then serialising them to JSON.

A schema generated for a [cert-manager Certificate resource](https://github.com/jetstack/cert-manager/blob/master/pkg/apis/certmanager/v1alpha1/types.go#L209):

```json
{
  "description": "Certificate is a type to represent a Certificate from ACME",
  "properties": {
    "spec": {
      "properties": {
        "acme": {
          "properties": {
            "config": {
              "type": "array",
              "additionalProperties": {
                "properties": {
                  "dns01": {
                    "properties": {
                      "provider": {
                        "type": "string"
                      }
                    }
                  },
                  "domains": {
                    "type": "array",
                    "additionalProperties": {
                      "type": "string"
                    }
                  },
                  "http01": {
                    "properties": {
                      "ingress": {
                        "type": "string"
                      },
                      "ingressClass": {
                        "type": "string"
                      }
                    }
                  }
                }
              }
            }
          }
        },
        "commonName": {
          "type": "string"
        },
        "dnsNames": {
          "description": "DNSNames is a list of subject alt names to be used on the Certificate",
          "type": "array",
          "additionalProperties": {
            "type": "string"
          }
        },
        "issuerRef": {
          "description": "IssuerRef is a reference to the issuer for this certificate. If the namespace field is not set, it is assumed to be in the same namespace as the certificate. If the namespace field is set to the empty value \"\", a ClusterIssuer of the given name will be used. Any other value is invalid.",
          "properties": {
            "kind": {
              "type": "string"
            },
            "name": {
              "type": "string"
            }
          }
        },
        "secretName": {
          "type": "string"
        }
      }
    },
    "status": {
      "properties": {
        "acme": {
          "properties": {
            "orderURL": {
              "type": "string"
            }
          }
        },
        "conditions": {
          "type": "array",
          "additionalProperties": {
            "properties": {
              "lastTransitionTime": {
                "description": "LastTransitionTime is the timestamp corresponding to the last status change of this condition.",
                "properties": {
                  "Time": {
                    "type": "string",
                    "format": "date-time"
                  }
                }
              },
              "message": {
                "type": "string"
              },
              "reason": {
                "type": "string"
              },
              "status": {
                "type": "string"
              },
              "type": {
                "type": "string"
              }
            }
          }
        }
      }
    }
  }
}
```

Open questions:

* If you look closely above, you can see that `lastTransitionTime` has been encoded as a struct with a field named `Time`.

This isn't going to validate correctly... I notice in the existing openapi.go file, there is some mention of skipping json refs for 'well known' fields (https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/openapi.go#L403). Here it can be handled a bit differently, as we defer resolving types in the Golang generator.

We need to teach the `generateMembers` function about these kinds of fields, and somehow instruct the calling method to set the `type` and `format` on *its own* schema. These feels hacky. Does anyone know how we can better handle this?

* In order to skip `metadata`, `kind` and `apiVersion` fields, I have restricted the generator to only generating schemas for `spec` and `status` fields. @nikhita does CRD validation already impose this restriction? I don't see it mentioned on `kubernetes.io/docs` (https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#validation)

/cc @sttts @frankgreco @deads2k

ref https://github.com/kubernetes/kubernetes/issues/59154

### DIFF
--- a/pkg/generators/openapi_json.go
+++ b/pkg/generators/openapi_json.go
@@ -26,21 +26,29 @@ type openAPIGeneratorJSON struct {
 	targetPkg, targetType string
 }
 
-func (g openAPIGeneratorJSON) Execute() error {
+func (g openAPIGeneratorJSON) ExecuteAndWrite() error {
+	schema, err := g.Generate()
+	if err != nil {
+		return err
+	}
+	return json.NewEncoder(g.output).Encode(schema)
+}
+
+func (g openAPIGeneratorJSON) Generate() (*spec.Schema, error) {
 	b := parser.New()
 	if err := b.AddDir(g.targetPkg); err != nil {
-		return err
+		return nil, err
 	}
 	u, err := b.FindTypes()
 	if err != nil {
-		return err
+		return nil, err
 	}
 	pkg := u.Package(g.targetPkg)
 	t := pkg.Type(g.targetType)
 	return g.generate(t)
 }
 
-func (g openAPIGeneratorJSON) generate(t *types.Type) error {
+func (g openAPIGeneratorJSON) generate(t *types.Type) (*spec.Schema, error) {
 	schema := &spec.Schema{}
 	// Only generate for struct type and ignore the rest
 	switch t.Kind {
@@ -49,14 +57,14 @@ func (g openAPIGeneratorJSON) generate(t *types.Type) error {
 		// This really would involve compiling the code in question and calling
 		// the function.
 		if hasOpenAPIDefinitionMethod(t) {
-			return nil
+			return nil, fmt.Errorf("Cannot compute validation schema for manual openapi definitions")
 		}
 		props := spec.SchemaProps{
 			Description: strings.TrimSpace(parseDescription(t.CommentLines)),
 		}
 		properties, req, err := g.generateMembers(t, true)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		props.Properties = properties
 		props.Required = req
@@ -64,11 +72,11 @@ func (g openAPIGeneratorJSON) generate(t *types.Type) error {
 
 		extensions, err := g.generateExtensions(t.CommentLines)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		schema.VendorExtensible = *extensions
 	}
-	return json.NewEncoder(g.output).Encode(schema)
+	return schema, nil
 }
 
 func (g openAPIGeneratorJSON) generateMembers(t *types.Type, root bool) (map[string]spec.Schema, []string, error) {
--- a/pkg/generators/openapi_json.go
+++ b/pkg/generators/openapi_json.go
@@ -1,0 +1,292 @@
+package generators
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/go-openapi/spec"
+	"github.com/golang/glog"
+	"k8s.io/gengo/parser"
+	"k8s.io/gengo/types"
+	openapi "k8s.io/kube-openapi/pkg/common"
+)
+
+func NewOpenAPIGenJSON(output io.Writer, targetPkg, targetType string) *openAPIGeneratorJSON {
+	return &openAPIGeneratorJSON{output, targetPkg, targetType}
+}
+
+// openAPIGeneratorJSON is a custom generator implementation that builds an
+// openapi spec structure and writes it out as JSON or YAML. It reuses the
+// gengo parser package for parsing types, but otherwise is a custom
+// implementation due to gengo being limited to generating Go code.
+type openAPIGeneratorJSON struct {
+	output                io.Writer
+	targetPkg, targetType string
+}
+
+func (g openAPIGeneratorJSON) Execute() error {
+	b := parser.New()
+	if err := b.AddDir(g.targetPkg); err != nil {
+		return err
+	}
+	u, err := b.FindTypes()
+	if err != nil {
+		return err
+	}
+	pkg := u.Package(g.targetPkg)
+	t := pkg.Type(g.targetType)
+	return g.generate(t)
+}
+
+func (g openAPIGeneratorJSON) generate(t *types.Type) error {
+	schema := &spec.Schema{}
+	// Only generate for struct type and ignore the rest
+	switch t.Kind {
+	case types.Struct:
+		// TODO: support manually defined openapi definitions.
+		// This really would involve compiling the code in question and calling
+		// the function.
+		if hasOpenAPIDefinitionMethod(t) {
+			return nil
+		}
+		props := spec.SchemaProps{
+			Description: strings.TrimSpace(parseDescription(t.CommentLines)),
+		}
+		properties, req, err := g.generateMembers(t, true)
+		if err != nil {
+			return err
+		}
+		props.Properties = properties
+		props.Required = req
+		schema.SchemaProps = props
+
+		extensions, err := g.generateExtensions(t.CommentLines)
+		if err != nil {
+			return err
+		}
+		schema.VendorExtensible = *extensions
+	}
+	return json.NewEncoder(g.output).Encode(schema)
+}
+
+func (g openAPIGeneratorJSON) generateMembers(t *types.Type, root bool) (map[string]spec.Schema, []string, error) {
+	props := map[string]spec.Schema{}
+	var required []string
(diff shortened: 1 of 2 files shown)


frankgreco:
What about fields like `oneOf`? Consider the following example where only one would be required. Were you planning on using the existing comment tag `optional` to dynamically decide whether `oneOf` should be used?

```go
type Foo struct {
  optionOne string
  optionTwo string
  optionThree string
}
```

nikhita:
> In order to skip metadata, kind and apiVersion fields, I have restricted the generator to only generating schemas for spec and status fields. @nikhita does CRD validation already impose this restriction?

CRD validation does not impose any such restrictions for validating fields (like metadata, etc).

We validate metadata, kind and apiVersion without CRD validation. But if a user wants to specify `metadata`, they can do so. https://github.com/kubernetes/kubernetes/blob/ae0f45e855381d3731f8e66e536d1ad093ea9948/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L126-L139.

> I have restricted the generator to only generating schemas for spec and status fields.

While this makes sense because we should have ideally have only spec and status, there might be custom resources which have other top level fields as well. We should probably support them too?

> Paired with a change to the kubernetes repo to openapi-gen we can use this to generate openapiv3 schemas for validating CRDs.

I still have to look at the PR thoroughly so apologies if this is wrong but if IIUC, this generates `go-openapi/spec.Schema`, right? 

Will this schema be used as input to `CRD.Spec.Validation.OpenAPIV3Schema`? For validating CRDs, won't we need the kube types - `JSONSchemaProps` to be generated then? The conversion of kube types to openapiv3 types is done internally in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/validation.go.

munnerz:
@nikhita 

> CRD validation does not impose any such restrictions for validating fields (like metadata, etc).
> While this makes sense because we should have ideally have only spec and status, there might be custom resources which have other top level fields as well. We should probably support them too?

👍 makes sense. Which fields should be excluded from generation then? I guess some/most `metadata` fields are *already* validated, so it seems redundant to generate the entire set?

> I still have to look at the PR thoroughly so apologies if this is wrong but if IIUC, this generates go-openapi/spec.Schema, right?
> Will this schema be used as input to CRD.Spec.Validation.OpenAPIV3Schema? For validating CRDs, won't we need the kube types - JSONSchemaProps to be generated then? The conversion of kube types to openapiv3 types is done internally in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/validation.go.

Yep it does, and yes that is what is intended. It is output as JSON to be copy/pasted into a CRD manifest. It does **not** generate golang code (that is what the existing generator does).

In terms of getting the kubernetes openapi descriptions (i.e. for metadata, or if a user is using a kube-type within one of their own types), this is handled via `--extra-peer-dirs` like normal with generators because we are using the same gengo parser package. That said, I'm not sure if I've exposed a way to set `ExtraPeerDirs` in this pull request.

munnerz:
Ping on this - who is best to review? I'll need to make a corresponding change in kubernetes/kubernetes (to openapi-gen) in order for this to be easily try-able.

Let me know if there's anything I can do to push this forward!

deads2k:
> Ping on this - who is best to review? I'll need to make a corresponding change in kubernetes/kubernetes (to openapi-gen) in order for this to be easily try-able.

Probaby @mbohlool

perotinus:
What's the status of this? I'm looking into CRDs for the cluster registry, and being able to generate the OpenAPI spec is very desirable to maintain compatibility with what's already there.

pkg/generators/openapi_json.go
```diff
@@ -0,0 +1,292 @@
...
+	switch elemType.Kind {
+	case types.Builtin:
+		return nil, fmt.Errorf("please add type %v to getOpenAPITypeFormat function", elemType)
+	case types.Struct:
+		props, req, err := g.generateMembers(elemType, false)
+		if err != nil {
+			return nil, err
+		}
+		glog.V(1).Infof("Got member props: %+v", props)
+		schema.AdditionalProperties = &spec.SchemaOrBool{Schema: &spec.Schema{}}
```
sttts: why do we create AdditionalProperties for members?

pkg/generators/openapi_json.go
```diff
@@ -0,0 +1,292 @@
...
+	schema := &spec.Schema{}
+	switch elemType.Kind {
+	case types.Builtin:
+		return nil, fmt.Errorf("please add type %v to getOpenAPITypeFormat function", elemType)
+	case types.Struct:
+		props, req, err := g.generateMembers(elemType, false)
+		if err != nil {
+			return nil, err
+		}
+		schema.AdditionalProperties = &spec.SchemaOrBool{
```
sttts: also here: AdditionalProperties confuses me. Would expect `Properties`.

sttts:
I think the overall direction is fine.

I would like to see test cases, maybe similar to gengo: https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/Makefile.

tamalsaha:
Are there any plans to merge/unify the the JsonSchema type for CRDs and the output of `openapi-gen` ? Also, will it be possible to generate all the different types of validation rules using the comments in the Go types like `openapi-gen` does?

munnerz:
Hey Tamal -

That's exactly what this PR aims to do 😄 although we cannot replace the
existing default output format of openapi-gen as it is used by apiservers.

This will introduce a new flag to the tool that specifies the output format
as JSON 🙂 I have code changes locally for that which are blocked on this
PR.

Regarding comments for validation rules - yes.

pkg/generators/openapi_json.go
anubhakushwaha: @sttts As per the OpenAPI Specification for *Object Schema Model with Map/Dictionary Properties* [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#model-with-mapdictionary-properties) the inline members are *additional properties* merged via [this](https://github.com/OAI/OpenAPI-Specification/pull/283) PR. This is the structure of schema generated [here](https://drive.google.com/file/d/1oEn2krZASH_RZ2OaWJe8uzaMpAkRRIzp/view?usp=sharing) @munnerz would be better to answer though.

sttts: @anubhakushwaha I think you are right. This is for maps with struct values. So the  `additionalProperties` is to be used to describe the values of unknown keys (of the map).

pkg/generators/openapi_json.go
sttts: for slices there is `items`, isn't it? I.e. `items` is for slices what `additionalPropoerties` is for structs.

anubhakushwaha: [Reference](https://arxiv.org/pdf/1701.02221.pdf)
Added if it makes sense. @sttts 
**Array schemas** 
 Array schemas are specified with the "type":  "array" keyword. For arrays there are two ways of specifying what kind of documents we find in arrays. We can use a pair
"items" : [J1,... JN] to specify a document with an array of n elements, where each i-th element must satisfy schema Ji.**We can also use "additionalItems":J to specify that all elements in the array must satisfy schema J.** If both keywords are used together, then we allow the array to have more values than those specified with items, as long as they agree with the schema specified in additionalItems.
For example, the following schema validates against arrays of at least 2 elements, where the first two are strings and the remaining ones, if they exists, are numbers.
{
"type": "array",
"items": [{"type":"string"}, {"type":"string"}],
"additionalItems": {"type":"number"},
"uniqueItems":true
}

sttts:
https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/crd/cmd can replace this.

PR #38, opened 2018-02-21
Client-side OpenAPI schema fixes

luksa:
These changes allow the client to properly handle the server-side changes planned to be introduced in kube v1.11:
- RawExtension type will be removed and replaced with `"type": "object"` without either the `additionalProperties` or the `properties` field.
- All type definitions stemming from Go structs will contain the `"type": "object"` property (next to the `properties` field, which they already contain).


pkg/util/proto/document.go
```diff
@@ -149,11 +154,13 @@ func (d *Definitions) parsePrimitive(s *openapi_v2.Schema, path *Path) (Schema,
 	}
 	switch t {
 	case String:
+		fallthrough
```
apelisse: I don't know which version I dislike the most :-)

pkg/util/proto/document.go
```diff
@@ -219,27 +226,37 @@ func (d *Definitions) parseArbitrary(s *openapi_v2.Schema, path *Path) (Schema,
 // ParseSchema creates a walkable Schema from an openapi schema. While
 // this function is public, it doesn't leak through the interface.
 func (d *Definitions) ParseSchema(s *openapi_v2.Schema, path *Path) (Schema, error) {
+	if s.GetXRef() != "" {
+		return d.parseReference(s, path)
+	}
 	objectTypes := s.GetType().GetValue()
-	if len(objectTypes) == 1 {
+	if len(objectTypes) == 0 {
```
apelisse: We could avoid the repetition here I think, by extracting the `objectTypes[0]` into a variable, that we would default to `object`?

pkg/util/proto/document.go
```diff
@@ -219,27 +226,37 @@ func (d *Definitions) parseArbitrary(s *openapi_v2.Schema, path *Path) (Schema,
...
-	}
-	if s.GetProperties() != nil {
-		return d.parseKind(s, path)
-	}
-	if len(objectTypes) == 0 || (len(objectTypes) == 1 && objectTypes[0] == "") {
-		return d.parseArbitrary(s, path)
+		return d.parsePrimitive(s, path)
+	} else {
+		// the OpenAPI generator never generates (nor it ever did in the past) OpenAPI type definitions with multiple types
+		return nil, newSchemaError(path, "definitions with multiple types aren't supported")
```
apelisse: Maybe this should be at the beginning

pkg/util/proto/document.go
luksa: Yeah, I've now removed `fallthrough` and instead added a comment (`// do nothing`). Give how go doesn't fallthrough by default, a comment will help most of us that are used to how switch statements behave in other languages.

pkg/util/proto/document.go
luksa: I've replaced the if/else if with a switch. Not sure what you meant about extracting it into a variable - it wouldn't change anything as it's only used once.

pkg/util/proto/document.go
luksa: This is now the default case in a switch statement.

pkg/util/proto/openapi_test.go
```diff
@@ -132,6 +133,63 @@ var _ = Describe("Reading apps/v1beta1/Deployment from openAPIData", func() {
 	})
 })
 
+var _ = Describe("Reading apps/v1beta1/Deployment from v1.11 openAPIData", func() {
```
apelisse: Having test for both versions is great!

pkg/util/proto/openapi_test.go
```diff
@@ -27,8 +27,9 @@ import (
 )
 
 var fakeSchema = testing.Fake{Path: filepath.Join("testing", "swagger.json")}
+var fakeSchemaNext = testing.Fake{Path: filepath.Join("testing", "swagger_next.json")}
```
apelisse: I'm also writing a change to that code, and I don't want to create a new swagger file for every new piece of feature. Can we just have a new type that uses your feature?

apelisse:
/lgtm

That's a net improvement, we might want to simplify the test with the two swagger files.

PR #41, opened 2018-03-10
Typo fix "to provides"->"to provide"

AdamDang:
"... to provides access to it..."  appears 4 times in this doc.
We should use "...to provide access to it...".


mbohlool:
Can you please send one single PR/commit for all of typo fixes?

AdamDang:
Hi @mbohlool
Thanks for your reply. I started to send PR in github from about 1 month ago. At that time, I knew little
about the github, such as how to change two or more different files in one PR. Now I've got such skills. I'm glad to join the github and do something to make the codes better, and happy for those who helping me get skilled.
BTW, is the openapi project maintained monthly?

PR #43, opened 2018-04-02
Refactor to make it easier to add extension tags

seans3:
Refactors openapi generator to make it easier to add extension tags within comments.
Refactored "patchMergeStrategy" and "patchMergeKey" extension tags.
No new functionality.
Current unit tests pass.

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -40,14 +40,21 @@ const tagOptional = "optional"
 
 // Known values for the tag.
 const (
-	tagValueTrue               = "true"
-	tagValueFalse              = "false"
-	tagExtensionPrefix         = "x-kubernetes-"
-	tagPatchStrategy           = "patchStrategy"
-	tagPatchMergeKey           = "patchMergeKey"
-	patchStrategyExtensionName = "patch-strategy"
-	patchMergeKeyExtensionName = "patch-merge-key"
+	tagValueTrue  = "true"
+	tagValueFalse = "false"
 )
+
+// First-class Kubernetes extension tags within comments.
+var tags = [...]string{
+	"patchMergeKey",
+	"patchStrategy",
+}
+
+// Extension tag to openapi extension string.
+var tagToExtension = map[string]string{
+	"patchStrategy": "x-kubernetes-patch-strategy",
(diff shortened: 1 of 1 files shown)


apelisse:
Looks good to me. @mbohlool can you please review and merge this? Thanks

pkg/generators/openapi.go
```diff
@@ -453,23 +457,17 @@ func (g openAPITypeWriter) generateExtensions(CommentLines []string) error {
...
-		return err
-	}
-	if patchMergeKeyStructTag != patchMergeKeyCommentTag {
-		return fmt.Errorf("patchMergeKey in comment and struct tags should match for member (%s) of (%s)",
-			m.Name, parent.Name.String())
-	}
-	if patchStrategyStructTag != patchStrategyCommentTag {
-		return fmt.Errorf("patchStrategy in comment and struct tags should match for member (%s) of (%s)",
-			m.Name, parent.Name.String())
+func (g openAPITypeWriter) validateTags(m *types.Member, parent *types.Type) error {
```
mbohlool: Per our offiline discussion, please keep a temporary list of all tags need to be present in both comments and struct tags. Because the server is going to consume openapi spec back for apply, we do not need any of these tags in struct tags and comment tags should suffice.

mbohlool:
assuming you tested this within main repo and all tests are passing.

apelisse:
> assuming you tested this within main repo and all tests are passing.

Do you know what tests should be run specifically? To be safe.

mbohlool:
Not any specific test. I assume whatever we change here, we put them in the main repo and run at least unit/integration tests.

lavalamp:
It is time to write local tests here. Making people test by running stuff in the other repo is not a good dev experience and it's likely to lead to errors.

mbohlool:
We do have unit tests but they may not be complete. I need to check on that. And we will catch any failure when we vendor this repo in the main repo but things like client-go may use the head of this repo (`go get` hell) and we need to find a way to prevent breaking them.

apelisse:
> And we will catch any failure when we vendor this repo in the main repo

That seems wrong. Let's try to do better.

PR #50, opened 2018-04-12
Multi-value extensions

seans3:
* Multi-value OpenAPI extensions (e.g. x-kubernetes-list-type-keys: "port", "protocol")
* Validation of extension enum values (e.g. x-kubernetes-patch-strategy must be either "merge" or "retainKeys"). This validation only logs a warning now.
* Moved extension code into its own file.
* Unit tests for extension code.

### DIFF
--- a/pkg/generators/extension.go
+++ b/pkg/generators/extension.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generators
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/golang/glog"
+	"k8s.io/gengo/examples/set-gen/sets"
+	"k8s.io/gengo/types"
+)
+
+const extensionTag = "x-kubernetes-"
+
+// Extension tag to openapi extension
+var tagToExtension = map[string]string{
+	"patchMergeKey": "x-kubernetes-patch-merge-key",
+	"patchStrategy": "x-kubernetes-patch-strategy",
+	"listType":      "x-kubernetes-list-type",
+	"listMapKey":    "x-kubernetes-list-map-keys",
+}
+
+// Enum values per extension
+var allowedExtensionValues = map[string]sets.String{
+	"x-kubernetes-patch-strategy": sets.NewString("merge", "retainKeys"),
+	"x-kubernetes-list-type":      sets.NewString("atomic", "set", "map"),
+}
+
+// Extension encapsulates information necessary to generate an OpenAPI extension.
+// TODO(seans3): Create interface for an extension validation function, and create
+// validation functions for different extensions.
+type extension struct {
+	name   string   // Example: x-kubernetes-list-type
+	values []string // Example: atomic
+}
+
+func (e extension) validateAllowedValues() error {
+	// allowedValues not set means no restrictions on values.
+	allowedValues, exists := allowedExtensionValues[e.name]
+	if !exists {
+		return nil
+	}
+	// For each extension value, validate that it is allowed.
+	if !allowedValues.HasAll(e.values...) {
+		return fmt.Errorf("%s: value(s) %v not allowed. Allowed values: %v\n",
+			e.name, e.values, allowedValues.List())
+	}
+	return nil
+}
+
+func (e extension) hasMultipleValues() bool {
+	return len(e.values) > 1
+}
+
+// Returns sorted list of map keys. Needed for deterministic testing.
+func sortedMapKeys(m map[string][]string) []string {
+	keys := make([]string, len(m))
+	i := 0
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+// Parses comments to return openapi extensions
+func parseExtensions(comments []string) []extension {
+	extensions := []extension{}
+	// First, generate extensions from "+k8s:openapi-gen=x-kubernetes-*" annotations.
+	values := getOpenAPITagValue(comments)
+	for _, val := range values {
+		// Example: x-kubernetes-member-tag:member_test
+		if strings.HasPrefix(val, extensionTag) {
+			parts := strings.SplitN(val, ":", 2)
+			if len(parts) != 2 {
+				// TODO(seans3): Spit out source file/line number with error.
+				glog.V(2).Info(fmt.Sprintf("Invalid extension value: %v", val))
+				continue
+			}
+			e := extension{
+				name:   parts[0],           // Example: x-kubernetes-member-tag
+				values: []string{parts[1]}, // Example: member_test
+			}
+			extensions = append(extensions, e)
+		}
+	}
+	// Next, generate extensions from "tags" (e.g. +listType)
+	tagValues := types.ExtractCommentTags("+", comments)
+	for _, tag := range sortedMapKeys(tagValues) {
+		extensionName, exists := tagToExtension[tag]
+		if !exists {
+			continue
+		}
+		extensionValues := tagValues[tag]
+		e := extension{
+			name:   extensionName,
+			values: extensionValues,
+		}
+		// TODO(seans3): Validate the number of allowed values
+		// TODO(seans3): Validate the field type for this extension.
+		// Example: +listType is only allowed on an array.
+		if err := e.validateAllowedValues(); err != nil {
+			// For now, only log the extension validation errors.
+			glog.V(2).Info(err)
+		}
+		extensions = append(extensions, e)
+	}
+	return extensions
+}
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -51,13 +51,6 @@ var tempPatchTags = [...]string{
 	"patchStrategy",
 }
 
-// Extension tag to openapi extension string.
-var tagToExtension = map[string]string{
-	"patchMergeKey": "x-kubernetes-patch-merge-key",
-	"patchStrategy": "x-kubernetes-patch-strategy",
-	"listAttribute": "x-kubernetes-list-attribute",
-}
-
 func getOpenAPITagValue(comments []string) []string {
 	return types.ExtractCommentTags("+", comments)[tagName]
 }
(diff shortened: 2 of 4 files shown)


apelisse:
Minor nit, looks good to me otherwise.

pkg/generators/openapi_test.go
```diff
@@ -495,3 +495,62 @@ Dependencies: []string{
...
+	// +listTypeKey=port
+	// +listTypeKey=protocol
+	WithListField []string
+}
+		`)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(`"base/foo.Blah": {
+Schema: spec.Schema{
```
lavalamp: This is really hard to read. Can we print prettified json instead?

pkg/generators/openapi_test.go
```diff
@@ -261,11 +261,11 @@ Format: "",
 "WithListAttribute": {
 VendorExtensible: spec.VendorExtensible{
 Extensions: spec.Extensions{
-"x-kubernetes-list-attribute": "la",
+"x-kubernetes-list-type": "la",
```
lavalamp: what is "la"?

pkg/generators/openapi_test.go
```diff
@@ -115,8 +115,8 @@ type Blah struct {
 	// +patchStrategy=ps
 	// +patchMergeKey=pmk
 	WithStructTagExtension string `+"`"+`patchStrategy:"ps" patchMergeKey:"pmk"`+"`"+`
-	// a member with a list attribute
-	// +listAttribute=la
+	// a member with a list type
+	// +listType=la
```
lavalamp: Please use valid data in tests, since people look at them as examples.

lavalamp: Also why is a list attribute put on a string? Is this testing that it emits a warning/error?

pkg/generators/openapi.go
```diff
@@ -452,26 +433,29 @@ func (g openAPITypeWriter) generateExtensions(CommentLines []string) error {
...
-		}
-		if len(tagValue) > 0 {
-			extensions = append(extensions, NameValue{tagToExtension[tagKey], tagValue})
-		}
-	}
-	// If there are generated extensions, write them out to schema.
-	// Example: "x-kubernetes-patch-strategy" : "merge"
-	if len(extensions) == 0 {
+	// Next, generate "tag" extensions (e.g. "+patchMergeStrategy", "+listAttribute=...")
+	tagExtensions := getTagExtensions(CommentLines)
```
lavalamp: This is confusing to me. Why do we want two different parsing mechanisms for the tags?

pkg/generators/openapi_test.go
```diff
@@ -495,3 +495,62 @@ Dependencies: []string{
...
+	assert.Equal(`"base/foo.Blah": {
+Schema: spec.Schema{
+SchemaProps: spec.SchemaProps{
+Description: "Blah is a test.",
+Properties: map[string]spec.Schema{
+"WithListField": {
+VendorExtensible: spec.VendorExtensible{
+Extensions: spec.Extensions{
+"x-kubernetes-list-type": "map",
+"x-kubernetes-list-type-keys": []string{"port",
```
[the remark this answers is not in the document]
lavalamp: Yes, map-keys is better.

pkg/generators/openapi.go
```diff
@@ -452,26 +433,29 @@ func (g openAPITypeWriter) generateExtensions(CommentLines []string) error {
...
+	if len(extensions) == 0 && len(tagExtensions) == 0 {
 		return nil
 	}
+	// If any extensions exist, then emit code to create them.
 	g.Do("VendorExtensible: spec.VendorExtensible{\nExtensions: spec.Extensions{\n", nil)
 	for _, extension := range extensions {
 		g.Do("\"$.$\": ", extension.Name)
 		g.Do("\"$.$\",\n", extension.Value)
 	}
+	for _, tagExtension := range tagExtensions {
```
lavalamp: Please add tests until you find the bug in this code. :)

pkg/generators/openapi_test.go
lavalamp: Lol, this is not json, is it. It's not hard to programmatically invoke gofmt, I think. I guess that can be another PR.

pkg/generators/openapi.go
apelisse: I found one. I don't know if it's the same, but I found one :-)

pkg/generators/extension.go
```diff
@@ -0,0 +1,112 @@
...
+	"listTypeKey":   "x-kubernetes-list-type-keys",
+}
+
+// Enum values per extension
+var allowedExtensionValues = map[string][]string{
+	"x-kubernetes-patch-strategy": []string{"merge", "retainKeys"},
+	"x-kubernetes-list-type":      []string{"atomic", "set", "map"},
+}
+
+// TODO: Create interface for an extension validation function, and create
```
lavalamp: Godoc comment before TODO?

pkg/generators/extension.go
```diff
@@ -0,0 +1,112 @@
...
+var tagToExtension = map[string]string{
+	"patchMergeKey": "x-kubernetes-patch-merge-key",
+	"patchStrategy": "x-kubernetes-patch-strategy",
+	"listType":      "x-kubernetes-list-type",
+	"listTypeKey":   "x-kubernetes-list-type-keys",
+}
+
+// Enum values per extension
+var allowedExtensionValues = map[string][]string{
+	"x-kubernetes-patch-strategy": []string{"merge", "retainKeys"},
```
lavalamp: map[string]bool{} makes a loop below a 1-liner

pkg/generators/extension.go
```diff
@@ -0,0 +1,112 @@
...
+func (e Extension) validateAllowedValues() error {
+	// allowedValues not set means no restrictions on values.
+	allowedValues, exists := allowedExtensionValues[e.name]
+	if !exists {
+		return nil
+	}
+	// For each extension value, validate that it is allowed.
+	for _, value := range e.values {
+		if !valueInArray(value, allowedValues) {
+			return fmt.Errorf("%s: value (%s) not allowed", e.name, value)
```
lavalamp: Will users be able to figure out what source file/line produced this error? Please test the error output.

Also it's nice to list the allowed values.

pkg/generators/extension.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	return nil
+}
+
+func valueInArray(a string, array []string) bool {
```
lavalamp: I don't know if we're allowed to import it, but the sets.Strings type does nearly all of these helper things.

pkg/generators/openapi.go
seans3: I was hoping to address this in a future PR. I was hoping to keep this PR as small and focused as possible. I'll combine the two tags types in this PR if you like.

pkg/generators/openapi_test.go
seans3: I'll address this formatting issue (for the previous tests as well) in a future PR.

pkg/generators/openapi_test.go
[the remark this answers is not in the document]
seans3: Done

pkg/generators/openapi_test.go
seans3: Fixed with valid list type.

pkg/generators/openapi.go
seans3: As we agreed offline, I will combine the two parsing mechanisms.

pkg/generators/extension.go
seans3: Fixed.

pkg/generators/extension.go
seans3: Seems like a premature optimization. The size of the lists are small. And it will be easier to print out the allowed values. We can revisit this in another PR.

pkg/generators/openapi_test.go
seans3: Fixed. Used valid listType and put this extension on a list. The old tests are not on a list. Those old tests will be fixed when "gofmt" makes the output more readable in a future PR.

pkg/generators/extension.go
seans3: We probably don't want to make this repository dependent on Kubernetes.

pkg/generators/extension.go
seans3: Error output is tested. Allowed values are now added to error.

It looks like incorporating the source file/line in the error would require some refactoring. I suggest we add that in a subsequent PR.

pkg/generators/extension.go
lavalamp: Hm, we can actually get it from here: https://godoc.org/k8s.io/gengo/examples/set-gen/sets

Any issue with grabbing it from the examples directory?

pkg/generators/openapi_test.go
```diff
@@ -495,3 +502,62 @@ Dependencies: []string{
...
+package foo
+
+// Blah is a test.
+// +k8s:openapi-gen=true
+// +k8s:openapi-gen=x-kubernetes-type-tag:type_test
+type Blah struct {
+	// a member with a list type
+	// +listType=map
+	// +listMapKey=port
+	// +listMapKey=protocol
```
lavalamp: Why not `listMapKeys=port,protocol`?

pkg/generators/openapi_test.go
```diff
@@ -495,3 +502,62 @@ Dependencies: []string{
...
+Schema: spec.Schema{
+SchemaProps: spec.SchemaProps{
+Description: "Blah is a test.",
+Properties: map[string]spec.Schema{
+"WithListField": {
+VendorExtensible: spec.VendorExtensible{
+Extensions: spec.Extensions{
+"x-kubernetes-list-map-keys": []string{"port",
+"protocol",
+}"x-kubernetes-list-type": "map",
```
lavalamp: The test now exercises the bug I was referring to :)

pkg/generators/openapi.go
lavalamp: I'm still a bit confused by several things...

1. `tagExtensions` vs `extensions` is a confusing name combo. There should either be only one of them, or the name should indicate the reason for there being two.
2. It's still not clear why there are two for loops below and two sources?
3. Now that I have stared at this code a lot, do we have a good reason for not prefixing these new tags with the k8s:openapi-gen prefix?

pkg/generators/extension.go
seans3: The scope of this PR is already too large. I'll address this in another PR.

pkg/generators/openapi_test.go
seans3: We would have to make sure the separator character is not in any value. If we want to change it, let's do it in another small, focused PR.

pkg/generators/extension.go
```diff
@@ -0,0 +1,112 @@
...
+		extension := Extension{
+			name:   extensionName,
+			values: extensionValues,
+		}
+		// TODO: Validate the number of allowed values
+		// TODO: Validate the field type for this extension.
+		// Example: +listType is only allowed on an array.
+		if err := extension.validateAllowedValues(); err != nil {
+			// For now, only log the extension validation errors.
+			glog.V(2).Info(err)
```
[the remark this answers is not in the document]
seans3: It appears that using this mechanism would require some refactoring. I'll look into this in another PR.

pkg/generators/openapi_test.go
lavalamp: Commas are not going to be in any field names?

apelisse: I actually like this form, because it doesn't make it obvious to people that they can have multiple keys (which we certainly don't want to encourage)

pkg/generators/extension.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Enum values per extension
+var allowedExtensionValues = map[string][]string{
+	"x-kubernetes-patch-strategy": []string{"merge", "retainKeys"},
+	"x-kubernetes-list-type":      []string{"atomic", "set", "map"},
+}
+
+// Extension encapsulates information necessary to generate an OpenAPI extension.
+// TODO: Create interface for an extension validation function, and create
```
mbohlool: please add an issue or person for any TODO.

pkg/generators/extension.go
mbohlool: Not sure if I understand your comment here Daniel. Is this list a map from an extension to its allowed values? how map[string]bool can serve that purpose?

pkg/generators/extension.go
mbohlool: I agree that if we can find any library to have these helpers, it is better but have reservation on using an example repo for this.

pkg/generators/extension.go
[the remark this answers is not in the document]
mbohlool: If you are going to do any of these changes in another PR, please file an issue and add a TODO here with those issue numbers so we don't forget them.

pkg/generators/openapi_test.go
mbohlool: the generated code is not gofmt compliant and the scripts will run gofmt on them. if you format this string then you need to gofmt the output of generator (which is dynamic) too before comparing it with assert.Equal. I would say this is just too much to do for tests but feel free to find a solution for it if you want to spend time on it.

pkg/generators/openapi.go
```diff
@@ -416,61 +409,24 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
...
+	extensions := parseExtensions(CommentLines)
 	if len(extensions) == 0 {
 		return nil
 	}
+	// If any extensions exist, then emit code to create them.
 	g.Do("VendorExtensible: spec.VendorExtensible{\nExtensions: spec.Extensions{\n", nil)
 	for _, extension := range extensions {
-		g.Do("\"$.$\": ", extension.Name)
-		g.Do("\"$.$\",\n", extension.Value)
+		g.Do("\"$.$\": ", extension.name)
```
mbohlool: it would be nicer if Extension has a method to generate its code. something like:

```golang
for _, extension := range extensions {
  extension.generate(g)
}
```

seans3: This looks cleaner. But currently the Extension struct is a simple data object. It doesn't understand formatting or have any dependency on formatting objects. This type of function would add that dependency, and create code that is more tightly coupled.  But If we want to add this dependency, let's do it in a future PR.

mbohlool: What dependency are you talking about? It only needs a type writer. Anything else? I think this way is much cleaner and I prefer if you do it here. If you feel strong about doing it in another PR please add the TODO and create an issue for it.

seans3: @mbohlool Notice how simple, easy, and clean the unit tests are for Extension. Notice how the unit tests with the type writer are difficult to create, read, and understand. This is because the Extension struct has no formatting dependencies. Since the scope of this PR has already ballooned, we'll address this in a future PR.

mbohlool:
Discussed this PR with @lavalamp. We think it is best if we work in a feature branch. We still need to review changes in that branch but the code review would be checking for obvious bugs only. This will allow you guys to move faster. I create a branch called `feature-serverside-apply`. Please resend the PR to that branch.

pkg/generators/extension.go
```diff
@@ -0,0 +1,125 @@
...
+// Enum values per extension
+var allowedExtensionValues = map[string]sets.String{
+	"x-kubernetes-patch-strategy": sets.NewString("merge", "retainKeys"),
+	"x-kubernetes-list-type":      sets.NewString("atomic", "set", "map"),
+}
+
+// Extension encapsulates information necessary to generate an OpenAPI extension.
+// TODO: Create interface for an extension validation function, and create
+// validation functions for different extensions.
+type Extension struct {
```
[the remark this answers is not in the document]
seans3: No--fixed.

lavalamp:
Can you squash? Or I guess I can squash while merging?

But if you squash can you take my username out of the commit message? Github emails me every time a commit like that gets rebased or forked or whatever.

PR #54, opened 2018-04-19
Allow concurrent readers of the spec in OpenAPIService

jennybuckley:
This PR adds a public function to the OpenAPIService to allow multiple processes to read the same underlying spec, while still allowing an entire new spec to set, and adds comments telling OpenAPIService users how to avoid race conditions.


jennybuckley:
orgSpec doesn't exist anymore

PR #56, opened 2018-04-20
Start putting up documentation for the Go IDL

apelisse:
This PR is the beginning of the documentation for the IDL. It's obviously not complete, the goal is to make sure we agree on the direction.

### DIFF
--- a/pkg/idl/doc.go
+++ b/pkg/idl/doc.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+The IDL package describes comment directives that may be applied to API types.
+*/
+package idl
+
+// ListType annotates a list to further describe how it works.
+// Currently, it can have 3 possible values: "atomic", "map", or "set".
+// Note that there is no default, and eventually the generation step
+// will fail if a list is found that doesn't have that tag. This tag
+// MUST be used on lists, or the generation step will fail.
+//
+// Atomic
+//
+// The tag looks like this:
+//  +listType=atomic
+//
+// Atomic is a list that should be entirely replaced when changed. It
+// should usually be treated as a scalar type. This tag can be used on
+// any type of list (struct, scalar, ...).
+//
+// Using that tag will generate the following OpenAPI extension:
+//  "x-kubernetes-list-type": "atomic"
+//
+// Map
+//
+// The tag looks like this:
+//  +listType=map
+//
+// These lists are like maps. Order is preserved upon merge. Using the
+// map tag on a non-struct will report an error during the generation
+// step.
+//
+// Using that tag will generate the following OpenAPI extension:
+//  "x-kubernetes-list-type": "map"
+//
+// Set
+//
+// The tag looks like this:
+//  +listType=set
+//
+// Sets are lists that can't have multiple times the same value. Each
+// value must be a scalar or an atomic list and can't be map, set or struct.
+//
+// Using that tag will generate the following OpenAPI extension:
+//  "x-kubernetes-list-type": "set"
+type ListType string
--- a/pkg/idl/listtype_test.go
+++ b/pkg/idl/listtype_test.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package idl_test
+
+// This example shows how you can use the listType map attribute and how
+// to specify a key that will be used to identify specific elements of
+// the list. The listMapKey attribute is used to specify that Name will
+// be used as the key of the map.
+func ExampleListType_map() {
+	type SomeAPI struct {
+		// +listType=map
+		// +listMapKey=name
+		elements []struct {
+			Name string
+
+			Value int
+		}
+	}
+}
+
+// This example shows how you can use the listType set attribute to
+// specify that this list should be treated as a set: items in the list
+// can't be duplicated.
+func ExampleListType_set() {
+	type SomeAPI struct {
+		// +listType=set
+		keys []string
+	}
+}
+
+// This example shows how you can use the listType atomic attribute to
+// specify that this list should be treated as a whole.
+func ExampleListType_atomic() {
+	type SomeAPI struct {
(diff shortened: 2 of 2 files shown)


apelisse:
@pwittrock @seans3  FYI
/assign @mbohlool @lavalamp

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
lavalamp: This boilerplate is missing some newlines.

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,49 @@
...
+		keys []string
+	}
+}
+
+// This example shows how you can use the listType atomic attribute to
+// specify that this list should be treated as a whole.
+func ExampleListType_atomic() {
+	type SomeAPI struct {
+		// +listType=atomic
+		elements []struct{}
```
lavalamp: Maybe an empty struct is not the clearest thing to use for an example.

pkg/idl/doc.go
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+/*
+The IDL package describes comment directives that may be applied to API types.
+*/
+package idl
+
+// ListType annotates a list to further describe how it works.
+// Currently, it can have 3 possible values: "atomic", "map", or "set". Note
+// that there is no default, and eventually the generation step should
```
lavalamp: s/should/will/

pkg/idl/doc.go
```diff
@@ -0,0 +1,46 @@
...
+// fail if a list is found that doesn't have that tag.
+//
+// Atomic
```
lavalamp: ```+listType=atomic```

And same for others: write exactly the line users will use, otherwise it is confusing to have capitals here, I think

lavalamp:
just nits

pkg/idl/doc.go
apelisse: Yeah, this is a section header, so I think it has to start with a capital letter and finish with no points if I want to make this a title. I could definitely specify the actual full tag in the section thouhg

lavalamp:
Then maybe `// Atomic (+listType=atomic)` or something?

lavalamp:
I guess this is pretty clear, actually.

/lgtm

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+limitations under the License.
+*/
+
+/*
+The IDL package describes comment directives that may be applied to API types.
+*/
+package idl
+
+// ListType annotates a list to further describe how it works.
+// Currently, it can have 3 possible values: "atomic", "map", or "set".
```
pwittrock: Drop currently?

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// Note that there is no default, and eventually the generation step
```
pwittrock: Drop eventually

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// will fail if a list is found that doesn't have that tag. This tag
+// MUST be used on lists, or the generation step will fail.
+//
+// Atomic
+//
+// The tag looks like this:
+//  +listType=atomic
+//
+// Atomic is a list that should be entirely replaced when changed. It
```
pwittrock: Maybe 

> Atomic lists will...when updated.

Instead of 

> Atomic is a list that should...when changed.

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// should usually be treated as a scalar type. This tag can be used on
```
pwittrock: To be consistent with https://www.ietf.org/rfc/rfc2119.txt

> This tag may

instead of 

> This tag can

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+The IDL package describes comment directives that may be applied to API types.
+*/
+package idl
+
+// ListType annotates a list to further describe how it works.
```
pwittrock: > to define how it is updated

pkg/idl/doc.go
pwittrock: s/can/may/

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// Currently, it can have 3 possible values: "atomic", "map", or "set".
+// Note that there is no default, and eventually the generation step
+// will fail if a list is found that doesn't have that tag. This tag
```
pwittrock: s/doesn't have that/missing the/

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+// This example shows how you can use the listType map attribute and how
+// to specify a key that will be used to identify specific elements of
+// the list. The listMapKey attribute is used to specify that Name will
+// be used as the key of the map.
+func ExampleListType_map() {
+	type SomeAPI struct {
+		// +listType=map
+		// +listMapKey=name
+		elements []struct {
```
pwittrock: This is confusing because the type is inlined, so it isn't clear whether it is on the field or the type.  Pull the struct out into its own type definition to make this clear.

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,55 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package idl_test
+
+// This example shows how you can use the listType map attribute and how
```
pwittrock: s/you can/to/

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,55 @@
...
+// to specify a key that will be used to identify specific elements of
```
pwittrock: s/that will be used//

pkg/idl/listtype_test.go
```diff
@@ -0,0 +1,55 @@
...
+// the list. The listMapKey attribute is used to specify that Name will
```
pwittrock: s/will be used as/ is/

pkg/idl/listtype_test.go
pwittrock: s/specific/identical/

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// Using that tag will generate the following OpenAPI extension:
+//  "x-kubernetes-list-type": "atomic"
+//
+// Map
+//
+// The tag looks like this:
+//  +listType=map
+//
+// These lists are like maps. Order is preserved upon merge. Using the
```
pwittrock: > These lists are like maps in that elements have a non-index key.

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// map tag on a non-struct will report an error during the generation
```
pwittrock: >  map tag on lists with non-struct elements will...

pkg/idl/listtype_test.go
lavalamp: Please just add a section explaining this. :)

apelisse: Yeah, obviously :-) I've just started with the `listType` attribute so far!

pkg/idl/doc.go
```diff
@@ -0,0 +1,62 @@
...
+// Using that tag will generate the following OpenAPI extension:
+//  "x-kubernetes-list-type": "map"
+//
+// Set
+//
+// The tag looks like this:
+//  +listType=set
+//
+// Sets are lists that can't have multiple times the same value. Each
+// value must be a scalar or an atomic list and can't be map, set or struct.
```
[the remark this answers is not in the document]
apelisse: I'll change that constraint for now. It might be true later, but I think we don't have to be that permissive right now.

PR #58, opened 2018-04-23
idl-doc: Further improvements

apelisse:


apelisse:
I've actually added much more changes.

pkg/idl/doc.go
```diff
@@ -14,49 +14,126 @@ See the License for the specific language governing permissions and
...
-// Using that tag will generate the following OpenAPI extension:
+// Using this tag will generate the following OpenAPI extension:
 //  "x-kubernetes-list-type": "set"
 type ListType string
+
+// ListMapKey annotates map lists by specifying the key used as the index of the map.
+//
+// This tag MUST only be used on lists that have the listType=map
+// attribute, or the generation step will fail. Also, the value
+// specified for this attribute must be a field, of a scalar type, of
```
pwittrock: > field, of a scalar type

scalar typed field

pkg/idl/doc.go
```diff
@@ -14,49 +14,126 @@ See the License for the specific language governing permissions and
...
+//  "x-kubernetes-map-type": "atomic"
+type MapType string
+
+// OpenAPIGen needs to be described.
+type OpenAPIGen string
+
+// Optional needs to be described.
+type Optional string
+
+// PatchStrategy needs to be described.
```
pwittrock: These will be deprecated I think

pwittrock:
lgtm

pkg/idl/doc.go
apelisse: Yeah I still want to describe them as such

apelisse:
Addressed the comment and fixed a couple of other bugs that I noticed.

pkg/idl/doc.go
```diff
@@ -14,49 +14,126 @@ See the License for the specific language governing permissions and
...
 // Set
 //
-// The tag looks like this:
+// Example:
 //  +listType=set
 //
-// Sets are lists that can't have multiple times the same value. Each
-// value must be a scalar or an atomic list and can't be map, set or struct.
+// Sets are lists that must not have multiple times the same value. Each
+// value must be a scalar.
```
lavalamp: a scalar (or another atomic type).

pkg/idl/listtype_test.go
```diff
@@ -16,40 +16,41 @@ limitations under the License.
...
+	}
 	type SomeAPI struct {
 		// +listType=map
 		// +listMapKey=name
-		elements []struct {
-			Name string
-
-			Value int
-		}
+		elements []SomeStruct
```
lavalamp: I would like more clarity that the tags are there on account of the list type and not on account of the field. Next PR?

pkg/idl/doc.go
```diff
@@ -14,49 +14,126 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*
-The IDL package describes comment directives that may be applied to API types.
-*/
+// The IDL package describes comment directives that may be applied to API types.
```
lavalamp: Types and fields. And we should say which tags are which.

lavalamp:
LGTM, can you squash?

pkg/idl/listtype_test.go
apelisse: Let's do next PR so that we don't mix too many changes together.

apelisse:
Fixed and squashed! Thanks Daniel, please merge when tests pass.

PR #59, opened 2018-04-23
Remove unnecessarily stored spec

jennybuckley:
We store a reference, orgSpec, to the spec.Swagger (go typed openapi spec), but we never serve it. once the specBytes we serve are generated from the orgSpec we don't need a reference to the original spec anymore

@mbohlool


apelisse:
Well, if this builds (and it does), I'm going to assume it's automatically correct.

PR #60, opened 2018-04-24
[WIP] Validate extension type

seans3:
* validates member (field) extensions
* encapsulate extension attributes into struct
* add simple extension methods: hasAllowedValues, allowedValues, hasKind, kind, validateType
* No unit tests yet

### DIFF
--- a/pkg/generators/extension.go
+++ b/pkg/generators/extension.go
@@ -27,18 +27,33 @@ import (
 
 const extensionPrefix = "x-kubernetes-"
 
-// Extension tag to openapi extension
-var tagToExtension = map[string]string{
-	"patchMergeKey": "x-kubernetes-patch-merge-key",
-	"patchStrategy": "x-kubernetes-patch-strategy",
-	"listType":      "x-kubernetes-list-type",
-	"listMapKey":    "x-kubernetes-list-map-keys",
+// extensionAttributes encapsulates common traits for particular extensions.
+type extensionAttributes struct {
+	xName         string
+	kind          types.Kind
+	allowedValues sets.String
 }
 
-// Enum values per extension
-var allowedExtensionValues = map[string]sets.String{
-	"x-kubernetes-patch-strategy": sets.NewString("merge", "retainKeys"),
-	"x-kubernetes-list-type":      sets.NewString("atomic", "set", "map"),
+// Extension tag to openapi extension attributes
+var tagToExtension = map[string]extensionAttributes{
+	"patchMergeKey": extensionAttributes{
+		xName: "x-kubernetes-patch-merge-key",
+		kind:  types.Slice,
+	},
+	"patchStrategy": extensionAttributes{
+		xName:         "x-kubernetes-patch-strategy",
+		kind:          types.Slice,
+		allowedValues: sets.NewString("merge", "retainKeys"),
+	},
+	"listMapKey": extensionAttributes{
+		xName: "x-kubernetes-list-map-keys",
+		kind:  types.Slice,
+	},
+	"listType": extensionAttributes{
+		xName:         "x-kubernetes-list-type",
+		kind:          types.Slice,
+		allowedValues: sets.NewString("atomic", "set", "map"),
+	},
 }
 
 // Extension encapsulates information necessary to generate an OpenAPI extension.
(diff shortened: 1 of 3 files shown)


pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
 	values []string // Example: [atomic]
 }
 
+func (e extension) hasAllowedValues() bool {
+	xAttrs, exists := tagToExtension[e.idlTag]
+	if !exists || xAttrs.allowedValues == nil {
```
apelisse: I'm sure you're more interested in feedback wrt the structure of the code more than this kind of nit, but yet, here I go: this function can be simplified to:
```
func (e extension) hasAllowedValues() bool {
   len(tagToExtension[e.idleTag].allowedValues) > 0
}
```

I'm not sure I'd suggest this extreme form, but you get the idea. Actually yeah, I think the intent is there.

pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
...
+		return false
+	}
+	return len(xAttrs.allowedValues) > 0
+}
+
+func (e extension) allowedValues() sets.String {
+	xAttrs, exists := tagToExtension[e.idlTag]
```
apelisse: somewhat same

pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
...
+	allowedValues := e.allowedValues()
 	if !allowedValues.HasAll(e.values...) {
 		return fmt.Errorf("%v not allowed for %s. Allowed values: %v",
 			e.values, e.idlTag, allowedValues.List())
 	}
 	return nil
 }
 
+func (e extension) hasKind() bool {
+	xAttrs, exists := tagToExtension[e.idlTag]
```
apelisse: same

pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
...
+	if !exists {
+		return false
+	}
+	return len(xAttrs.kind) > 0
+}
+
+func (e extension) kind() types.Kind {
+	xAttrs, exists := tagToExtension[e.idlTag]
```
apelisse: same

pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
...
+	if !exists {
+		return ""
```
apelisse: is empty string a kind?

pkg/generators/extension.go
```diff
@@ -48,24 +63,67 @@ type extension struct {
...
+	}
+	return xAttrs.kind
+}
+
+func (e extension) validateType(t types.Kind) error {
+	if !e.hasKind() {
```
apelisse: is this condition necessary? can `t` be "empty string" kind?

pkg/generators/extension_test.go
```diff
@@ -138,7 +138,7 @@ func TestMultipleTagExtensions(t *testing.T) {
 
 }
 
-func TestExtensionErrors(t *testing.T) {
+func TestParseExtensionErrors(t *testing.T) {
```
apelisse: more code but less tests. I'm assuming that's because it's WIP. Nevermind :-)

pkg/generators/extension.go
```diff
@@ -27,18 +27,33 @@ import (
...
+	kind          types.Kind
+	allowedValues sets.String
 }
 
-// Enum values per extension
-var allowedExtensionValues = map[string]sets.String{
-	"x-kubernetes-patch-strategy": sets.NewString("merge", "retainKeys"),
-	"x-kubernetes-list-type":      sets.NewString("atomic", "set", "map"),
+// Extension tag to openapi extension attributes
+var tagToExtension = map[string]extensionAttributes{
```
apelisse: The number of conditions has increased in that pull-request. I think this PR goes back to what we were initially discussing: we should have some sort of virtual dispatch with `extensionAttributes` being an interface, with different types of attributes. 

For example, things that are not looking as targeted as they could be:
- `allowedValues` is optional rather than mandatory, creating a new corner-case
- it's not clear if `kind` is mandatory or not, but that creates a new corner-case
- I'm expecting a few more corner cases when you'll want to write more specific `kind` validation
- you still have a list of values in `extension` where we might not have a list (a bunch of new corner-cases)
- you have a list of strings in `extension`, even though they might not always be

For now it's fine (though it's less and less), you can keep going in that direction, but my opinion is that it's going to become more and more obvious that many of these things should be handled by a virtual dispatch.

pkg/generators/extension.go
apelisse: you could also write it `return e.kind() != ""`.

PR #61, opened 2018-04-26
Type validation for member/field extensions

seans3:
* validates member (field) extensions
* encapsulate extension attributes into struct
* add simple extension methods: hasAllowedValues, allowedValues, hasKind, kind, validateType


pkg/generators/extension.go
```diff
@@ -48,24 +63,52 @@ type extension struct {
...
+func (e extension) hasAllowedValues() bool {
+	return tagToExtension[e.idlTag].allowedValues.Len() > 0
+}
+
+func (e extension) allowedValues() sets.String {
+	return tagToExtension[e.idlTag].allowedValues
+}
+
+func (e extension) hasKind() bool {
+	return len(tagToExtension[e.idlTag].kind) > 0
```
apelisse: `!= ""` ?

pkg/generators/extension.go
```diff
@@ -108,15 +153,15 @@ func parseExtensions(comments []string) ([]extension, []error) {
...
 		values := tagValues[idlTag]
 		e := extension{
-			idlTag: idlTag, // listType
-			xName:  xName,  // x-kubernetes-list-type
-			values: values, // [atomic]
+			idlTag: idlTag,       // listType
+			xName:  xAttrs.xName, // x-kubernetes-list-type
+			values: values,       // [atomic]
 		}
 		if err := e.validateAllowedValues(); err != nil {
```
julianvmodesto: Remove validation in parse?

pkg/generators/extension.go
```diff
@@ -115,15 +123,15 @@ func parseExtensions(comments []string) ([]extension, []error) {
...
 			continue
 		}
 		values := tagValues[idlTag]
 		e := extension{
-			idlTag: idlTag, // listType
-			xName:  xName,  // x-kubernetes-list-type
-			values: values, // [atomic]
+			idlTag: idlTag,       // listType
+			xName:  xAttrs.xName, // x-kubernetes-list-type
+			values: values,       // [atomic]
```
lavalamp: Consider removing the trailing comments here next time you touch this, the names should be clear enough now?

pkg/generators/extension.go
```diff
@@ -63,17 +63,25 @@ type extension struct {
 	values []string // Example: [atomic]
 }
 
+func (e extension) hasAllowedValues() bool {
```
lavalamp: I think the relevant detail is whether the allowed values are known ahead of time or not, not whether there are allowed values. Tricky to think of a better name. [just a comment, no action requested]

pkg/generators/extension.go
```diff
@@ -71,6 +71,10 @@ func (e extension) allowedValues() sets.String {
 	return tagToExtension[e.idlTag].allowedValues
 }
 
+func (e extension) kind() types.Kind {
```
lavalamp: This is talking about the kind of go type the extension applies to? The name is a bit confusing but I don't have a better suggestion. Maybe change in a follow-up if you think of something better.

pkg/generators/extension.go
lavalamp: > 0 is preferred by many (not me, but I'm in the minority)

pkg/generators/extension_test.go
```diff
@@ -361,8 +361,17 @@ func TestExtensionKind(t *testing.T) {
 		},
 	}
 	for _, test := range methodTests {
-		if test.kind != test.e.kind() {
-			t.Errorf("%v: kind() expected (%v), but received: %v", test.e, test.kind, test.e.kind())
+		if len(test.kind) > 0 {
```
lavalamp: It feels better to add something to the test structure rather than copy-paste the logic we're trying to test. Up to you if you want to fix.

pkg/generators/extension.go
lavalamp: I didn't follow this comment, can you expand? :)

lavalamp:
this lgtm, please squash.

pkg/generators/extension.go
julianvmodesto: Oops, I mean remove `validateAllowedValues` from `parseExtensions`.

This is in reference to the comment on `parseExtensions`:
```Validating extensions is performed separately.```

https://github.com/seans3/kube-openapi/blob/69890351242d7509fc807275114ab78514e62b32/pkg/generators/openapi.go#L444-L448

Because validation actually is separate from parsing now:
https://github.com/seans3/kube-openapi/blob/69890351242d7509fc807275114ab78514e62b32/pkg/generators/extension.go#L128-L132

lavalamp: oh, good point. @seans3 is the comment at the top of this function wrong?

seans3: That's a good catch @julianvmodesto . `validateAllowedValues()` is duplicated. I could have sworn I removed it from the parsing function. As you probably noticed, `parseExtensions()` should only return the parse errors, and the `validateMemberExtensions()` returns the validation errors of well-formed extensions.

Issue #66, opened 2018-05-04
Aggregator should not fail on OpenAPI specs with no definition or Path


mbohlool:
Recently I encountered a bug that aggregator failed because an OpenAPI spec did not have a Path item (null pointer exception). We should make sure all nullable fields are accounted for in the aggregation.

ref: https://github.com/kubernetes/kube-openapi/blob/39a7bf85c140f972372c2a0d1ee40adbf0c8bfe1/pkg/aggregator/aggregator.go#L145

mbohlool:
Look like path is a required field in OpenAPI spec. We should, however, return an error here when Paths is nil (interesting that Paths is a pointer when it shouldn't be).

PR #70, opened 2018-05-09
cherry-picking #64 and #67 to release-1.8

mbohlool:
Cherry-pick 1.8 branch with two commits fixing #66

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -142,6 +142,9 @@ func (s *referenceWalker) walkOperation(op *spec.Operation) {
 }
 
 func (s *referenceWalker) Start() {
+	if s.root.Paths == nil {
+		return
+	}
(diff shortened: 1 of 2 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -239,6 +246,13 @@ func MergeSpecs(dest, source *spec.Swagger) error {
 
 func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConflicts bool) (err error) {
 	specCloned := false
+	// Paths may be empty, due to [ACL constraints](http://goo.gl/8us55a#securityFiltering).
+	if source.Paths == nil {
```
mbohlool: Now that I am thinking more about this, if the source path is empty, we should not do anything. It is true that in current implementation, we do merge definitions (which we should fix) but when a source spec does not have any path, that means none of the definitions are used thus we should not do anything. just return.

PR #73, opened 2018-05-15
More listType validations

seans3:


pkg/generators/extension_test.go
```diff
@@ -424,29 +449,81 @@ func TestValidateMemberExtensions(t *testing.T) {
...
 	}{
-		// Test single member extension
+		// List type extension must be defined on a slice.
 		{
-			extensions: []extension{patchStrategyExtension},
+			extensions: []extension{listTypeAtomicExtension},
 			member:     mapField,
 		},
-		// Test multiple member extensions
+		// Map keys defined, but missing listType=map
```
apelisse: There are probably many other use-cases here that are missing. You could almost create a more thorough list, and then use TDD, and maybe comment out those that are not implemented yet.

pkg/generators/extension.go
```diff
@@ -34,25 +34,38 @@ type extensionAttributes struct {
 	allowedValues sets.String
 }
 
+const LIST_TYPE = "listType"
+const LIST_MAP_KEY = "listMapKey"
```
mbohlool: Should these names be camel-case?

mbohlool: Also do they need to be public? (i.e. start with capital letter)

apelisse:
@seans3 @lavalamp Are we dropping this?

PR #74, opened 2018-05-15
Allow openapi definitions to be built for a single resource

jennybuckley:
This PR allows for the openapi definitions to be built for a single resource. Needed for server side apply.

**Implementation:**

Split the init([]*restful.WebService) function into three functions:
- init(): runs before the swagger spec is built
- buildPaths([]*restful.WebService): The function which actually builds the swagger spec
- finalizeSwagger(): runs after the spec is built

This way we can substitute a different function in for buildPaths, which doesn't need to take a web service as an argument.

cc @mbohlool @apelisse @seans3

### DIFF
--- a/pkg/builder/openapi.go
+++ b/pkg/builder/openapi.go
@@ -45,6 +45,16 @@ type openAPI struct {
 
 // BuildOpenAPISpec builds OpenAPI spec given a list of webservices (containing routes) and common.Config to customize it.
 func BuildOpenAPISpec(webServices []*restful.WebService, config *common.Config) (*spec.Swagger, error) {
+	o := newOpenAPI(config)
+	err := o.buildPaths(webServices)
+	if err != nil {
+		return nil, err
+	}
+	return o.finalizeSwagger()
+}
+
+// newOpenAPI sets up the openAPI object so we can build the spec.
+func newOpenAPI(config *common.Config) openAPI {
 	o := openAPI{
 		config: config,
 		swagger: &spec.Swagger{
@@ -56,16 +66,6 @@ func BuildOpenAPISpec(webServices []*restful.WebService, config *common.Config) 
 			},
 		},
 	}
-
-	err := o.init(webServices)
-	if err != nil {
-		return nil, err
-	}
-
-	return o.swagger, nil
-}
-
-func (o *openAPI) init(webServices []*restful.WebService) error {
 	if o.config.GetOperationIDAndTags == nil {
 		o.config.GetOperationIDAndTags = func(r *restful.Route) (string, []string, error) {
 			return r.Operation, nil, nil
@@ -83,22 +83,25 @@ func (o *openAPI) init(webServices []*restful.WebService) error {
 	if o.config.CommonResponses == nil {
 		o.config.CommonResponses = map[int]spec.Response{}
 	}
-	err := o.buildPaths(webServices)
-	if err != nil {
-		return err
-	}
+	return o
+}
+
+// finalizeSwagger is called after the spec is built and returns the final spec.
+// NOTE: finalizeSwagger also make changes to the final spec, as specified in the config.
+func (o *openAPI) finalizeSwagger() (*spec.Swagger, error) {
 	if o.config.SecurityDefinitions != nil {
 		o.swagger.SecurityDefinitions = *o.config.SecurityDefinitions
 		o.swagger.Security = o.config.DefaultSecurity
 	}
 	if o.config.PostProcessSpec != nil {
+		var err error
 		o.swagger, err = o.config.PostProcessSpec(o.swagger)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 
-	return nil
+	return o.swagger, nil
 }
 
 func getCanonicalizeTypeName(t reflect.Type) string {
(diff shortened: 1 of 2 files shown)


pkg/builder/openapi.go
```diff
@@ -45,7 +45,46 @@ type openAPI struct {
 
 // BuildOpenAPISpec builds OpenAPI spec given a list of webservices (containing routes) and common.Config to customize it.
 func BuildOpenAPISpec(webServices []*restful.WebService, config *common.Config) (*spec.Swagger, error) {
-	o := openAPI{
+	o := newOpenAPI(config)
+
+	err := o.init()
```
apelisse: Nit: It's a private struct so it doesn't matter that much, but you've created an implicit dependency in the order in which these functions need to be called. Can you call `init` in `newOpenAPI`, since you already have a sort of constructor? 

`finishSpec` is also suffering from these syndroms, you need to call it before you can get the `o.swagger`, if you forget to call it then, who knows what will happen. Can we "finishSpec" before (like in init) or do we need all the data before? Then you would have:

```
o, err := newOpenAPIConfig(config)
// handle err
o.addPath(webservices)
return o.Swagger() // But this function should be const.
```

pkg/builder/openapi.go
```diff
@@ -45,7 +45,46 @@ type openAPI struct {
...
+	err = o.finishSpec()
+	if err != nil {
+		return nil, err
+	}
+
+	return o.swagger, nil
+}
+
+// BuildOpenAPISpecForResource builds a partial OpenAPI spec given a sample object and common.Config to customize it.
+func BuildOpenAPISpecForResource(model interface{}, config *common.Config) (*spec.Swagger, error) {
```
apelisse: Maybe you should split into different PRs or at least commits when you're doing different things (refactor AND add new methods)

pkg/builder/openapi.go
jennybuckley: Can definitely call init in newOpenAPI, but we do need the spec to be built before calling finishSpec (it contains things like a postProcessSpec function, which is defined by the openapiconfig)

jennybuckley: Would a function like this be any better?
```
func createSpec(config common.OpenAPIConfig, buildSwaggerFunc func(o *openAPI)) *spec.Swagger {
o := openAPI{
...
}
o.init()
buildSwaggerFunc(o)
o.finish()
return o.swagger
}
```
(plus some error handling)

apelisse: ~~Meh, I don't know if you have many options here. It looks like the `openAPI` struct is having too many side-effects already. Let's see how we can make things better without changing everything ...~~

~~I'm tempted to think that these methods in general should be free-functions instead (because they don't really maintain a cohesive state for the struct anyway).~~

Yeah, maybe you should leave it as it is. :-)

jennybuckley: Okay, and I added a comment documenting the requirement to call finish before returning the spec.

jennybuckley:
@seans3 updated to add another unit test, the coverage is now up from 72 to 74%

pkg/builder/openapi.go
```diff
@@ -83,22 +95,25 @@ func (o *openAPI) init(webServices []*restful.WebService) error {
...
-	err := o.buildPaths(webServices)
-	if err != nil {
-		return err
-	}
+	return o
+}
+
+// getSwagger is called after the spec is built and returns the final spec.
+// NOTE: getSwagger also make changes to the final spec, as specified in the config.
+func (o *openAPI) getSwagger() (*spec.Swagger, error) {
```
seans3: Consider the name "finalizeSwagger" to emphasize that the swagger is modified before being returned

pkg/builder/openapi.go
```diff
@@ -45,6 +45,28 @@ type openAPI struct {
 
 // BuildOpenAPISpec builds OpenAPI spec given a list of webservices (containing routes) and common.Config to customize it.
 func BuildOpenAPISpec(webServices []*restful.WebService, config *common.Config) (*spec.Swagger, error) {
+	o := newOpenAPI(config)
+	err := o.buildPaths(webServices)
```
seans3: This looks much cleaner, since the "buildPaths" is not lost in the "init"

pkg/builder/openapi.go
mbohlool: It is actually not valid if it does not have paths.

pkg/builder/openapi.go
jennybuckley: Done.
Sorry, can you lgtm again? I had to recommit.

pkg/builder/openapi.go
```diff
@@ -45,6 +45,28 @@ type openAPI struct {
...
+	if err != nil {
+		return nil, err
+	}
+	return o.finalizeSwagger()
+}
+
+// BuildOpenAPIDefinitionsForResource builds a partial OpenAPI spec given a sample object and common.Config to customize it.
+func BuildOpenAPIDefinitionsForResource(model interface{}, config *common.Config) (*spec.Swagger, error) {
```
mbohlool: Can we return definitions instead of spec.Swagger here? the return spec is not a valid spec and hence we should not return it in my opinion.

jennybuckley: Okay

PR #76, opened 2018-05-29
proto: Support/parse OneOf types in OpenAPI

apelisse:
Currently implementation should be 100% backward compatible. The code
doesn't have to support oneof just yet, as oneof will be converted to
normal types automatically.

### DIFF
--- a/pkg/util/proto/testing/swagger_oneof.json
+++ b/pkg/util/proto/testing/swagger_oneof.json
@@ -1,0 +1,49 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Kubernetes",
+    "version": "v1.8.0"
+  },
+  "paths": {},
+  "definitions": {
+    "A": {
+      "description": "The first letter of the alphabet.",
+      "type": "object",
+      "properties": {
+        "apple": {
+          "description": "The round fruit of a tree of the rose family, which typically has thin red or green skin and crisp flesh. Many varieties have been developed as dessert or cooking fruit or for making cider.",
+          "type": "string"
+        },
+        "apricot": {
+          "description": "A juicy, soft fruit, resembling a small peach, of an orange-yellow color.",
+          "type": "string"
+        },
+        "fruit": {
+          "description": "The sweet and fleshy product of a tree or other plant that contains seed and can be eaten as food.",
+          "type": "string"
+        }
+      },
+      "x-kubernetes-oneof": "fruit",
+      "x-kubernetes-group-version-kind": [
+        {
+          "group": "alphabet",
+          "kind": "letter",
+          "version": "roman"
+        }
+      ]
+    }
+  },
+  "securityDefinitions": {
+    "BearerToken": {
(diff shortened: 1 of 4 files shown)


pkg/util/proto/testing/swagger_oneof.json
```diff
@@ -0,0 +1,49 @@
...
+        "apricot": {
+          "description": "A juicy, soft fruit, resembling a small peach, of an orange-yellow color.",
+          "type": "string"
+        },
+        "fruit": {
+          "description": "The sweet and fleshy product of a tree or other plant that contains seed and can be eaten as food.",
+          "type": "string"
+        }
+      },
+      "x-kubernetes-oneof": "fruit",
```
lavalamp: I should have caught this in your doc, sorry! This is super confusing, not 1:1 with v3, and makes the discriminator mandatory. We should split this into at least two extensions: `x-kubernetes-oneof: true/false` (lines up with openapi v3's one of, hopefully), and some TBD way of indicating the discriminator, it's not obvious to me how that should work.

pkg/util/proto/testing/swagger_oneof.json
```diff
@@ -0,0 +1,49 @@
...
+          "description": "The round fruit of a tree of the rose family, which typically has thin red or green skin and crisp flesh. Many varieties have been developed as dessert or cooking fruit or for making cider.",
+          "type": "string"
+        },
+        "apricot": {
+          "description": "A juicy, soft fruit, resembling a small peach, of an orange-yellow color.",
+          "type": "string"
+        },
+        "fruit": {
+          "description": "The sweet and fleshy product of a tree or other plant that contains seed and can be eaten as food.",
+          "type": "string"
```
lavalamp: I think the discriminator marker should maybe go here. This field is super confusing. I spent longer than I'd care to say trying to figure out why apple, apricot, and fruit were in the same list.

apelisse: Yeah, the more I think about it, the more I disagree (sorry :-)). The discriminator is really not a proprety of the field, but a property of the struct that holds it.

The struct is a oneof that will discriminate on field Y. OpenAPI v3 has a "discriminator" field that gives the property, and it's at the "oneof" level.

lavalamp:
The union "has a" discriminator. The field which is the discriminator "is
a" discriminator. It is more straightforward & understandable to make the
IDL document "is a" rather than "has a" relationships.

If you don't buy that, I really will insist that it has to be this:

```
x-kubernetes-oneof=true
x-kubernetes-oneof-discriminator=fruit
```

and not `x-kubernetes-oneof=fruit`; this is super not obvious, because the
concept of a discriminator is not obvious.

apelisse:
> ```
x-kubernetes-oneof=true
x-kubernetes-oneof-discriminator=fruit
```

Describing this in the document right now.

lavalamp:
Could also do `x-kubernetes-type=oneof` or something like that.

PR #82, opened 2018-06-13
Typo fix: identifer -> identifier/mulitple -> multiple

mirake:


philips:
Is mentioned in #81 I think these files are generated but I can't figure out from where...

cc @lavalamp @mbohlool

PR #84, opened 2018-06-20
include descriptions in types with custom defs

liggitt:
Include description so that custom types can be documented as well.

See in action in https://github.com/kubernetes/kubernetes/pull/65256

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -383,8 +383,9 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 		if hasOpenAPIDefinitionMethods(t) {
 			g.Do("return $.OpenAPIDefinition|raw${\n"+
 				"Schema: spec.Schema{\n"+
-				"SchemaProps: spec.SchemaProps{\n"+
-				"Type:$.type|raw${}.OpenAPISchemaType(),\n"+
+				"SchemaProps: spec.SchemaProps{\n", args)
+			g.generateDescription(t.CommentLines)
+			g.Do("Type:$.type|raw${}.OpenAPISchemaType(),\n"+
 				"Format:$.type|raw${}.OpenAPISchemaFormat(),\n"+
 				"},\n"+
 				"},\n"+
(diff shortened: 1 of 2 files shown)


pkg/generators/openapi_test.go
```diff
@@ -378,6 +378,7 @@ func TestCustomDefs(t *testing.T) {
 	callErr, funcErr, assert, callBuffer, funcBuffer := testOpenAPITypeWriter(t, `
 package foo
 
+// Blah is a custom type
 type Blah struct {
```
sttts: have you tried to add a `json:"-"` tag? Does kube-openapi cope with that?

liggitt: not sure what you're asking... this is testing a type that does its own custom marshaling, and so overrides the type/format presented in openapi. this PR is allowing that type to still include documentation.

liggitt:
/assign @seans3

this is needed for a 1.11 kube/kube PR

pkg/generators/openapi.go
```diff
@@ -383,8 +383,9 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 		if hasOpenAPIDefinitionMethods(t) {
 			g.Do("return $.OpenAPIDefinition|raw${\n"+
 				"Schema: spec.Schema{\n"+
-				"SchemaProps: spec.SchemaProps{\n"+
-				"Type:$.type|raw${}.OpenAPISchemaType(),\n"+
+				"SchemaProps: spec.SchemaProps{\n", args)
+			g.generateDescription(t.CommentLines)
```
liggitt: this is copying what we're already doing for non-custom types on line 395

pkg/generators/openapi_test.go
sttts: I meant if you create a field `Blah Blah 'json:"-"'`. Will kube-openapi ignore that json tag? It's a general question, not so much about this PR.

liggitt: ah, I don't know, I didn't try

PR #85, opened 2018-06-20
Server-side OpenAPI fixes (RawExtension & type:object for structs)

luksa:
DO NOT MERGE until v1.12, because clients v1.10 and older break when they encounter a server with these changes. A v1.11+ client won't break.

This PR fixes two things:
- adds type:object to all OpenAPI definitions created from structs
- prevents the RawExtension type from being included in the OpenAPI schema by annotating it with `+k8s:openapi-gen:schema-type-format=object,`

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -393,6 +393,7 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 		}
 		g.Do("return $.OpenAPIDefinition|raw${\nSchema: spec.Schema{\nSchemaProps: spec.SchemaProps{\n", args)
 		g.generateDescription(t.CommentLines)
+		g.Do("Type: []string{\"object\"},\n", nil)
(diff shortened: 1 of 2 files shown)


apelisse:
@seans3 Can you please take a look at this?

ash2k:
/reopen
/remove-lifecycle rotten
/lifecycle frozen

I think it'd be great to get this feature in.

apelisse:
I'm suggesting we close this now... :-)

ash2k:
It looks like the `type: object` part was done in https://github.com/kubernetes/kube-openapi/pull/120. I'm not sure about `RawExtension` - I remember it caused me some pain, but I don't remember what exactly was wrong and hence I don't know if there is something that needs to be fixed.

ash2k:
Ok, looks like the second part was https://github.com/kubernetes/kubernetes/issues/55890 which was fixed in https://github.com/kubernetes/kubernetes/pull/80773

I've looked at the schema and it looks fine to me. Let's close.

/close

PR #102, opened 2018-09-20
Move getCanonicalTypeName to util package and export it

jennybuckley:
To be used by https://github.com/kubernetes/kubernetes/pull/68576


seans3:
This function seems perfect for a unit test. Should we add one?

jennybuckley:
@seans3 I added a unit test, does this seem good to merge now?

PR #104, opened 2018-09-25
Sort API rule violations before reporting

roycaihw:
The API rule violations reported by openapi-gen are in format of

 - `{rule-name},{package},{type},{(optional) field}`

The `k8s.io/gen` performs sorting on package level and type level when walking through the input packages, but currently the API rule violations reported from openapi-gen are un-sorted on field level and rule-name level. 

The partially-sorted report file makes it hard for API authors to update the exception list when necessary, e.g. https://github.com/kubernetes/kubernetes/pull/60195#discussion_r204056998, https://github.com/kubernetes/kubernetes/pull/66722

This PR makes openapi-gen sort the violations before reporting to a file or stderr. As a followup I will add hack script in k/k for API authors to update the exception list automatically instead of manually.

/assign @lavalamp


test/integration/pkg/generated/openapi_generated.go
```diff
@@ -1,3 +1,5 @@
+// +build !ignore_autogenerated
```
apelisse: This file used to be kept mostly empty. I honestly believe this is a bug, it shouldn't have to be committed at all.

test/integration/testdata/dummytype/alpha.go
```diff
@@ -0,0 +1,27 @@
+package dummytype
```
apelisse: Do we need the new package? Maybe I'd have put everything together

test/integration/testdata/dummytype/alpha.go
```diff
@@ -0,0 +1,27 @@
...
+// checker. The API rule violations are in format of:
+//
+// `{rule-name},{package},{type},{(optional) field}`
+//
+// The checker should sort the violations before
+// reporting to a file or stderr.
+
+// +k8s:openapi-gen=true
+type Foo struct {
+	B string
```
apelisse: I find the letters a little bit confusing, maybe it's just me

test/integration/pkg/generated/openapi_generated.go
roycaihw: I have similar feeling. I don't know the reason for committing this file at all.

I first ran the commands in README.md and got the file updated like in this PR.

When I then ran `go test -v .` to test locally, the file was updated again: 
```
```

I didn't look deep into it because it doesn't fail the test. Maybe we should just have the integration test remove the file as cleanup, and also add a command in README.md to remove the file, so that it won't be committed. WDYT?

test/integration/testdata/dummytype/alpha.go
roycaihw: I'd like to test the sorting behavior on package level, e.g.
- `-i "./testdata/listtype,./testdata/dummytype"`
- `-i "./testdata/dummytype,./testdata/listtype"`

The violations from `dummytype` should come first in report.

It turns out that `k8s.io/gen` handles the package-level sorting when walking through the package, but I think it's still good to test the behavior.

apelisse: Good point! Maybe add a comment? Or maybe it's just me ...

test/integration/testdata/dummytype/alpha.go
roycaihw: I agree. Especially the `CD` can be confusing. I can add some comments in the type def to document that the letters are just for testing the behavior of sorting the parameters relatively and do not take any real meaning. Do you have any other suggestions?

test/integration/testdata/dummytype/alpha.go
roycaihw: Will do

test/integration/testdata/dummytype/alpha.go
apelisse: I'd just come-up with random names

roycaihw:
I tried to remove and gitignore `test/integration/pkg/generated/openapi_generated.go`. It turns out that it's a dependency of the [builder](https://github.com/kubernetes/kube-openapi/blob/e3762e86a74c878ffed47484592986685639c2cd/test/integration/builder/main.go#L29). Removing the file will fail the [travis job](https://travis-ci.org/kubernetes/kube-openapi/jobs/433226148) because it needs to download the packages required to build test (`go get -t`). 

The checked-in file is mostly empty because it only serves the `go get -t`. The actual content will be re-generated by `openapi-gen` during the test. I reverted the file to the empty state, but it's not clear to me how to prevent future PR from changing the file.

@apelisse I think the two-step generation is annoying but understandable, because after all some of the openapi spec generation need to happen in kube-apiserver to reflect the actual APIs served, and we have to compile some information (Go type definition / partial openapi spec) into the binary. I'd like to see if we can have a one-step generation in the future. 

The other comments addressed. PTAL

apelisse:
@roycaihw Yeah, I know ... I'd still like to see all these steps merged into one step so that we wouldn't have to commit the intermediary files.

roycaihw:
As chatted offline with @apelisse, re-apply LGTM after squashed and tests green.

/lgtm

PR #105, opened 2018-10-08
Allow using top-level Responses Definitions field in openapi v2

roycaihw:
This PR allows setting a `ResponseDefinitions` field in common config, which will be propagated to the top-level [Responses Definitions field](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields) during building. From the v2 specification, the field is:
> An object to hold responses that can be used across operations. This property does not define global responses for all operations.

The idea comes from the k8s issue https://github.com/kubernetes/kubernetes/issues/69014. We want to document all the possible error codes that apiserver may respond with. Presumably we will add 10+ responses to every operations under every paths. Since all the error responses have same schema in body (metav1.Status), we can put the full response definition in the top-level Responses Definitions field, and have the operation-level responses refer to it, to make the spec size smaller.

Note that `go-openapi` doesn't verify if the referred definition exists when [creating a Response](https://github.com/go-openapi/spec/blob/5bae59e25b21498baea7f9d46e9c147ec106a42e/response.go#L89-L94). It's up to user to set the definition first and write the refURI properly.

The next step will be changing k8s to set `ResponseDefinitions` and `CommonResponses` when [completing the openapi Config](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/config.go#L379-L388).

/assign @mbohlool 
cc @lswith

### DIFF
--- a/pkg/builder/openapi.go
+++ b/pkg/builder/openapi.go
@@ -91,6 +91,7 @@ func newOpenAPI(config *common.Config) openAPI {
 			SwaggerProps: spec.SwaggerProps{
 				Swagger:     OpenAPIVersion,
 				Definitions: spec.Definitions{},
+				Responses:   config.ResponseDefinitions,
 				Paths:       &spec.Paths{Paths: map[string]spec.PathItem{}},
 				Info:        config.Info,
 			},
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -59,6 +59,12 @@ type Config struct {
 	// will show up as ... "responses" : {"default" : $DefaultResponse} in the spec.
 	DefaultResponse *spec.Response
 
+	// ResponseDefinitions will be added to "responses" under the top-level swagger object. This is an object
+	// that holds responses definitions that can be used across operations. This property does not define
+	// global responses for all operations. For more info please refer:
+	//     https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields
+	ResponseDefinitions map[string]spec.Response
+
 	// CommonResponses will be added as a response to all operation specs. This is a good place to add common
 	// responses such as authorization failed.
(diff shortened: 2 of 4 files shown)


mbohlool:
can this be done with PostProcessSpec in the same config?

roycaihw:
@mbohlool Yes, it is achievable by using `PostProcessSpec`. I believe go-openapi (and kube-openapi) allows responses being referred prior to them being defined.  

One heads-up is that in k/k we have multiple places setting the `OpenAPIConfig`: 

- In [cmd/kube-apiserver/app/server.go](https://github.com/kubernetes/kubernetes/blob/851444f382adbc43942c4c34053fab1ec59cda52/cmd/kube-apiserver/app/server.go#L462): we set `OpenAPIConfig.PostProcessSpec` to add removed definitions for backward compatibility, after the `DefaultOpenAPIConfig` is created. 

- In [staging/src/k8s.io/apiserver/pkg/server/config.go](https://github.com/kubernetes/kubernetes/blob/851444f382adbc43942c4c34053fab1ec59cda52/staging/src/k8s.io/apiserver/pkg/server/config.go#L385-L391): the `genericapiserver.Config.Complete()` sets `OpenAPIConfig.CommonResponses` to add HTTP 401 response

which kind of makes sense to me because the backwards-compatibility is kube-apiserver specific therefore belongs to `cmd`; and 401/CommonResponses is our API convention that we want to promote to `staging` for others to vendor/consume. 

My opinion is that response definitions should live in `staging` as well. But if we are going to use `PostProcessSpec` to add response definitions, I'd probably put both [removed definitions](https://github.com/kubernetes/kubernetes/blob/851444f382adbc43942c4c34053fab1ec59cda52/cmd/kube-apiserver/app/server.go#L742) and response definitions in `cmd`. WDYT?

PR #106, opened 2018-10-15
Add linting test for omitempty

liggitt:
Would have caught https://github.com/kubernetes/kubernetes/pull/69694

### DIFF
--- a/pkg/generators/rules/omitempty_match_case.go
+++ b/pkg/generators/rules/omitempty_match_case.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rules
+
+import (
+	"reflect"
+	"strings"
+
+	"k8s.io/gengo/types"
+)
+
+// OmitEmptyMatchCase implements APIRule interface.
+// "omitempty" must appear verbatim (no case variants).
+type OmitEmptyMatchCase struct{}
+
+func (n *OmitEmptyMatchCase) Name() string {
+	return "omitempty_match_case"
+}
+
+func (n *OmitEmptyMatchCase) Validate(t *types.Type) ([]string, error) {
+	fields := make([]string, 0)
+
+	// Only validate struct type and ignore the rest
+	switch t.Kind {
+	case types.Struct:
+		for _, m := range t.Members {
+			goName := m.Name
+			jsonTag, ok := reflect.StructTag(m.Tags).Lookup("json")
+			if !ok {
+				continue
+			}
+
+			parts := strings.Split(jsonTag, ",")
+			if len(parts) < 2 {
+				// no tags other than name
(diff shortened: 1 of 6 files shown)


pkg/generators/rules/omitempty_match_case.go
```diff
@@ -0,0 +1,70 @@
...
+package rules
+
+import (
+	"reflect"
+	"strings"
+
+	"k8s.io/gengo/types"
+)
+
+/*
```
apelisse: Are we using these types of comments? :-)

pkg/generators/rules/omitempty_match_case.go
```diff
@@ -0,0 +1,70 @@
...
+"omitempty" must appear verbatim (no case variants).
+*/
+type OmitEmptyMatchCase struct{}
+
+// Name returns the name of APIRule
+func (n *OmitEmptyMatchCase) Name() string {
+	return "omitempty_match_case"
+}
+
+// Validate evaluates API rule on type t and returns a list of field names in
```
apelisse: That sounds like a generic comment for the `Validate` method, rather than what this method does specifically.

pkg/generators/rules/omitempty_match_case_test.go
```diff
@@ -0,0 +1,113 @@
...
+					},
+				},
+			},
+			expected: []string{"PodSpec"},
+		},
+	}
+
+	n := &OmitEmptyMatchCase{}
+	for _, tc := range tcs {
+		if violations, _ := n.Validate(tc.t); !reflect.DeepEqual(violations, tc.expected) {
```
apelisse: nit, but `t.Run`?

pkg/generators/rules/omitempty_match_case.go
```diff
@@ -0,0 +1,70 @@
...
+// the type that violate the rule. Empty field name [""] implies the entire
+// type violates the rule.
+func (n *OmitEmptyMatchCase) Validate(t *types.Type) ([]string, error) {
```
apelisse: I'm assuming we don't have to nil-check?

apelisse:
Do we have any integration testing for this tool?

pkg/generators/rules/omitempty_match_case.go
liggitt: whole file was a copy/paste of https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/rules/names_match.go, can change these if you want

pkg/generators/rules/omitempty_match_case_test.go
liggitt: same, copy of existing test file, can change if you want

pkg/generators/rules/omitempty_match_case.go
liggitt: yeah, type is always non-nil

liggitt:
don't think so. @roycaihw, do you know?

apelisse:
@liggitt Would you mind fixing the broken comment? I guess that's the most important one.

liggitt:
comments addressed

pkg/generators/rules/omitempty_match_case_test.go
```diff
@@ -0,0 +1,114 @@
...
+func TestOmitEmptyMatchCase(t *testing.T) {
+	tcs := []struct {
+		// name of test case
+		name string
+		t    *types.Type
+
+		// expected list of violation fields
+		expected []string
+	}{
+		// The comments are in format of {goName, jsonName, match},
```
roycaihw: Please remove this line and `{goName, jsonName, match}`s below

liggitt: done

liggitt:
added integration test

roycaihw:
/assign @mbohlool 
for approval

PR #107, opened 2018-10-15
Add roycaihw to reviewers and update integration test README

roycaihw:
Add @roycaihw to reviewers in OWNER file per request from @apelisse. 

Also update `test/integration/README.md` to mention that the checked-in `pkg/generated/openapi_generated.go` file [should not be updated](https://github.com/kubernetes/kube-openapi/pull/104#issuecomment-424858284). 

/cc @apelisse

### DIFF
--- a/pkg/generators/rules/OWNERS
+++ b/pkg/generators/rules/OWNERS
@@ -1,0 +1,4 @@
+reviewers:
+- roycaihw
+approvers:
+- roycaihw
--- a/OWNERS
+++ b/OWNERS
@@ -5,6 +5,7 @@ reviewers:
 - philips
 - seans3
 - apelisse
+- roycaihw
 approvers:
 - mbohlool
 - lavalamp
--- a/test/integration/README.md
+++ b/test/integration/README.md
@@ -29,3 +29,12 @@ the current directory.
 ```bash
 $ go run builder/main.go testdata/golden.json
 ```
+
+After the golden spec is generated, please clean up the generated file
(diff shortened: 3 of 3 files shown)


apelisse:
I have the feeling that you should be an approver somewhere in there, maybe create a new OWNERS file?

roycaihw:
@apelisse Added a hierarchical `pkg/generators/rules/OWNERS` file. It should likely be enough for me to approve when people are updating existing rules. 

However when new rules are being added, I'd expect `pkg/generators/api_linter.go` (one level higher) and files under `test/integration/testdata/` to also be updated (e.g. https://github.com/kubernetes/kube-openapi/pull/106), in which case I need to bug other approvers. 

Also still adding me to top-level reviewers if there is no objection.

pkg/generators/rules/OWNERS
```diff
@@ -0,0 +1,4 @@
+reviewers:
+- roycaihw
+approvers:
+- roycaihw
```
mbohlool: we should have more than one approver to make sure one person is not bottleneck. You want to go on vacations, right? :) I suggest at least at me and lavalamp here.

mbohlool: my assumption is the ONWERS here overwrites the top level one, if that is not the case,  ignore my comment.

apelisse: it automatically inherits ownership from parent OWNERS. We're good!

PR #109, opened 2018-10-17
More minor fixes.

lavalamp:


lavalamp:
How good is the testing in this repo? I'm pretty confident this doesn't break anything but if it did, would the tests catch it?

apelisse:
There is not a lot of confidence, but it's getting better. I suspect the integration test should catch the functionality of this change.

You are technically changing the API of the generator package (removing `NewOpenAPIGen`), so we'd need to find all the consumer of this package?

lavalamp:
I did search k/k and no one appeared to be calling the function.

lavalamp:
oh whoops, I thought this merged and accidentally reused the branch. one sec.

lavalamp:
This should be mergable again. @seans or @mbohlool could one of you approve?

roycaihw:
Closing as the change has been included in #111 

/close

PR #110, opened 2018-10-18
Convert OpenAPI to structured schema

lavalamp:


lavalamp:
This now works on everything from the main schema, should be ready to go.

/approve

pkg/schemaconv/smd.go
```diff
@@ -122,32 +162,34 @@ func (c *convert) VisitArray(a *proto.Array) {
...
+				c.reportError("missing map keys")
 			}
 		} else {
-			// TODO: emit warning?
+			c.reportError("unknown list type %v", val)
 			l.ElementRelationship = schema.Atomic
 		}
 	} else if val, ok := ext["x-kubernetes-patch-strategy"]; ok {
-		if val == "merge" {
+		if val == "merge" || val == "merge,retainKeys" {
```
apelisse: You could maybe handle retainsKeys correctly and transform it into something useful

apelisse:
> This now works on everything from the main schema, should be ready to go.

That's awesome

PR #111, opened 2018-10-22
Fix api linter

lavalamp:
Includes #109.


lavalamp:
I've only run the unit tests.

lavalamp:
I'm fighting a bit with the integration test.

Ordinary usage should not need to do anything fancy, just call with the --verify flag rather than checking with the diff command. The integration test here does indeed need to do something different :/

lavalamp:
OK, with this change it prepends the output package to the report file name. I'll fix it later.

lavalamp:
OK, I've now fixed the integration test. PTAL.

roycaihw:
LGTM.

One nit is that the `test/integration/pkg/generated/openapi_generated.go` change doesn't need to be checked-in (was trying to update the [README](https://github.com/kubernetes/kube-openapi/pull/107/files#diff-e1d8df67760b2710208b2dd3263cf74a)).

PR #112, opened 2018-10-24
Move general setup to its own file.

lavalamp:
Minor cleanup.


lavalamp:
(This is just a move, except for the comment I added.)

pkg/generators/config.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
seans3: Small nit: 2018 for a new file?

lavalamp: fixed

lavalamp:
> does it make sense to add a unit test for this new file?

Looks like the code I moved wasn't unit tested to begin with.

> If not in this PR, in another?

At the moment I think the integration test gives sufficient coverage, and unit tests wouldn't add additional safety? If it gets more complex then it deserves some unit tests.

PR #118, opened 2018-11-13
smd conversion: handle more string formats

lavalamp:


lavalamp:
Should I tell it that "byte" should still be a string?

apelisse:
Yeah, I'm fine either way

/lgtm
/hold

lavalamp:
Updated to still treat "byte" as a string.

/hold cancel

PR #120, opened 2018-11-29
Add type: object to all generated struct objects

mbohlool:
Look like we did not set type on definitions of struct type. Most of the OpenAPI tools should assume "object" I guess, thats why we didn't catch this.

Ref: https://github.com/kubernetes/kubernetes/issues/56428


roycaihw:
Please update the golden file for integration test: https://github.com/kubernetes/kube-openapi/blob/master/test/integration/README.md

mbohlool:
@roycaihw done.

PR #123, opened 2018-12-05
Make the namesMatch function vendorable

luxas:
/assign @sttts @lavalamp

This is needed for moving out the API violations check from consuming kube-openapi as a binary and writing the violations to an exceptions file in `api/` of k/k to an unit test per API group that want to check its types. This is part of the k8s.io/component KEP.

### DIFF
--- a/pkg/generators/rules/names_match.go
+++ b/pkg/generators/rules/names_match.go
@@ -104,7 +104,7 @@ func (n *NamesMatch) Validate(t *types.Type) ([]string, error) {
 				continue
 			}
 			jsonName := strings.Split(jsonTag, ",")[0]
-			if !namesMatch(goName, jsonName) {
+			if !GoAndJSONNameMatch(goName, jsonName) {
 				fields = append(fields, goName)
 			}
 		}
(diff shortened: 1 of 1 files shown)


lavalamp:
Sorry, I don't understand the justification. This should be a compile time check, not a run time check? I want users of kube-openapi to consume the binary, not the libraries, at least as far as the generators / linters go.

pkg/generators/rules/names_match.go
```diff
@@ -104,15 +104,15 @@ func (n *NamesMatch) Validate(t *types.Type) ([]string, error) {
...
+			if !GoAndJSONNameMatch(goName, jsonName) {
 				fields = append(fields, goName)
 			}
 		}
 	}
 	return fields, nil
 }
 
-// namesMatch evaluates if goName and jsonName match the API rule
+// GoAndJSONNameMatch evaluates if goName and jsonName match the API rule
```
lavalamp: I don't think this should be exported. But if it must be exported, make a new package for it to live in; no one should be importing this rules package IMO, in fact we should move it under an /internal directory.

luxas:
@lavalamp I'm planning to make the tag checking for API types we have in k8s (in pkg/master right now) better and generically useful. So API types not in the k8s core repo can check for naming violations in an unit test in that exact package.

I'm 100% happy with moving this to an other pkg, what do you prefer? e.g. pkg/util/tagnaming?

roycaihw:
How do you plan to generate Openapi specs for the components? It looks like you want to have specs generated (in the KEP) but you don't want to use kube-openapi?

luxas:
> How do you plan to generate Openapi specs for the components?

We want to use the kube-openapi binary to generate specs, but will vendor this code for allowing naming exceptions. Does that sound feasible?

luxas:
decided in person with @lavalamp at KubeCon and will go an other route.

PR #124, opened 2018-12-13
aggregator: optimize 40x speed, 1/8 mem, 1/27 allocs

sttts:
- avoid deep-copy
- rename in one batch

Old kube merging (with real specs) took 738ms, this reduces it to <19ms.


pkg/aggregator/aggregator_test.go
```diff
@@ -1170,13 +1194,18 @@ definitions:
 `), &expected)
 
 	ast := assert.New(t)
-	if !ast.NoError(MergeSpecs(spec1, spec2)) {
+	orig_spec2, _ := cloneSpec(spec2)
```
smarterclayton: Is there anything in here already that verifies cloneSpec works correctly?  A bug in cloneSpec would silently fail some of these tests.

smarterclayton:
First pass looked ok to me, but it's been a while since I dug into the openapi walker and I may have missed something.  Extra eyes appreciated.

mbohlool:
@smarterclayton I will take a look today

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+	return walker.Start(sp)
+}
+
+func (w *mutatingReferenceWalker) walkSchema(schema *spec.Schema) *spec.Schema {
+	if schema == nil {
+		return nil
+	}
+
+	orig := schema
+	clone := func() {
```
mbohlool: can you share the code between mutating and non-mutating walkers by defining this function at the type level?

mbohlool: Then obviously have one type with two different factory function.

mbohlool: I see, the difference is not this clone function but is the `mutatingReferenceWalker`. I think there is still a chance to share most of the code by defining a new type that implements both mutatingReferenceWalker using a readonlyReferenceWalker.

pkg/aggregator/aggregator.go
```diff
@@ -28,21 +27,42 @@ import (
 )
 
 // usedDefinitionForSpec returns a map with all used definitions in the provided spec as keys and true as values.
-func usedDefinitionForSpec(sp *spec.Swagger) map[string]bool {
+func usedDefinitionForSpec(root *spec.Swagger) map[string]bool {
 	usedDefinitions := map[string]bool{}
-	walkOnAllReferences(func(ref spec.Ref) spec.Ref {
-		if refStr := ref.String(); refStr != "" && strings.HasPrefix(refStr, definitionPrefix) {
-			usedDefinitions[refStr[len(definitionPrefix):]] = true
+	alreadyVisited := map[string]bool{}
```
mbohlool: Why the logic of alreadyVisited is removed from the walker? Wasn't it nicer for the walker to do the job locally?

pkg/aggregator/aggregator.go
```diff
@@ -28,21 +27,42 @@ import (
...
-		return ref
-	}, sp)
+
+		defName := refStr[len(definitionPrefix):]
+		usedDefinitions[defName] = true
+
+		if _, found := root.Definitions[defName]; found && !alreadyVisited[refStr] {
+			alreadyVisited[refStr] = true
+			def := root.Definitions[defName]
+			walker.walkSchema(&def)
```
mbohlool: I don't think the callback function should be responsible for this or even be able to call a part of walker.

pkg/aggregator/aggregator.go
```diff
@@ -28,21 +27,42 @@ import (
...
+	walker.Start()
+
 	return usedDefinitions
 }
 
 // FilterSpecByPaths removes unnecessary paths and definitions used by those paths.
 // i.e. if a Path removed by this function, all definitions used by it and not used
 // anywhere else will also be removed.
-func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
+// It does not modify the input, but the output shares data structures with the input.
```
mbohlool: Is this a changed behaviour? `but the output shares data structures with the input.`? if it is, it would be a backward incompatible change as this function is public.

mbohlool: Oh, the function has a new signature. I think it should be fine.

mbohlool: sorry, goine back and forth on this. This is definitely backward incompatible. Any caller outside of the repo expect FilterSpecByPath to mutate the input.

pkg/aggregator/aggregator.go
```diff
@@ -51,71 +71,89 @@ func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
...
-	delete(s.Definitions, old)
+	ret.Definitions = renamedDefinitions
+
+	return ret
 }
 
-// MergeSpecsIgnorePathConflict is the same as MergeSpecs except it will ignore any path
+// MergePathsAndDefinitionsIgnorePathConflict is the same as MergePathAndDefinitions except it will ignore any path
 // conflicts by keeping the paths of destination. It will rename definition conflicts.
-func MergeSpecsIgnorePathConflict(dest, source *spec.Swagger) error {
```
mbohlool: renaming a public method is backward incompatible.

pkg/aggregator/aggregator.go
```diff
@@ -51,71 +71,89 @@ func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
...
+// The source is not mutated.
+func MergePathsAndDefinitionsIgnorePathConflict(dest, source *spec.Swagger) error {
 	return mergeSpecs(dest, source, true, true)
 }
 
-// MergeSpecsFailOnDefinitionConflict is differ from MergeSpecs as it fails if there is
+// MergePathsAndDefinitionsFailOnDefinitionConflict is differ from MergePathAndDefinitions as it fails if there is
 // a definition conflict.
-func MergeSpecsFailOnDefinitionConflict(dest, source *spec.Swagger) error {
```
mbohlool: renaming a public method is backward incompatible.

pkg/aggregator/aggregator.go
```diff
@@ -51,71 +71,89 @@ func FilterSpecByPaths(sp *spec.Swagger, keepPathPrefixes []string) {
...
+// The source is not mutated.
+func MergePathsAndDefinitionsFailOnDefinitionConflict(dest, source *spec.Swagger) error {
 	return mergeSpecs(dest, source, false, false)
 }
 
-// MergeSpecs copies paths and definitions from source to dest, rename definitions if needed.
+// MergePathAndDefinitions copies paths and definitions from source to dest, rename definitions if needed.
 // dest will be mutated, and source will not be changed. It will fail on path conflicts.
-func MergeSpecs(dest, source *spec.Swagger) error {
```
mbohlool: renaming a public method is backward incompatible.

pkg/aggregator/aggregator_test.go
sttts: It's a basic `json.Marshal+Unmarshal`. Not so much that can go wrong. But I can add a test of course.

pkg/aggregator/mutating_walker.go
sttts: We could implement the readonly version on-top of the mutating one. But I don't see much sense the other way around.

pkg/aggregator/aggregator.go
sttts: The logic does no belong into a recursive walker as it is orthogonal. Why should we mix things if we can keep them nicely separated?

pkg/aggregator/aggregator.go
sttts: why not? I prefer to have this special logic cleanly separated. Note that we only do that in the readonly walker. The mutating walker does not need this and should not do this. Another sign, this is special behaviour.

pkg/aggregator/aggregator.go
sttts: fixed

pkg/aggregator/aggregator.go
sttts: fixed

pkg/aggregator/aggregator.go
sttts: fixed

pkg/aggregator/aggregator.go
sttts: fixed

pkg/aggregator/aggregator_test.go
sttts: added a test

pkg/aggregator/walker.go
```diff
@@ -0,0 +1,140 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aggregator
+
+import (
+	_ "net/http/pprof"
```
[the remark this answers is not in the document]
sttts: done

pkg/aggregator/mutating_walker.go
mbohlool: That's what I meant, readonly one be on top of mutating one.

pkg/aggregator/aggregator.go
mbohlool: So this will make sure we won't visit a reference twice, right? I consider it a feature of the walker if we define the walker as visiting every reference of the spec once.

pkg/aggregator/aggregator.go
mbohlool: Why the mutating one do not need this?

pkg/aggregator/aggregator.go
sttts: We could create a reference following walker on-top of the normal walker, as a compromise.

sttts:
@mbohlool I move the "follow definition references" logic into the `walkAllTheReferences` func, compare https://github.com/kubernetes/kube-openapi/pull/124/files#diff-578525fe9e9bb713d8aa0f67bef8733eR40. This way it "is part of the walker" and I am happy that structural recursion and semantic reference following is cleanly separated.

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+	responsesCloned := false
+	for k, v := range resps.ResponsesProps.StatusCodeResponses {
+		r := w.walkResponse(&v)
+		if r != &v && !responsesCloned {
+			responsesCloned = true
+			clone()
+			resps.ResponsesProps.StatusCodeResponses = make(map[int]spec.Response, len(orig.StatusCodeResponses))
+			for k2, v2 := range orig.StatusCodeResponses {
+				if k2 == k {
+					break
```
liggitt: why break? I expected this to just copy the whole map. doesn't this stop copying at an arbitrary point and possibly miss items the outer for loop already iterated over?

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+		clone()
+		pathItem.Head = op
+	}
+	if op := w.walkOperation(pathItem.Options); op != pathItem.Options {
+		clone()
+		pathItem.Delete = op
+	}
+	if op := w.walkOperation(pathItem.Options); op != pathItem.Options {
+		clone()
+		pathItem.Delete = op
```
liggitt: copy/paste, should be pathItem.Options?

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+	if op := w.walkOperation(pathItem.Get); op != pathItem.Get {
+		clone()
+		pathItem.Get = op
+	}
+	if op := w.walkOperation(pathItem.Head); op != pathItem.Head {
+		clone()
+		pathItem.Head = op
+	}
+	if op := w.walkOperation(pathItem.Options); op != pathItem.Options {
```
liggitt: copy/paste, should be pathItem.Delete?

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+	for k, v := range schema.Definitions {
+		s := w.walkSchema(&v)
+		if s != &v {
+			if !definitionsCloned {
+				definitionsCloned = true
+				clone()
+				schema.Definitions = make(spec.Definitions, len(orig.Definitions))
+				for k2, v2 := range orig.Definitions {
+					if k2 == k {
+						break
```
liggitt: why break? I expected this to copy fully. I think this short-circuits copying at an arbitrary point and could make us miss things

pkg/aggregator/mutating_walker.go
liggitt: also needs a test that would have caught this

pkg/aggregator/mutating_walker.go
liggitt: also needs a test that would have caught this

pkg/aggregator/mutating_walker.go
liggitt: also needs a test that would have caught this

pkg/aggregator/mutating_walker.go
liggitt: also needs a test that would have caught this

pkg/handler/handler.go
```diff
@@ -190,13 +194,30 @@ func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
 	return nil
 }
 
-func toProtoBinary(spec []byte) ([]byte, error) {
-	var info yaml.MapSlice
-	err := yaml.Unmarshal(spec, &info)
-	if err != nil {
-		return nil, err
+func jsonToYAML(j map[string]interface{}) yaml.MapSlice {
```
liggitt: does it matter if this loses the distinction between `nil` and `{}`?

pkg/handler/handler_test.go
```diff
@@ -25,11 +25,15 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 		t.Errorf("Unexpected error in unmarshalling SwaggerJSON: %v", err)
 	}
 
-	returnedJSON, err := json.MarshalIndent(s, " ", " ")
+	returnedJSON, err := json.Marshal(s)
```
liggitt: any reason this didn't switch to jsoniter as well?

Godeps/Godeps.json
```diff
@@ -396,6 +429,10 @@
 		{
 			"ImportPath": "k8s.io/klog",
 			"Rev": "b9b56d5dfc9208f60ea747056670942d8b0afdc8"
+		},
+		{
+			"ImportPath": "sigs.k8s.io/structured-merge-diff/schema",
```
liggitt: this surprised me... what pulled this in?

Godeps/Godeps.json
```diff
@@ -377,6 +406,10 @@
 			"ImportPath": "k8s.io/gengo/args",
 			"Rev": "75356185a9af8f0464efa792e2e9508d5b4be83c"
 		},
+		{
+			"ImportPath": "k8s.io/gengo/examples/set-gen/sets",
```
liggitt: this surprised me... what pulled this in?

liggitt:
thanks a lot, this looks good overall

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+					}
+					schema.Definitions[k2] = v2
+				}
+			}
+		}
+		if definitionsCloned {
+			schema.Definitions[k] = *s
```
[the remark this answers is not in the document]
sttts: done

pkg/aggregator/mutating_walker.go
```diff
@@ -0,0 +1,513 @@
...
+
+	if r := w.walkResponse(resps.ResponsesProps.Default); r != resps.ResponsesProps.Default {
+		clone()
+		resps.Default = r
+	}
+
+	responsesCloned := false
+	for k, v := range resps.ResponsesProps.StatusCodeResponses {
+		r := w.walkResponse(&v)
+		if r != &v && !responsesCloned {
```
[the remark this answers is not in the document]
sttts: Yes, that's simpler, but also slower because we copy even those values that will be changed lateron.

On the other hand, my variant depends on the same iteration order. And as I just found out (https://blog.golang.org/go-maps-in-action#TOC_7.), golang explicitly does not guarantee that, not even for iterations over the same unmodified map.

pkg/aggregator/mutating_walker.go
sttts: the break was correct with the assumption of same order for each iteration. But as the assumption is wrong, this code does not make sense.

pkg/handler/handler.go
sttts: probably not. Fixed.

pkg/handler/handler_test.go
sttts: no, just forgotten. Fixed.

Godeps/Godeps.json
sttts: Has nothing to do with the PR, but the godeps were out of date:
```
$ git grep k8s.io/gengo/examples/set-gen/sets                                                                                                                  Godeps/Godeps.json:                     "ImportPath": "k8s.io/gengo/examples/set-gen/sets",
pkg/generators/extension.go:    "k8s.io/gengo/examples/set-gen/sets"
pkg/generators/extension_test.go:       "k8s.io/gengo/examples/set-gen/sets"
```

pkg/aggregator/mutating_walker.go
sttts: fixed and tested

pkg/aggregator/mutating_walker.go
sttts: fixed and tested

pkg/aggregator/mutating_walker.go
sttts: fixed and tested

sttts:
Addressed all comment. @liggitt @mbohlool @soltysh ptal

pkg/aggregator/mutating_walker_test.go
```diff
@@ -0,0 +1,263 @@
...
+		t.Fatalf("failed to compile ref regex: %v", err)
+	}
+	invisibleRE, err := regexp.Compile("\"\\$ref\":\"(http://invisible-[^\"]*)\"")
+	if err != nil {
+		t.Fatalf("failed to compile ref regex: %v", err)
+	}
+
+	for i := 0; i < 1000; i++ {
+		visibleRefs := sets.NewString()
+		invisibleRefs := sets.NewString()
```
liggitt: need to reset visibleRefs and invisibleRefs at the top of the for loop, right? if we construct more than once, they might have stale data from the first construction

pkg/aggregator/mutating_walker_test.go
```diff
@@ -0,0 +1,263 @@
...
+			c.Fuzz(&p.Schema)
+			if c.RandBool() {
+				p.Items = &spec.Items{}
+				c.Fuzz(&p.Items.Ref)
+			} else {
+				p.Items = nil
+			}
+		},
+		func(s *spec.Response, c fuzz.Continue) {
+			c.FuzzNoCustom(s)
```
liggitt: this has the potential to fuzz headers visibly

pkg/aggregator/mutating_walker_test.go
sttts: done. invisibleRefs at least needed that. Does not hurt though without.

pkg/aggregator/mutating_walker_test.go
sttts: Isn't that overridden by the following call?

liggitt:
/lgtm

thanks!

lavalamp:
/approve

Since Jordan reviewed and I think Mehdy's comments were addressed.

Godeps/Godeps.json
```diff
@@ -57,6 +53,10 @@
 			"ImportPath": "github.com/go-openapi/swag",
 			"Rev": "1d0bd113de87027671077d3c71eb3ac5d7dbba72"
 		},
+		{
+			"ImportPath": "github.com/golang/glog",
```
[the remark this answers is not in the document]
sttts: I just updated Godeps. Would be good to have a CI job checking that they are up-to-date. Looks like it was forgotten in previous PRs.

jennybuckley:
Looks like this removes the exported CloneSpec function, which is used by [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-aggregator/pkg/controllers/openapi/aggregator.go#L195)

I'm trying to update the kube-openapi version here https://github.com/kubernetes/kubernetes/pull/73756 and it looks like it doesn't build anymore. Is there a recommendation for how to change kubernetes/kubernetes to pull in this change? Should I copy CloneSpec there? should I use some different function added by this PR?

Thanks.

liggitt:
> Looks like this removes the exported CloneSpec function, which is used by [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-aggregator/pkg/controllers/openapi/aggregator.go#L195)
> 
> I'm trying to update the kube-openapi version here [kubernetes/kubernetes#73756](https://github.com/kubernetes/kubernetes/pull/73756) and it looks like it doesn't build anymore. Is there a recommendation for how to change kubernetes/kubernetes to pull in this change?

In progress in https://github.com/kubernetes/kubernetes/pull/71223/commits/293fde19a9aeb44b2d2a1149e710a298662f668a

jennybuckley:
@liggitt Thanks!

PR #132, opened 2019-01-29
Allow model to seed name without go type

roycaihw:
Allow models (e.g. CRDs) which have schema but not go types to seed canonical name. 

This allows us to use the kube-openapi builder more naturally for CRDs: https://github.com/kubernetes/kubernetes/pull/71192

/cc @liggitt @mbohlool @sttts


liggitt:
looks reasonable. can we exercise this in a test?

roycaihw:
@liggitt updated

pkg/util/util.go
```diff
@@ -41,9 +41,17 @@ func ToCanonicalName(name string) string {
 	return strings.Join(nameParts, ".")
 }
 
+// CanonicalTypeNamer is an interface for models without Go types to seed model name
+type CanonicalTypeNamer interface {
+	CanonicalTypeName() string
```
liggitt: upon consideration, this is a fairly vague name for a fairly specific piece of information. the other interface we let things implement is `OpenAPIDefinitionGetter` via `OpenAPIDefinition() *OpenAPIDefinition`. Starting the method with `OpenAPI` is probably a good idea for coherence:

`OpenAPICanonicalTypeName() string`

liggitt: Needs documentation about the expected (required?) format of the returned name. Examples are probably a good idea.

pkg/util/util.go
```diff
@@ -41,9 +41,29 @@ func ToCanonicalName(name string) string {
 	return strings.Join(nameParts, ".")
 }
 
+// OpenAPICanonicalTypeNamer is an interface for models without Go types to seed model name.
+//
+// OpenAPI canonical names are REST resource friendly names, converted from full package path
+// for Go types or from group/version/kind (GVK) for custom resource definition (CRD)
+// Examples:
+//	Go type:                k8s.io/api/core/v1.Pod
+//	OpenAPI canonical name: io.k8s.api.core.v1.Pod
```
liggitt: I don't see the reverse dns transformation happening below... just the package extraction and /vendor/ stripping

pkg/util/util.go
roycaihw: updated. I don't have a good doc/spec that defines "REST friendly name" to refer to, so I just write down some example

I remember this util came from https://github.com/kubernetes/kubernetes/pull/40124/commits/239169a5a45d94b7f404c080743953fcc26f50b6#diff-1f607a313d799650deac9489e362de2cR131. @mbohlool do you have suggestion on how we document this?

pkg/util/util.go
roycaihw: you're right... I'm mixing things up here

the namer is just for providing `package.type` name

pkg/util/util.go
roycaihw: nvm, the namer has nothing to do with canonical (REST friendly) name above

pkg/util/util.go
roycaihw: updated. 

I will also switch 
https://github.com/kubernetes/kubernetes/blob/91d2bd3951f3a2e7290ef0cf5ec546ba228944d7/staging/src/k8s.io/apiserver/pkg/endpoints/openapi/openapi.go#L124
to use `ToRESTFriendlyName` here

pkg/util/util.go
```diff
@@ -21,14 +21,17 @@ import (
...
 //	Input:  k8s.io/api/core/v1.Pod
 //	Output: io.k8s.api.core.v1.Pod
 //
 //	Input:  k8s.io/api/core/v1
 //	Output: io.k8s.api.core.v1
-func ToCanonicalName(name string) string {
+//
+//	Input:  csi.storage.k8s.io/v1alpha1.CSINodeInfo
+//	Output: io.k8s.storage.csi.v1alpha1.CSINodeInfo
+func ToRESTFriendlyName(name string) string {
```
liggitt: I probably wouldn't change the exported function

pkg/util/util.go
```diff
@@ -41,9 +44,30 @@ func ToCanonicalName(name string) string {
...
+// should be used. For custom resource definition (CRD), the canonical name is expected to be
+//     group/version.kind
+//
+// Examples of canonical name:
+//     Go type: k8s.io/kubernetes/pkg/apis/core.Pod
+//     CRD:     csi.storage.k8s.io/v1alpha1.CSINodeInfo
+//
+// Example for vendored Go type:
+//     Original full path:  k8s.io/kubernetes/vendor/k8s.io/api/core/v1.Pod
+//     Canonical name:      k8s.io/api/core/v1.Pod
```
liggitt: excellent doc, thanks

test/integration/builder/main.go
```diff
@@ -109,8 +102,34 @@ func createOpenAPIBuilderConfig() *common.Config {
...
+// Implements OpenAPICanonicalTypeNamer
+var _ = util.OpenAPICanonicalTypeNamer(&typeNamer{})
+
+type typeNamer struct {
+	pkg  string
+	name string
+}
+
+func (t *typeNamer) OpenAPICanonicalTypeName() string {
+	return fmt.Sprintf("./testdata/%s.%s", t.pkg, t.name)
```
liggitt: I'm amused this doesn't follow the documented example return values... what's with the non-absolute path?

roycaihw: type-based naming [enforces absolute path](https://github.com/golang/go/blob/ed15e82413c7b16e21a493f5a647f68b46e965ee/src/reflect/type.go#L76-L81)
namer-based naming doesn't. here we just document it

on the other hand, gengo supports [both absolute and relative path](https://github.com/kubernetes/gengo/blob/0689ccc1d7d65d9dd1bedcc3b0b1ed7df91ba266/parser/parse.go#L210-L212). `"./testdata/"` worked because we used to generate the [definition](https://github.com/kubernetes/kube-openapi/blob/d8b685cd2daadc40121e864b5e87ca8ededd9eed/test/integration/pkg/generated/openapi_generated.go#L32-L39) using relative path

switched to use absolute path instead

pkg/util/util.go
roycaihw: marked the function with misleading name as deprecated

PR #133, opened 2019-02-01
schemaconv: Make RawExtension into an untyped atomic scalar

apelisse:
Because of how RawExtension is built, the schema we generate is slightly confused when we see them in real-life. Make them untyped since there is not much we can do right now.

### DIFF
--- a/pkg/schemaconv/testdata/new-schema.yaml
+++ b/pkg/schemaconv/testdata/new-schema.yaml
@@ -229,7 +229,7 @@ types:
         scalar: string
     - name: data
       type:
-        namedType: io.k8s.apimachinery.pkg.runtime.RawExtension
+        untyped: {}
     - name: kind
       type:
         scalar: string
@@ -753,7 +753,7 @@ types:
         scalar: string
     - name: data
       type:
-        namedType: io.k8s.apimachinery.pkg.runtime.RawExtension
+        untyped: {}
     - name: kind
       type:
         scalar: string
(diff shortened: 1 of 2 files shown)


apelisse:
> continuous-integration/travis-ci/pr Pending — The Travis CI build is in progress

This status is not required on pull-requests, shouldn't it be?

pkg/schemaconv/testdata/new-schema.yaml
```diff
@@ -229,7 +229,7 @@ types:
         scalar: string
     - name: data
       type:
-        namedType: io.k8s.apimachinery.pkg.runtime.RawExtension
```
apelisse: I notice that we're not removing the type though, maybe we should drop it altogether?

lavalamp: Can we leave the name here and just make it declare a thing that is untyped?

lavalamp: (my original plan was to make the schema set the "guess at keys" flag for these)

apelisse: /shrug
I don't really have a strong opinion.

apelisse: "these" being  `untyped` or `RawExtension`?

lavalamp: Specifically, RawExtension maps to Untyped{GuessAtNames: true}.

PR #136, opened 2019-02-01
aggregator: fix and extend mutating walker tests

sttts:
- fix mutating ref numbering
- implement max depth fuzzing logic
- test mutated subsets, not only one ref at a time

### DIFF
--- a/pkg/aggregator/mutating_walker_test.go
+++ b/pkg/aggregator/mutating_walker_test.go
@@ -173,6 +173,7 @@ func TestReplaceReferences(t *testing.T) {
 	for i := 0; i < 1000; i++ {
 		var visibleRefs, invisibleRefs sets.String
 		var seed int64
+		var randSource rand.Source
 		var s *spec.Swagger
 		for {
 			visibleRefs = sets.NewString()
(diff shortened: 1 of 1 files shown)


pkg/aggregator/mutating_walker_test.go
```diff
@@ -52,22 +79,32 @@ func fuzzFuncs(f *fuzz.Fuzzer, refFunc func(ref *spec.Ref, c fuzz.Continue, visi
...
+			enter(s, false, c)
+			defer leave(false)
 
 			// only fuzz those fields we walk into with invisible==false
 			c.Fuzz(&s.Parameters)
 			c.Fuzz(&s.Responses)
 			c.Fuzz(&s.Definitions)
 			c.Fuzz(&s.Paths)
 		},
+		func(p *spec.PathItem, c fuzz.Continue) {
```
sttts: this one was missing before, the walker does not mutate the ref if there is one in a `PathItem`.

pkg/aggregator/mutating_walker_test.go
```diff
@@ -234,15 +236,26 @@ func TestReplaceReferences(t *testing.T) {
 		}
 
 		t.Run(fmt.Sprintf("iteration %d", i), func(t *testing.T) {
-			mutatedRef := visibleRefs.List()[rand.Intn(visibleRefs.Len())]
+			mutatedRefs := sets.NewString()
+			mutationPropability := rand.New(randSource).Float64()
```
soltysh: Probability

sttts: Fixed.

pkg/aggregator/mutating_walker_test.go
sttts: split into another commit.

PR #138, opened 2019-02-05
handler: fix ToProtoBinary and add tests

sttts:
The json2yaml func produced wrong output and ToProtoBinary failed. There was no test whatsoever calling it.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -221,11 +221,11 @@ func jsonToYAML(j map[string]interface{}) yaml.MapSlice {
 	if j == nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


sttts:
/hold

until smoke test is green within kube.

sttts:
@liggitt add the sorting to the tests https://github.com/kubernetes/kube-openapi/pull/138/files#diff-6d69019ef7f24ec1959c9fbf92061168R131. ptal.

sttts:
@liggitt @mbohlool with this PR https://github.com/kubernetes/kubernetes/pull/71223 leads to an equal openapi spec, only different in indention.

PR #140, opened 2019-02-07
generator: make output roundtrippable

sttts:
- Extension values are JSON data. `[]string` is not a valid type therefore (only `[]interface{}` is for slices). The `[]string` objects did not roundtrip.
- If `SchemaOrBool.Schema` is non-nil, `Allows` does not matter (both variants are semantically equivalent). But for roundtrip checks it's helpful to set `Allows` to `true` because `SchemaOrBool.UnmarshalJSON` does the same.
- don't create empty `Properties` fields.

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -562,7 +562,7 @@ func (g openAPITypeWriter) generateMapProperty(t *types.Type) error {
 		return fmt.Errorf("map with non-string keys are not supported by OpenAPI in %v", t)
 	}
 	g.Do("Type: []string{\"object\"},\n", nil)
-	g.Do("AdditionalProperties: &spec.SchemaOrBool{\nSchema: &spec.Schema{\nSchemaProps: spec.SchemaProps{\n", nil)
+	g.Do("AdditionalProperties: &spec.SchemaOrBool{\nAllows: true,\nSchema: &spec.Schema{\nSchemaProps: spec.SchemaProps{\n", nil)
 	typeString, format := openapi.GetOpenAPITypeFormat(elemType.String())
 	if typeString != "" {
 		g.generateSimpleProperty(typeString, format)
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -409,7 +409,7 @@ func (g openAPITypeWriter) emitExtensions(extensions []extension) {
 	for _, extension := range extensions {
 		g.Do("\"$.$\": ", extension.xName)
 		if extension.hasMultipleValues() {
-			g.Do("[]string{\n", nil)
+			g.Do("[]interface{}{\n", nil)
 		}
 		for _, value := range extension.values {
 			g.Do("\"$.$\",\n", value)
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -363,13 +363,14 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 			return err
 		}
 		g.Do("},\n", nil)
-		g.Do("Dependencies: []string{\n", args)
+
 		// Map order is undefined, sort them or we may get a different file generated each time.
 		keys := []string{}
 		for k := range g.refTypes {
 			keys = append(keys, k)
 		}
 		sort.Strings(keys)
+		deps := []string{}
 		for _, k := range keys {
 			v := g.refTypes[k]
 			if t, _ := openapi.GetOpenAPITypeFormat(v.String()); t != "" {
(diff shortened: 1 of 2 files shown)


liggitt:
lgtm, can you add a test type that exercises the empty properties logic if it's easy?

liggitt:
and also add a test that takes the generated object, round-trips it through json, and ensures reflect.DeepEqual?

pkg/generators/openapi.go
```diff
@@ -337,12 +339,22 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
...
+		bsw := g
+		bsw.SnippetWriter = generator.NewSnippetWriter(&propertiesBuf, g.context, "$", "$")
+		required, err := bsw.generateMembers(t, []string{})
 		if err != nil {
 			return err
 		}
-		g.Do("},\n", nil)
+		if propertiesBuf.Len() > 0 {
+			g.Do("Properties: map[string]$.SpecSchemaType|raw${\n", args)
+			g.Do(strings.Replace(propertiesBuf.String(), "$", "$\"$\"$", -1), nil)
```
liggitt: a comment about what this is doing would be very helpful (and maybe a sample before/after string in the comment). also, raw string would let you skip the escapes to make this bit easier to read:
```
`$"$"$`
```

liggitt:
one question/suggestion on the string.Replace, lgtm otherwise

sttts:
travis go get timeout.

/retest

apelisse:
We need to be extra-careful with these changes that:
1. They work with the openapi-parser used by kubectl,
2. They are compatible with how the openapi-parser worked with version skew.

We should at least have a good mechanism to test the first, since the parser belongs to that repository.

apelisse:
The code that parses the schema considers a "type object" with no properties to be `Arbitrary` (any fields are permitted) while a "type object" with an empty properties is an object that doesn't have any fields (most likely to fail any kind of validation). They are strictly the opposite.

liggitt:
/hold

want to think about implications of https://github.com/kubernetes/kube-openapi/pull/140#issuecomment-461532758

liggitt:
> The code that parses the schema considers a "type object" with no properties to be `Arbitrary` (any fields are permitted) while a "type object" with an empty properties is an object that doesn't have any fields (most likely to fail any kind of validation)

The distinction between empty and nil properties didn't survive serialization. As an example, this is the served swagger.json in kubernetes/kubernetes master for CustomResourceSubresourceStatus, which (prior to this change) had an empty Properties set:
```
   "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceSubresourceStatus": {
    "description": "CustomResourceSubresourceStatus defines how to serve the status subresource for CustomResources. Status is represented by the `.status` JSON path inside of a CustomResource. When set, * exposes a /status subresource for the custom resource * PUT requests to the /status subresource take a custom resource object, and ignore changes to anything except the status stanza * PUT/POST/PATCH requests to the custom resource ignore changes to the status stanza",
    "type": "object"
   },
```

liggitt:
The additionalProperties change in this PR also doesn't affect serialization, only round-tripping. A non-nil schema in a SchemaOrBool object means that is the only thing serialized.

liggitt:
> We need to be extra-careful with these changes that:
> 
> 1. They work with the openapi-parser used by kubectl,
> 2. They are compatible with how the openapi-parser worked with version skew.
> 
> We should at least have a good mechanism to test the first, since the parser belongs to that repository.

I agree with this. I think that falls in the requirements for the kubectl skew test bucket. We need tests in k/k ensuring `kubectl create` client-side validation works with an n-1 version of kubectl against current master (for some representative set of objects, including CRD with validation, CRD without validation, objects with properties and without).

For this particular change, we will verify that pulling it into k/k will demonstrate there is no change in the published /openapi/v2 data, so it is a no-op for any consumers of that API.

liggitt:
satisfied myself that there is no externally visible change in the published openapi schema from this PR.

will let @apelisse remove symbolic post-merge hold when he is satisfied as well

apelisse:
I think that's fine. I'm assuming that we have only very few objects that require to have no sub-fields (what would they be good for?), and might already be interpreted as arbitrary today but nobody noticed.

apelisse:
> For this particular change, we will verify that pulling it into k/k will demonstrate there is no change in the published /openapi/v2 data, so it is a no-op for any consumers of that API.

We discussed this offline, but yeah, I don't think we should be pulling k/k in this repository. We can test at least the decoding part since the code used by kubectl leaves in `pkg/util/proto` inside that same repository.

sttts:
> The code that parses the schema considers a "type object" with no properties to be `Arbitrary` (any fields are permitted) while a "type object" with an empty properties is an object that doesn't have any fields (most likely to fail any kind of validation). They are strictly the opposite.

This is wrong. We should not invent our own interpretation of Openapi: 

> Omitting this keyword has the same behavior as an empty object. (https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.4)

sttts:
kubectl does use the normal go-openapi validator, doesn't it?

I don't think the upper, wrong semantics is actually implemented. There is no difference between empty and nil here, both lead to an validation error:

https://github.com/go-openapi/validate/blob/b0a3ed684d0fdd3e1eda00433382188ce8aa7169/object_validator.go#L121

apelisse:
Symbolically removing the hold  ...

/hold cancel

apelisse:
> No it doesn't. We parse the openapi ourselves and have our very own interpretation of it ...

And we use that very same ~code~ interpretation to merge objects in server-side apply.

PR #141, opened 2019-02-07
aggregator: compare spec with semantic equality during merging

sttts:
Depends on https://github.com/kubernetes/utils/pull/81.


sttts:
Probably we don't want that.

PR #142, opened 2019-02-07
Add sttts as approver

liggitt:
sttts has made significant contributions to the repo, both directly and in kubernetes/kubernetes

/assign @lavalamp


apelisse:
I can't approve, but I agree,
/lgtm

PR #143, opened 2019-02-08
Add analysis of pkg/util/proto/validator

sttts:
The OpenAPI validator used in kubectl is full of wrong assumptions. We cannot reject OpenAPI spec features because openapi-gen does not generate them. Other providers (other aggregated apiservers, CRDs) can make use of them.

As general rule of thumb, we have to apply: we can ignore features, making us incomplete (which we are anyway because we do not check for many/most OpenAPI features), but keeping us correct. But we cannot return errors for features we do not understand.

tl/dr: less strict validation is fine.

This PR add analysis comments of the following shape:

- `TODO(wrong)`: here the validation is returning errors which aren't errors according to the OpenAPI standard.
- `TODO(incomplete)`: here the valiation could return an error, but doesn't.

The first class is critical and we should get rid of them ASAP. The second class is cosmetics, no need for action.

Note: As we support one version of skew between master and kubectl, we have to generate reduced (= stripped) CRD validation schemata for at least one master version such that existing, old kubectl will not fall over.

### DIFF
--- a/pkg/util/proto/document.go
+++ b/pkg/util/proto/document.go
@@ -92,13 +92,16 @@ func NewOpenAPIData(doc *openapi_v2.Document) (Models, error) {
 // We believe the schema is a reference, verify that and returns a new
 // Schema
 func (d *Definitions) parseReference(s *openapi_v2.Schema, path *Path) (Schema, error) {
+	// TODO(wrong): a schema with a $ref can have properties. We can ignore them (would be incomplete), but we cannot return an error.
 	if len(s.GetProperties().GetAdditionalProperties()) > 0 {
 		return nil, newSchemaError(path, "unallowed embedded type definition")
 	}
+	// TODO(wrong): a schema with a $ref can have a type. We can ignore it (would be incomplete), but we cannot return an error.
 	if len(s.GetType().GetValue()) > 0 {
 		return nil, newSchemaError(path, "definition reference can't have a type")
 	}
 
+	// TODO(wrong): $refs outside of the definitions are completely valid. We can ignore them (would be incomplete), but we cannot return an error.
 	if !strings.HasPrefix(s.GetXRef(), "#/definitions/") {
 		return nil, newSchemaError(path, "unallowed reference to non-definition %q", s.GetXRef())
 	}
@@ -127,6 +130,7 @@ func (d *Definitions) parseMap(s *openapi_v2.Schema, path *Path) (Schema, error)
 		return nil, newSchemaError(path, "invalid object type")
 	}
 	var sub Schema
+	// TODO(incomplete): this misses the boolean case as AdditionalProperties is a bool+schema sum type.
 	if s.GetAdditionalProperties().GetSchema() == nil {
 		sub = &Arbitrary{
 			BaseSchema: d.parseBaseSchema(s, path),
@@ -157,6 +161,7 @@ func (d *Definitions) parsePrimitive(s *openapi_v2.Schema, path *Path) (Schema, 
 	case Number: // do nothing
 	case Integer: // do nothing
 	case Boolean: // do nothing
+	// TODO(wrong): this misses "null". Would skip the null case (would be incomplete), but we cannot return an error.
 	default:
 		return nil, newSchemaError(path, "Unknown primitive type: %q", t)
 	}
@@ -175,6 +180,7 @@ func (d *Definitions) parseArray(s *openapi_v2.Schema, path *Path) (Schema, erro
 		return nil, newSchemaError(path, `array should have type "array"`)
 	}
 	if len(s.GetItems().GetSchema()) != 1 {
+		// TODO(wrong): Items can have multiple elements. We can ignore Items then (would be incomplete), but we cannot return an error.
 		return nil, newSchemaError(path, "array should have exactly one sub-item")
 	}
 	sub, err := d.ParseSchema(s.GetItems().GetSchema()[0], path)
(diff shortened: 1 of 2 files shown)


pkg/util/proto/document.go
```diff
@@ -227,18 +233,24 @@ func (d *Definitions) parseArbitrary(s *openapi_v2.Schema, path *Path) (Schema,
...
+		// TODO(incomplete): ignoring the rest of s is wrong. As long as there are no conflict, everything from s must be considered
+		// Reference: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#path-item-object
 		return d.parseReference(s, path)
 	}
 	objectTypes := s.GetType().GetValue()
 	switch len(objectTypes) {
 	case 0:
 		// in the OpenAPI schema served by older k8s versions, object definitions created from structs did not include
 		// the type:object property (they only included the "properties" property), so we need to handle this case
+		// TODO: validate that we ever published empty, non-nil properties. JSON roundtripping nils them.
```
sttts: with supported version skew of 1 only 1.13 really matters.

liggitt:
thanks for digging into this
/lgtm

we can chip away at these, adding fixtures to cover the incorrect cases as we do

apelisse:
I would be surprised if there are not more problems :-)

We used to live in a very hermetic world where we generated and consumed our own openapi and a subset of it was fine. 

Just so that I understand properly, are we intending to accept any valid openapi or do we still have a plan to support a specific subset of features?

liggitt:
> Just so that I understand properly, are we intending to accept any valid openapi or do we still have a plan to support a specific subset of features?

Since client-side validation is optional/non-authoritative, it should not fail on valid openapi schemas it doesn't understand and block submission to the server. Otherwise, it hinders our ability to make the published schemas more expressive over time.

apelisse:
Side-note, the title of the PR is inaccurate, this is not "validator" :-)

lavalamp:
It is fine for kubectl to not choke on specs that are valid but not generated by us (although I don't see the point, it doesn't hurt anything, and certainly adding comments explaining the spec to folks like me is useful).

It would not be fine for CRDs to start accepting specs that are valid but not something we could have generated. We should be very explicit that CRDs accept a subset of the spec. (And, IMO, we should switch to a schema description that won't have this problem.)

apelisse:
This is mostly stating facts and questions I think. None of this is my opinion (I share Daniel's opinion his previous comment). 

There are a couple of dimensions that we need to be careful with
- How is this used. For core-types, we are using the openapi definitions for mostly three use-cases
  - Client-side validation, which we may or may not want to discontinue in the future. We do allow certain and reject certain things with our own interpretations (we typically don't accept extra-fields, which we are supposed to)
  - `kubectl explain` is walking through the fields to describe the structure of an object.
  - server-side apply also now walks *this* openapi model to understand how to merge and uses some custom-extensions to add extra semantic.
- What should be accepted and what are the failure modes:
  - Do we want to allow the entire set of possibilities from openapi v3, v2, jsonschema? Are we really able to understand and process oneOf/anyOf/allOf properly? Should we pretend we do, ignore these, fail these?
  - Should we fail when the CRD validation field is set, should kubectl crash/report an error, should the server fail to merge such objects, or fail in cryptic ways?

sttts:
> * Do we want to allow the entire set of possibilities from openapi v3, v2, jsonschema? Are we really able to understand and process oneOf/anyOf/allOf properly? Should we pretend we do, ignore these, fail these?

It is fine if we don't process advanced fields in kubectl. If you are careful, dropping (= ignoring) fields in OpenAPI during validation does not hurt in the sense that we do not reject valid objects. I argue that rejecting valid objects is a bad thing. Allowing bad objects during client side validation is fine (if the server catches it). The very same we have for native types as well today.

We are coming back here to the discussion in https://github.com/kubernetes/enhancements/pull/709 that we mainly care about the skeleton, not about the logic with oneOf/anyOf/allOf/not. We extensively discussed this for pruning. But the amount of spec information we want to publish is a similar question.

> * Should we fail when the CRD validation field is set, should kubectl crash/report an error

Why should it?

>, should the server fail to merge such objects, or fail in cryptic ways?

Why fail and why in a cryptic way? What are the specs that server side apply would fall over? Why can't it extract its very own skeleton with the information it needs?

liggitt:
> It is fine if we don't process advanced fields in kubectl. If you are careful, dropping (= ignoring) fields in OpenAPI during validation does not hurt in the sense that we do not reject valid objects. I argue that rejecting valid objects is a bad thing. Allowing bad objects during client side validation is fine (if the server catches it). The very same we have for native types as well today.

Correct. Client-side validation is advisory, not authoritative, so the the failure mode of unhandled aspects of the spec *must* be to defer to the authoritative validation, the server.

If we want to limit what is allowed in CRD validation schemas, that is a valid conversation to have for v1 CRD planning, but is orthogonal to how kubectl should handle published openapi specs.

pkg/util/proto/document.go
liggitt: we're trying to improve the supported skew so the latest version of kubectl can work against apiserver versions currently in support, but improving behavior of kubectl in areas like this is an important first step

liggitt:
still lgtm
/lgtm

sttts:
Here is a CRD openapi schema filter that ensures that the published spec is compatible with 1.13 kube-openapi: https://github.com/soltysh/origin/pull/31

mbohlool:
Did a pass on this. Looks good.

apelisse:
I'm fine with it, hopefully this code will die quickly anyway, so I don't know if you still want to merge it.
/approve
/hold

Feel free to unhold if you want to get this in.

sttts:
/hold cancel

Doesn't hurt to have this documented.

PR #146, opened 2019-02-15
cleanup: remove deprecated service registration

roycaihw:
for format-separated endpoints, which was deprecated in 1.10 and removed in 1.14

now k8s uses `BuildAndRegisterOpenAPIVersionedService` and `RegisterOpenAPIVersionedService` methods in the same file to register openapi service

/assign @mbohlool


mbohlool:
more code deletion, good day :)

/approve

PR #147, opened 2019-02-21
util/proto: add optional, permissive spec parsing mode

sttts:
When being used in kubectl to do client-side validation or for kubectl-explain, the OpenAPI code should not reject specs that have uninterpreted OpenAPI features. OpenAPI is monotonic in its fields (= weaker field values makes the whole schema weaker, not stronger) as long as the negative quantors `not` and `oneOf` are not used. This means we can just ignore unsupported fields in validation that is by definition incomplete (kubectl validation is incomplete because we have stricter server-side validation). This means:

* it is ok to accept an object that will later be rejected by the server,
* but it is bad to reject objects which are accepted by the server.

Or in yet other terms: kubectl validation should be best effort, but never stricter than the server.

The permissive mode switches the excessively strict validation (in respect to the formal OpenAPI semantics) to be permissive and weaker than server side validation. The only exception to this is that client-side validation will assume `additionalProperties: false` for all objects, while the server-side validation assumes `additionalPropertiers: true` by default.


apelisse:
I'm still curious what the strategy here. I'm not opposed to this PR, but I don't know what the end-goal is, so I don't know if it's going in the right direction.

Here are more thoughts on this topic, since I don't think we're still agreeing. 
We use/want to use the openapi for mainly two things:
1. Validate the object on the client and on the server (not sure why we really want to validate on the client to be honest, so we might want to discuss that too).
2. Understand the possible structure of the object, as well as reading some annotations so that we can manipulate/understand that object, currently/in a near-future, that's mostly `kubectl explain` and server-side apply, and maybe a union semantic.

If we allow any arbitrary openapi model to be used for CRs, then we can still do 1. (client and server) by using `go-openapi/validation` rather than change the currently broken (as you point out) `proto.Model`. Why are we not doing that? We could also fix the openapi we generate for in-tree types so that the validation would be closer to what we actually want (rather than wrongly interpreting the openapi into another thing that we want). To be clear, I vote that we stop using `proto.Models` for client-side validation, and switch to `go-openapi/validation`. That seems like a reasonable comfortable effort.

If we want to do 2., I'm afraid that it might be hard sometimes to extract the "skeleton"/structure from the schema as well as the appropriate `x-kubernetes-...` openapi extensions. I think this is the problem we're trying to address? My concern here is that sometimes we would be enable to extact the proper schema/extensions to do what we're trying to do, and wouldn't be able to report that error to users on time (at CRD creation, not when trying to use it). If we can fix that problem though, then I think that's where the effort should be direct.

liggitt:
> I vote that we stop using `proto.Models` for client-side validation, and switch to `go-openapi/validation`. That seems like a reasonable comfortable effort.

if we can do that, and satisfy `kubectl validation should be best effort, but never stricter than the server` for all combinations of {old kubectl, new kubectl} against {old server, new server}, I'm in favor of it.

sttts:
> if we can do that, and satisfy `kubectl validation should be best effort, but never stricter than the server` for all combinations of {old kubectl, new kubectl} against {old server, new server}, I'm in favor of it.

This sounds good. The only question mark is behind the transition period when we have to send an old openapi spec from an old server to the "new" library. But we could also keep go-openapi and our own proto model based validationin place and switch by server version.

apelisse:
I'm happy we have an agreement here. I'll open an issue so that we can track some of the work that has to be done, and finally stop derailing this PR (sorry about that Stefan ...)

pkg/util/proto/document.go
```diff
@@ -255,18 +291,23 @@ func (d *Definitions) ParseSchema(s *openapi_v2.Schema, path *Path) (Schema, err
...
 	default:
-		// the OpenAPI generator never generates (nor it ever did in the past) OpenAPI type definitions with multiple types
-		return nil, newSchemaError(path, "definitions with multiple types aren't supported")
+		if !d.opts.Permissive {
+			return nil, newSchemaError(path, "definitions with multiple types aren't supported")
+		}
+
+		// Definition has many types. We cannot cope with that. Ignore all of them.
+		// TODO(incomplete): add multi-type support
+		return d.parseArbitrary(s, path)
```
liggitt: what does parseArbitrary result in here?

pkg/util/proto/document.go
```diff
@@ -217,30 +243,40 @@ func (d *Definitions) parseKind(s *openapi_v2.Schema, path *Path) (Schema, error
...
+			} else {
+				// Definition has no type and no properties. Treat it as an arbitrary value
+				// TODO(incomplete): what if it has additionalProperties=false or patternProperties?
+				return d.parseArbitrary(s, path)
+			}
 		}
+
+		// Definition has no type. Treat it as an arbitrary value
+		// TODO(incomplete): this ignores many fields, e.g. properties
+		return d.parseArbitrary(s, path)
```
liggitt: what does parseArbitrary result in here?

pkg/util/proto/document.go
```diff
@@ -217,30 +243,40 @@ func (d *Definitions) parseKind(s *openapi_v2.Schema, path *Path) (Schema, error
...
 	case 0:
-		// in the OpenAPI schema served by older k8s versions, object definitions created from structs did not include
-		// the type:object property (they only included the "properties" property), so we need to handle this case
-		if s.GetProperties() != nil {
-			return d.parseKind(s, path)
-		} else {
-			// Definition has no type and no properties. Treat it as an arbitrary value
-			// TODO: what if it has additionalProperties or patternProperties?
-			return d.parseArbitrary(s, path)
+		// at the root of schemas we forgot the type somewhen between 1.11 and 1.13, to be fixed again for 1.14.
```
liggitt: so the `if s.GetProperties() != nil` check *only* applied to root-level structs? Is there a test of a schema from "in the OpenAPI schema served by older k8s versions" that exercises this path?

pkg/util/proto/document.go
```diff
@@ -56,18 +56,31 @@ func VendorExtensionToMap(e []*openapi_v2.NamedAny) map[string]interface{} {
 	return values
 }
 
-// Definitions is an implementation of `Models`. It looks for
+// definitions is an implementation of `Models`. It looks for
```
liggitt: why unexport this?

pkg/util/proto/openapi_test.go
```diff
@@ -26,117 +26,12 @@ import (
 	"k8s.io/kube-openapi/pkg/util/proto/testing"
 )
 
-var fakeSchema = testing.Fake{Path: filepath.Join("testdata", "swagger.json")}
-var fakeSchemaNext = testing.Fake{Path: filepath.Join("testdata", "swagger_next.json")}
-
-var _ = Describe("Reading apps/v1beta1/Deployment from v1.8 openAPIData", func() {
```
liggitt: can we add tests for 1.13-level openapi?

pkg/util/proto/validation/types.go
```diff
@@ -282,6 +292,12 @@ func itemFactory(path proto.Path, v interface{}) (validationItem, error) {
 			Value:    v,
 			Kind:     proto.String,
 		}, nil
+	case reflect.Interface:
+		return &primitiveItem{
+			baseItem: baseItem{path: path},
+			Value:    v,
+			Kind:     proto.Null,
```
liggitt: what does kind null imply? is there a way to indicate "arbitrary"? kind object?

pkg/util/proto/document.go
apelisse: I shouldn't have exported it in the first place ... Now I don't think anything builds on that, and if it does, it should stop anyway.

PR #149, opened 2019-02-25
handler: convert int types correctly like yaml2 does and gnostic expects

sttts:
Json always unmarshals to float64, yaml2 does best effort down-sizing of the returned numbers, returning int, int64, uint64 or float64. Gnostic expects the later.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -27,14 +27,14 @@ import (
 	"time"
 
 	"github.com/NYTimes/gziphandler"
-	restful "github.com/emicklei/go-restful"
+	"github.com/emicklei/go-restful"
 	"github.com/go-openapi/spec"
 	"github.com/golang/protobuf/proto"
-	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
+	"github.com/googleapis/gnostic/OpenAPIv2"
 	"github.com/googleapis/gnostic/compiler"
 	"github.com/json-iterator/go"
 	"github.com/munnerz/goautoneg"
-	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


apelisse:
I'm curious why this is not using `sigs.k8s.io/yaml`[1] that has a JSonToYaml method built-in, which, according to a comment[2] already does that properly.

[1]: https://godoc.org/sigs.k8s.io/yaml#JSONToYAML
[2]: https://github.com/kubernetes-sigs/yaml/blob/master/yaml.go#L85-L89

sttts:
> I'm curious why this is not using `sigs.k8s.io/yaml`[1](https://godoc.org/sigs.k8s.io/yaml#JSONToYAML) that has a JSonToYaml method built-in, which, according to a comment[2](https://github.com/kubernetes-sigs/yaml/blob/master/yaml.go#L85-L89) already does that properly.

Because it is slow. The one here avoid going through `[]byte`. This was one of bottlenecks in OpenAPI merges. Hence, we optimized it.

apelisse:
is it limited to an interface problem or an implementation? I'm asking because I wonder if we can improve the situation in general as opposed to just in this repo.

sttts:
@apelisse I have created https://github.com/kubernetes-sigs/yaml/pull/14. Can we please merge this first to fix this critical bug? I will create a follow-up with revendoering sig.k8s.io/yaml as soon as the upper PR merged.

sttts:
@mbohlool is this approved?

apelisse:
> Can we please merge this first to fix this critical bug?

Too late, but yes :-)

apelisse:
And really appreciate that you opened that PR!

PR #150, opened 2019-02-26
aggregator: merge x-kubernetes-group-version-kind slices

sttts:
Definitions are supposed to be reused independently from the assign GVKs in the x-kubernetes-group-version-kind field. Different x-kubernetes-group-version-kind can happen if e.g. shared meta/v1 structs like `Status` or `GetOptions` appear in different API groups.

Without this PR, the merging algorithm will create `n` copies of the shared types, which are many in case of kube-apiserver, and even more if there are many CRDs.

Prerequisite for https://github.com/kubernetes/kubernetes/pull/74133.

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -26,6 +26,8 @@ import (
 	"k8s.io/kube-openapi/pkg/util"
 )
 
+const gvkKey = "x-kubernetes-group-version-kind"
+
 // usedDefinitionForSpec returns a map with all used definitions in the provided spec as keys and true as values.
 func usedDefinitionForSpec(root *spec.Swagger) map[string]bool {
 	usedDefinitions := map[string]bool{}
@@ -152,7 +154,7 @@ func MergeSpecs(dest, source *spec.Swagger) error {
 	return mergeSpecs(dest, source, true, false)
 }
 
-// mergeSpecs merged source into dest while resolving conflicts.
+// mergeSpecs merges source into dest while resolving conflicts.
 // The source is not mutated.
 func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConflicts bool) (err error) {
 	// Paths may be empty, due to [ACL constraints](http://goo.gl/8us55a#securityFiltering).
@@ -186,7 +188,7 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 	conflicts := false
 	for k, v := range source.Definitions {
 		v2, found := dest.Definitions[k]
-		if found && !reflect.DeepEqual(v, v2) {
+		if found && !deepEqualDefinitionsModuloGVKs(&v2, &v) {
 			if !renameModelConflicts {
 				return fmt.Errorf("model name conflict in merging OpenAPI spec: %s", k)
 			}
@@ -207,7 +209,12 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 			if usedNames[k] {
 				v2, found := dest.Definitions[k]
 				// Reuse model if they are exactly the same.
-				if found && reflect.DeepEqual(v, v2) {
+				if found && deepEqualDefinitionsModuloGVKs(&v2, &v) {
+					if gvks, found, err := mergedGVKs(&v2, &v); err != nil {
+						return err
+					} else if found {
+						v2.Extensions[gvkKey] = gvks
+					}
 					continue
 				}
 
@@ -218,8 +225,13 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 					i++
 					newName = fmt.Sprintf("%s_v%d", k, i)
 					v2, found = dest.Definitions[newName]
-					if found && reflect.DeepEqual(v, v2) {
+					if found && deepEqualDefinitionsModuloGVKs(&v2, &v) {
 						renames[k] = newName
+						if gvks, found, err := mergedGVKs(&v2, &v); err != nil {
+							return err
+						} else if found {
+							v2.Extensions[gvkKey] = gvks
+						}
 						continue OUTERLOOP
 					}
 				}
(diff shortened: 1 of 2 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -257,3 +267,72 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
+	for _, x := range slice2 {
+		gvk, ok := x.(map[string]interface{})
+		if !ok {
+			return nil, false, fmt.Errorf(`expected {"group": <group>, "kind": <kind>, "version": <version>}, got: %#v`, x)
+		}
+		k := fmt.Sprintf("%s/%s.%s", gvk["group"], gvk["version"], gvk["kind"])
+		if seen[k] {
+			continue
+		}
+		ret = append(ret, x)
```
mbohlool: nit: I would return true if this line at least ran once, i.e. we actually appended anything. this can also be done by checking the length of ret.

pkg/aggregator/aggregator.go
```diff
@@ -257,3 +267,72 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 	}
 	return nil
 }
+
+// deepEqualDefinitionsModuloGVKs compares a and b, but ignores the x-kubernetes-group-version-kind extension.
```
mbohlool: can this temporary modification changes the order of fields? that could be ignorable, but I prefer not to mutate the objects we promised we won't even temporarily. what about checking deepEqual on spec.Schema.SchemaProps and spec.Schema.SwaggerSchemaProps and manually on spec.Schema.Extensions?

pkg/aggregator/aggregator.go
```diff
@@ -257,3 +267,72 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
+// It will mutate a temporarily, but not b.
+func deepEqualDefinitionsModuloGVKs(a, b *spec.Schema) bool {
+	old, oldFound := a.ExtraProps["x-kubernetes-group-version-kind"]
```
mbohlool: ExtraProps is not serialized and not what we use to store GVK. it is spec.Schema.VendorExtensible

pkg/aggregator/aggregator.go
sttts: actually we can just set the whole map to nil, do comparison and then compare the old values independently. No modification necessary.

sttts: fixed

pkg/aggregator/aggregator.go
sttts: good point. Fixed.

pkg/aggregator/aggregator.go
sttts: good idea. Fixed.

sttts:
@mbohlool addressed all your comments. Thanks for the review.

roycaihw:
talked with @mbohlool before. still trying to understand the use case here. maybe an example would help  explain (what different gvk extensions will be assigned to the same Go type definition (for delegated API servers))?

> Definitions are supposed to be reused independently from the assign GVKs in the x-kubernetes-group-version-kind field

the statement I totally agree. LGTM

mbohlool:
The code looks good now. An example usecase of this would be nice. @sttts  @roycaihw

pkg/aggregator/aggregator.go
```diff
@@ -257,3 +267,71 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
 	return nil
 }
+
+// deepEqualDefinitionsModuloGVKs compares s1 and s2, but ignores the x-kubernetes-group-version-kind extension.
+func deepEqualDefinitionsModuloGVKs(s1, s2 *spec.Schema) bool {
+	if !reflect.DeepEqual(s1.Extensions, s2.Extensions) {
+		if s1.Extensions != nil {
+			shallowCopy := *s1
+			s1 = &shallowCopy
+			s1.Extensions = nil
```
roycaihw: it's okay to do so now, but we need to keep in mind that this function is actually `Modulo"Extensions"`. New extension key could be dropped silently during merging if we add any in future

sttts:
@roycaihw @mbohlool the example is already in the description: shared meta/v1 structs like Status or GetOptions appear in different API groups.

pkg/aggregator/aggregator.go
```diff
@@ -207,7 +207,12 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
 			if usedNames[k] {
 				v2, found := dest.Definitions[k]
 				// Reuse model if they are exactly the same.
-				if found && reflect.DeepEqual(v, v2) {
+				if found && deepEqualDefinitionsModuloGVKs(&v2, &v) {
+					if gvks, found, err := mergedGVKs(&v2, &v); err != nil {
+						return err
+					} else if found {
+						v2.Extensions["x-kubernetes-group-version-kind"] = gvks
```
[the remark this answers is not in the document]
sttts: it mutates the `dest` which is intentional and correct.

pkg/aggregator/aggregator.go
sttts: fixed. And added tests.

sttts:
@roycaihw this is pre-requisite for the server-side-apply wiring for CRDs. I addressed your comment, including new unit tests for the equality-module-gvks. Ptal.

pkg/aggregator/aggregator.go
```diff
@@ -257,3 +269,96 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
+// s1's x-kubernetes-group-version-kind slice was changed at all.
+func mergedGVKs(s1, s2 *spec.Schema) (interface{}, bool, error) {
+	gvk1, found1 := s1.Extensions[gvkKey]
+	gvk2, found2 := s2.Extensions[gvkKey]
+
+	if !found1 {
+		return gvk2, found2, nil
+	}
+	if !found2 {
+		return gvk1, found1, nil
```
p0lyn0mial: nit: shouldn't this be `gvk1, false, nil` ? If we haven't found `s2` we don't change `s1` thus this is noop

pkg/aggregator/aggregator.go
```diff
@@ -257,3 +269,96 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
+		return nil, false, fmt.Errorf("expected slice of GroupVersionKinds, got: %+v", slice1)
+	}
+	slice2, ok := gvk2.([]interface{})
+	if !ok {
+		return nil, false, fmt.Errorf("expected slice of GroupVersionKinds, got: %+v", slice2)
+	}
+
+	ret := make([]interface{}, len(slice1), len(slice1)+len(slice2))
+	copy(ret, slice1)
+	seen := make(map[string]bool, len(slice1)+len(slice2))
```
p0lyn0mial: nit: I think that `seen` could be only `len(slice1)`.

pkg/aggregator/aggregator.go
sttts: yep, fixed.

pkg/aggregator/aggregator.go
sttts: fixed

sttts:
@p0lyn0mial addressed your comments.

PR #152, opened 2019-04-02
Add parsing of openapi oneof/union semantic into smd

apelisse:
Implements first part of https://github.com/kubernetes/enhancements/pull/926.
We still need to change the OpenAPI generation to add these extensions.

### DIFF
--- a/pkg/schemaconv/smd.go
+++ b/pkg/schemaconv/smd.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"sort"
 	"strings"
 
 	"k8s.io/kube-openapi/pkg/util/proto"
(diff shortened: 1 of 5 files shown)


pkg/schemaconv/smd.go
```diff
@@ -122,6 +122,78 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
 	return tr
 }
 
+func parseOneOfDiscriminator(v interface{}) (bool, error) {
```
apelisse: these two functions shouldn't be here.

apelisse:
tests are expected to fail, there is a dependency to https://github.com/kubernetes-sigs/structured-merge-diff/pull/72

pkg/schemaconv/smd.go
```diff
@@ -122,6 +122,56 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
 	return tr
 }
 
+func makeUnion(extensions map[string]interface{}) (*schema.Union, error) {
+	union := schema.Union{
+		Fields: map[string]string{},
+	}
+
+	if idiscriminator, ok := extensions["x-kubernetes-oneof-discriminator"]; ok {
```
sttts: is this a new extension or do we publish it already now?

pkg/schemaconv/testdata/swagger.json
```diff
@@ -78578,6 +78578,10 @@
    },
    "io.k8s.api.apps.v1beta1.DeploymentStrategy": {
     "description": "DeploymentStrategy describes how to replace existing pods with new ones.",
+    "x-kubernetes-oneof-discriminator": "type",
+    "x-kubernetes-oneof-fields": {
+     "rollingUpdate": "RollingUpdate"
```
lavalamp: I assume this is a map from field name to discriminator value? Can we change the name of "x-kubernetes-oneof-fields" until this is clearer?

pkg/schemaconv/smd.go
```diff
@@ -122,6 +122,56 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
+		discriminator, ok := idiscriminator.(string)
+		if !ok {
+			return nil, fmt.Errorf(`"x-kubernetes-oneof-discriminator" is of invalid type: %T`, idiscriminator)
```
lavalamp: Might be nice to print the value, too. I think `%#v` includes both.

pkg/schemaconv/smd.go
```diff
@@ -122,6 +122,56 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
+
+	if ifields, ok := extensions["x-kubernetes-oneof-fields"]; ok {
+		fields, ok := ifields.(map[interface{}]interface{})
+		if !ok {
+			return nil, fmt.Errorf(`"x-kubernetes-oneof-fields" is of invalid type: %T`, ifields)
+		}
+		for ifield, value := range fields {
+			field, ok := ifield.(string)
+			if !ok {
+				return nil, fmt.Errorf(`"x-kubernetes-oneof-fields": field is of invalid type: %T`, ifield)
```
lavalamp: Definitely need to print the values to make this and the following error actionable.

pkg/schemaconv/smd.go
```diff
@@ -122,6 +122,56 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
+			discriminated, ok := value.(string)
+			if !ok {
+				return nil, fmt.Errorf(`"x-kubernetes-oneof-fields"/%v: value is of invalid type: %T`, field, value)
+			}
+			union.Fields[field] = discriminated
+		}
+	}
+
+	if union.Discriminator != nil && len(union.Fields) == 0 {
+		return nil, fmt.Errorf("discriminator set to %v, but no fields in union", *union.Discriminator)
```
lavalamp: Is this actually an error?

lavalamp: How does one indicate whether or not "None" is a permitted value?

pkg/schemaconv/testdata/new-schema.yaml
```diff
@@ -6621,6 +6625,35 @@ types:
     - name: vsphereVolume
       type:
         namedType: io.k8s.api.core.v1.VsphereVirtualDiskVolumeSource
+    union:
+      fields:
```
lavalamp: Sorry to leave this comment here and not in an earlier change, but...
would this make more sense as "choices", where it's a list of structs which set "field" and "discriminatedBy" or something? e.g.

```
union:
  discriminator: type
  options:
  - fieldName: fc
    discriminatedBy: FC
  - fieldName: azureDisk
    discriminatedBy: AzureDisk
...
```

I think it is more understandable that way, what do you think?

lavalamp: You could even indicate a "none" option that way by permitting a single entry to have no fieldName and just a discriminatedBy.

(no strong opinion about the exact names intended)

apelisse: Yeah, Jenny made a similar comment on the other PR, I agree. That's also much more kubernetes-style.

pkg/schemaconv/smd.go
apelisse: > Is this actually an error?

What's the point? if the only thing you have a discriminator, then it's just an enum. I would suspect that they happen because of a user-error.

apelisse:
Updated this pull-request and rebased it on top of #155 so that it compiles, PTAL @sttts @lavalamp

pkg/schemaconv/smd.go
```diff
@@ -116,12 +140,72 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
 
+func makeUnion(extensions map[string]interface{}) (*schema.Union, error) {
+	union := schema.Union{
+		Fields: []schema.UnionField{},
+	}
+
+	if idiscriminator, ok := extensions["x-kubernetes-union-discriminator"]; ok {
+		discriminator, ok := idiscriminator.(string)
+		if !ok {
+			return nil, fmt.Errorf(`"x-kubernetes-union-discriminator" is of invalid type: %#v`, idiscriminator)
```
sttts: nit: write which type you expect

pkg/schemaconv/testdata/swagger.json
sttts: +1

pkg/schemaconv/smd.go
apelisse: Very good point, I just fixed them all! Thanks, PTAL

pkg/schemaconv/testdata/swagger.json
apelisse: Changed to `x-kubernetes-unino-fields-discriminated`

pkg/schemaconv/smd.go
```diff
@@ -145,6 +146,116 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
+		for ifield := range fields {
+			field, ok := ifield.(string)
+			if !ok {
+				return nil, fmt.Errorf(`"fields-discriminated": field must be a string, got: %#v`, ifield)
+			}
+			keys = append(keys, field)
+
+		}
+		sort.Strings(keys)
+		reverseMap := map[string]string{}
```
jennybuckley: What's the point of keeping the values in here?

pkg/schemaconv/smd.go
```diff
@@ -145,6 +146,116 @@ func (c *convert) makeRef(model proto.Schema) schema.TypeRef {
...
+				if !ok {
+					return nil, fmt.Errorf(`"x-kubernetes-unions" has non-string key: %#v`, k)
+				}
+				unionMap[key] = v
+			}
+			schemaUnion, err := makeUnion(unionMap)
+			if err != nil {
+				return nil, err
+			}
+			if schemaUnion != nil {
```
jennybuckley: Why would it be nil, should that be an error also?

apelisse:
Fixed, PTAL

pkg/schemaconv/testdata/swagger.json
```diff
@@ -78578,6 +78578,12 @@
    },
    "io.k8s.api.apps.v1beta1.DeploymentStrategy": {
     "description": "DeploymentStrategy describes how to replace existing pods with new ones.",
+    "x-kubernetes-unions": [{
+     "discriminator": "type",
+     "fields-discriminated": {
```
sttts: does this mean we don't support non-discriminator unions? Would rather call it `fields` instead of `fields-discriminated`.

sttts: as the type is different, we would in fact need `fields []string` and `discriminatedFields map[string]string`.

apelisse: Yeah I'll probably do that ... if `discriminator` is set, then `fields` must be unset and `fields-discriminated` must be set. and the other way around.

PR #154, opened 2019-04-08
Create go-modules to replace Godeps

apelisse:


go.mod
```diff
@@ -0,0 +1,37 @@
+module k8s.io/kube-openapi
+
+go 1.12
+
+require (
+	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
```
liggitt: were these primed from the godeps.json file? should we drop the Godeps dir or do we intend to maintain that file as well?

liggitt:
need to update travis.yml to stop using godeps:
* bump the version of go to 1.12
* drop all the godeps stuff
* set GO111MODULE=on (see https://github.com/kubernetes/client-go/pull/582/files for an example)

go.mod
apelisse: > were these primed from the godeps.json file?

Yes

> should we drop the Godeps dir or do we intend to maintain that file as well?

My gut feeling would be that we should remove it, do you have a recommendation?

apelisse:
Do we need to run some step to verify that the go modules things are up-to-date/correct?

go.mod
liggitt: I'd recommend removing it

liggitt:
> Do we need to run some step to verify that the go modules things are up-to-date/correct?

yes, `go mod tidy` should be zero-diff

go.mod
apelisse: Done

PR #156, opened 2019-04-19
Parse union tags in IDL, generate proper openapi

apelisse:
/assign @kwiesmueller @jennybuckley @lavalamp 
/cc @sttts @jpbetz

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -397,11 +397,18 @@ func (g openAPITypeWriter) generateStructExtensions(t *types.Type) error {
 	// Initially, we will only log struct extension errors.
 	if len(errors) > 0 {
 		for _, e := range errors {
-			klog.V(2).Infof("[%s]: %s\n", t.String(), e)
+			klog.Errorf("[%s]: %s\n", t.String(), e)
 		}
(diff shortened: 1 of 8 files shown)


pkg/generators/openapi.go
```diff
@@ -416,13 +423,13 @@ func (g openAPITypeWriter) generateMemberExtensions(m *types.Member, parent *typ
 			klog.V(2).Infof("%s %s\n", errorPrefix, e)
 		}
 	}
-	g.emitExtensions(extensions)
+	g.emitExtensions(extensions, nil)
 	return nil
 }
 
-func (g openAPITypeWriter) emitExtensions(extensions []extension) {
+func (g openAPITypeWriter) emitExtensions(extensions []extension, unions []union) {
```
jennybuckley: Is there a reason why unions isn't just a part of extensions?

apelisse: Yeah, I would have to completely change the way extensions are built (they're meant for list of scalars) and I didn't want to make the type worse by adding a bunch of unrelated things.

jennybuckley: sounds good

PR #159, opened 2019-04-29
Rename `fields-discriminated` to `fields-to-discriminateBy`

apelisse:


sttts:
I prefer `fields` very much.

/approve
lgtm

Giving @lavalamp a chance for a final look.

test/integration/testdata/golden.json
```diff
@@ -402,7 +402,7 @@
     "x-kubernetes-unions": [
      {
       "discriminator": "unionType",
-      "fields-discriminated": {
+      "fields": {
```
lavalamp: I think that just plain "fields" is going to be confusing, which is first, the field or its discriminator? How do I even know the second value is a discriminator?

How about "fields-to-discriminator-value"? It's really wordy :/ but I think just "fields" doesn't help people enough.

field-names? fields-to-names?

apelisse: Out of all these options, I like `fields-discriminated` most :-)

apelisse: Let's use `fields-to-discriminatedBy`?

apelisse: Renamed, PTAL

apelisse: renamed to `fields-to-discriminateBy`

PR #160, opened 2019-05-01
Remove constraint on discriminator type

apelisse:
This constraint doesn't work because discriminator are often defined as
a type alias to a string.


jennybuckley:
We could validate this using schemaconv, I think it would be able to tell whether the alias is for a string or not.

PR #162, opened 2019-05-30
util/proto: ignore uninterpreted OpenAPI features instead of rejecting

roycaihw:
This is mostly copied from https://github.com/kubernetes/kube-openapi/pull/147. Make the proto validator ignores valid OpenAPI features that it doesn't understand instead of rejecting. 

fixes https://github.com/kubernetes/kubernetes/pull/78326

tested locally and verified that the CR didn't get rejected with `array should have exactly one sub-item`: https://github.com/roycaihw/kubernetes/commits/local-test/proto-validator-fix

Note: 
1. I prefer https://github.com/kubernetes/kube-openapi/pull/147 which is more test-complete
2. go mod in k/k lags behind this repo, so updating dependency in k/k will involve multiple commits

/cc @sttts

### DIFF
--- a/pkg/util/proto/document.go
+++ b/pkg/util/proto/document.go
@@ -161,7 +161,7 @@ func (d *Definitions) parsePrimitive(s *openapi_v2.Schema, path *Path) (Schema, 
 	case Number: // do nothing
 	case Integer: // do nothing
 	case Boolean: // do nothing
-	// TODO(wrong): this misses "null". Would skip the null case (would be incomplete), but we cannot return an error.
+	case Null: // do nothing
 	default:
 		return nil, newSchemaError(path, "Unknown primitive type: %q", t)
 	}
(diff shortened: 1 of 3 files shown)


roycaihw:
test failed to compile because travis-ci doesn't honor go module: https://github.com/kubernetes/kube-openapi/issues/163

sttts:
Do we really want to touch this? I would rather switch to a go-openapi based (or forked) validation in 1.16 and deprecate the whole v2 code path.

sttts:
Also this needs proper testing, another reason I would prefer to not touch it at all, knowing all the limitation and wrong interpretation of OpenAPI. Our server side luckily is coping with all of that.

liggitt:
> Our server side luckily is coping with all of that.

apparently not. a schema with `type:array` and zero `items` schemas got published and caused kubectl failures. we should probably require len(items)==1 for schemas of `type:array` in order to be structural

roycaihw:
talked with @sttts. Closing this in favor of kubernetes/kubernetes#78521

Issue #163, opened 2019-05-30
CI doesn't honor go module and fails presubmit


roycaihw:
https://travis-ci.org/kubernetes/kube-openapi/builds/539009857

travis-ci in this repo doesn't honor go module
```
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
```

instead of fetching the pinged commit, the CI fetches package from HEAD, which includes [public interface change](https://github.com/kubernetes-sigs/structured-merge-diff/commit/f55a48602016b8ecfb93d6a7edc3162e08b91f1b), and fails to compile
```
pkg/schemaconv/smd.go:235:5: unknown field 'DiscriminatedBy' in struct literal of type schema.UnionField
pkg/schemaconv/smd.go:254:3: a.Struct undefined (type *schema.Atom has no field or method Struct)
pkg/schemaconv/smd.go:254:14: undefined: schema.Struct
pkg/schemaconv/smd.go:258:4: a.Struct undefined (type *schema.Atom has no field or method Struct)
pkg/schemaconv/smd.go:271:3: a.Struct undefined (type *schema.Atom has no field or method Struct)
```

cc @apelisse

roycaihw:
the CI fails at the step executing `travis_install_go_dependencies 1.12 -v`, in which [go get](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/bash/travis_install_go_dependencies.bash#L24) is invoked

```
The command "eval go get -v -t ./... " failed 3 times.
The command "go get -v -t ./..." failed and exited with 2 during .
```

`travis_install_go_dependencies` didn't see gomodules enabled because we didn't export the env

PR #164, opened 2019-05-30
make travis export GO111MODULE=on

roycaihw:
travis exports `GO111MODULE="auto"` by default in GOPATH. This change overwrites the environment

fixes https://github.com/kubernetes/kube-openapi/issues/163

### DIFF
--- a/cmd/openapi-gen/args/args.go
+++ b/cmd/openapi-gen/args/args.go
@@ -18,6 +18,7 @@ package args
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/spf13/pflag"
 	"k8s.io/gengo/args"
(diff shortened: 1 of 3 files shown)


cmd/openapi-gen/args/args.go
```diff
@@ -38,6 +39,8 @@ func NewDefaults() (*args.GeneratorArgs, *CustomArgs) {
 	// WithoutDefaultFlagParsing() disables implicit addition of command line flags and parsing,
 	// which allows registering custom arguments afterwards
 	genericArgs := args.Default().WithoutDefaultFlagParsing()
+	genericArgs.GoHeaderFilePath = filepath.Join(args.DefaultSourceTree(), "k8s.io/kube-openapi/boilerplate/boilerplate.go.txt")
```
roycaihw: @apelisse I hit problem missing text file in gengo package: https://travis-ci.org/kubernetes/kube-openapi/builds/539024167. I suspect if the non-go files are ignored by go mod (v.s. previously `go get` persists the files unconditionally)

is there way to ping non-go files in dependency using go mod? Let me know what do you think

apelisse: What you did looks good I think ...

PR #166, opened 2019-07-11
Support x-kubernetes-v2-spec extension

yue9944882:
now in v2 builder, any schema marked the v2-spec extension will be shadowed to the extension value.

/cc @sttts


yue9944882:
@sttts we dont need changes in the generator. e.g. for `IntOrString`, we will add this extension into it's  `OpenAPIDefinition` method. the v3 builder will filter out the extension, the v2 builder will shadow the schema w/ extension value.

pkg/builder/openapi.go
```diff
@@ -154,6 +155,11 @@ func (o *openAPI) buildDefinitionRecursively(name string) error {
 				schema.Extensions[k] = v
 			}
 		}
+		if v, ok := item.Schema.Extensions[extensionV2Spec]; ok {
```
sttts: wondering, can we mirror `OpenAPIV2Definition()` output there instead? so we will assume that `OpenAPIDefinition()` will output v3 schemas (which we down-convert by default), with `OpenAPIV2Definition` for cases where the down-conversion is not good enough.

sttts: or we go the opposite direction: `OpenAPIDefinition` will stay v2, and `OpenAPIV3Definition` can give an advanced v3 schema. That sounds more compatible. We won't have many cases where we need this.

yue9944882: the latter sounds better and i took a shot. updated the integration test a bit cover v2 spec generation.

yue9944882: the next step is to start v3 builder and verify the v2-extension by it's integration test

pkg/common/common.go
```diff
@@ -24,6 +24,12 @@ import (
 	"github.com/go-openapi/spec"
 )
 
+const (
+	// TODO: Make this configurable.
+	ExtensionPrefix = "x-kubernetes-"
+	ExtensionV2Spec = ExtensionPrefix + "v2-spec"
```
sttts: nit: v2-schema ?

pkg/generators/openapi.go
```diff
@@ -242,6 +242,16 @@ func methodReturnsValue(mt *types.Type, pkg, name string) bool {
 	return r.Name.Name == name && r.Name.Package == pkg
 }
 
+func hasOpenAPIV3DefinitionMethod(t *types.Type) bool {
+	for mn, mt := range t.Methods {
+		if mn != "OpenAPIV3Definition" {
+			continue
+		}
+		return methodReturnsValue(mt, openAPICommonPackagePath, "OpenAPIDefinition")
```
sttts: OpenAPIDefinition or OpenAPIV3Definition?

sttts:
Looks good.

/assign @p0lyn0mial 

for some more eyes on this.

pkg/common/common.go
```diff
@@ -172,3 +182,11 @@ func EscapeJsonPointer(p string) string {
 	p = strings.Replace(p, "/", "~1", -1)
 	return p
 }
+
+func EmbedOpenAPIDefinitionIntoV2Extension(main OpenAPIDefinition, embedded OpenAPIDefinition) OpenAPIDefinition {
```
yue9944882: for now, i cant think of a better name for this

pkg/generators/openapi.go
yue9944882: it's the name of struct `OpenAPIDefinition` under "pkg/common" package, instead of the function name

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -25,10 +25,15 @@ package generated
 import (
 	spec "github.com/go-openapi/spec"
 	common "k8s.io/kube-openapi/pkg/common"
+	custom "k8s.io/kube-openapi/test/integration/testdata/custom"
 )
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
+		"k8s.io/kube-openapi/test/integration/testdata/custom.Bac":              common.EmbedOpenAPIDefinitionIntoV2Extension(custom.Bac{}.OpenAPIV3Definition(), custom.Bac{}.OpenAPIDefinition()),
```
p0lyn0mial: so what does it mean?

it means that in the future (once we have `oas v3`) we could have a way to keep and generate both `v2` and `v2 ` schemas (`spec.Schema`)  in just one place, is that correct?

do we expect them to be different (`v2` vs `v3`) ? 

from what I have seen they are quite compatible:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaObject

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject

pkg/builder/openapi.go
p0lyn0mial: @yue9944882 I started working on builderv3, I should issue a WIP PR by the end of the week.

test/integration/pkg/generated/openapi_generated.go
yue9944882: > it means that in the future (once we have oas v3) we could have a way to keep and generate both v2 and v2 schemas (spec.Schema) in just one place, is that correct?

exactly, but only for those types have differences in v2 and v3 schema. for now, it's basically meant for `IntOrString`. in v2 spec, we made walkaround for those types we have to denote in v3 grammar e.g.`anyOf`. the pull will keep those v2 walkarounds in the builder in the extension so that:
- __for v2 publishing endpoint__: we shadow the generated schema using the new extension.
- __for v3 publishing endpoint__: we filter out the extension from the generated schema.


```go

// OpenAPISchemaType is used by the kube-openapi generator when constructing
// the OpenAPI spec of this type.
//
// See: https://github.com/kubernetes/kube-openapi/tree/master/pkg/generators
func (IntOrString) OpenAPISchemaType() []string { return []string{"string"} }

// OpenAPISchemaFormat is used by the kube-openapi generator when constructing
// the OpenAPI spec of this type.
func (IntOrString) OpenAPISchemaFormat() string { return "int-or-string" }
```

for `IntOrString`, the v3 spec will be an `anyOf`

pkg/generators/openapi.go
```diff
@@ -317,14 +339,30 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
...
+
+		if hasV2Definition || (hasV3Definition && !hasV2DefinitionTypeAndFormat) {
 			// already invoked directly
 			return nil
 		}
 
 		args := argsFromType(t)
 		g.Do("func "+nameTmpl+"(ref $.ReferenceCallback|raw$) $.OpenAPIDefinition|raw$ {\n", args)
-		if hasOpenAPIDefinitionMethods(t) {
+		switch {
```
p0lyn0mial: no separate case for `hasV3Definition`?

yue9944882: > no separate case for hasV3Definition?

hasV3Definition case is short-circuited at L329, do you think about adding a case for hasV3Definition and commenting `// this line will never reach` for readability? the branches is getting more complex tho

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -16,7 +16,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by openapi-gen. DO NOT EDIT.
+// Code generated by openapi-gen.go. DO NOT EDIT.
```
yue9944882: nit: minimal change

test/integration/testdata/custom/v3.go
```diff
@@ -0,0 +1,63 @@
+package custom
```
yue9944882: nit: license header

test/integration/testdata/custom/v2.go
```diff
@@ -0,0 +1,19 @@
+package custom
```
yue9944882: nit: license header

pkg/common/common.go
yue9944882: updated

yue9944882:
@sttts @p0lyn0mial PTAL

pkg/generators/openapi.go
p0lyn0mial: no, we can leave it as it is now.

p0lyn0mial:
/lgtm

leaving approval to @sttts

sttts:
@yue9944882 now we need a follow-up which vendors this into kube and which does not lead to any change for the published spec, right?

yue9944882:
> @yue9944882 now we need a follow-up which vendors this into kube and which does not lead to any change for the published spec, right?

yep, this extension is opt-in and backward-compatible. will send the follow-up

PR #170, opened 2019-08-14
aggregator: fix x-kubernetes-group-version-kinds

sttts:
When there was no definition schema conflict, we did not merge the `x-kubernetes-group-version-kind` lists. The PR fixes that.

### DIFF
--- a/pkg/aggregator/aggregator.go
+++ b/pkg/aggregator/aggregator.go
@@ -19,6 +19,7 @@ package aggregator
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/go-openapi/spec"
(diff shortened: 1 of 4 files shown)


pkg/aggregator/aggregator.go
```diff
@@ -185,78 +185,63 @@ func mergeSpecs(dest, source *spec.Swagger, renameModelConflicts, ignorePathConf
...
-			conflicts = true
-			break
+		existing, found := dest.Definitions[k]
+		if !found || deepEqualDefinitionsModuloGVKs(&existing, &v) {
+			// skip for now, we copy them after the rename loop
+			continue
 		}
-	}
 
-	if conflicts {
```
sttts: I removed this optimization (if it ever has been one, am not so sure) because all of our specs have conflicts (due to types like Status, DeleteOptions, ...).

PR #172, opened 2019-08-16
Handle x-kubernetes-preserve-unknown-fields correctly in schemaconv

jennybuckley:
This is necessary for https://github.com/kubernetes/kubernetes/pull/77354

If we are converting a schema for a CRD that supports field pruning at the top level, then any structs recursively below a field with `x-kubernetes-preserve-unknown-fields` need to allow unknown fields (as map items), otherwise ownership will not be tracked for those items and they would fail validation on apply.

If the CRD does not supports field pruning then we should treat every struct in the schema as having `x-kubernetes-preserve-unknown-fields`


apelisse:
That looks good to me.

/lgtm
/approve

Issue #177, opened 2019-11-05
Add support for +mapType, +structType


mariantalla:
The two annotations used to describe the topology of maps and structs, `+mapType` and `+structType`, should result in generating the equivalent openapi extensions (`x-kubernetes-map-type` and `x-kubernetes-struct-type` respectively). This is [documented](https://github.com/kubernetes/kube-openapi/blob/0270cf2f1c1d995d34b36019a6f65d58e6e33ad4/pkg/idl/doc.go) but not implemented at the moment.

/assign
(have a tiny PR in progress at the moment)

sttts:
:+1:

sttts:
Can you do the same addition to controller-tools? (compare https://github.com/kubernetes-sigs/controller-tools/pull/345 for an example). I would like to make sure we use the same markers for both kube-openapi and kubebuilder's controller-tools.

cc @DirectXMan12

mariantalla:
Yep, coming up next (sorry, should have created an issue 🤦‍♀ will go ahead and do that now). Thanks for the pointer too 🙏

PR #178, opened 2019-11-06
Support +mapType, +structType markers

mariantalla:
This PR adds support for markers `// +mapType=<atomic/granular>`, `// +structType=<atomic/granular>`; they should both map to the OpenAPI extension `x-kubernetes-map-type`.

Also adding minor changes to the integration tests README to ensure that
it's up to date after recent changes that add v2 support (https://github.com/kubernetes/kube-openapi/pull/166).

fixes https://github.com/kubernetes/kube-openapi/issues/177

### DIFF
--- a/pkg/idl/doc.go
+++ b/pkg/idl/doc.go
@@ -79,13 +79,15 @@ type ListType string
 type ListMapKey string
 
 // MapType annotates a map to further describe its topology. It may
-// have only one value: "atomic". Atomic means that the entire map is
-// considered as a whole, rather than as distinct values.
+// have one of two values: `atomic` or `granular`. `atomic` means that the entire map is
+// considered as a whole; actors that wish to update the map can only
+// entirely replace it. `granular` means that specific values in the map can be
+// updated separately from other fields.
 //
 // By default, a map will be considered as a set of distinct values that
-// can be updated individually. This default WILL NOT generate any
-// openapi extension, as this will also be interpreted as the default
-// behavior in the openapi definition.
+// can be updated individually (i.e. the equivalent of `granular`).
+// This default WILL NOT generate any openapi extension, as this will also
+// be interpreted as the default behavior in the openapi definition.
 //
 // This tag MUST only be used on maps, or the generation step will fail.
 //
@@ -114,11 +116,13 @@ type PatchMergeKey string
 type PatchStrategy string
 
 // StructType annotates a struct to further describe its topology. It may
-// have only one value: "atomic". Atomic means that the entire struct is
-// considered as a whole, rather than as distinct values.
+// have one of two values: `atomic` or `granular`. `atomic` means that the entire struct is
+// considered as a whole; actors that wish to update the struct can only
+// entirely replace it. `granular` means that specific fields in the struct can be
+// updated separately from other fields.
 //
 // By default, a struct will be considered as a set of distinct values that
-// can be updated individually. This default WILL NOT generate any
+// can be updated individually (`granular`). This default WILL NOT generate any
 // openapi extension, as this will also be interpreted as the default
 // behavior in the openapi definition.
 //
--- a/pkg/idl/doc.go
+++ b/pkg/idl/doc.go
@@ -86,8 +86,7 @@ type ListMapKey string
 //
 // By default, a map will be considered as a set of distinct values that
 // can be updated individually (i.e. the equivalent of `granular`).
-// This default WILL NOT generate any openapi extension, as this will also
-// be interpreted as the default behavior in the openapi definition.
+// This default will still generate an OpenAPI extension with key: "x-kubernetes-map-type".
 //
 // This tag MUST only be used on maps, or the generation step will fail.
 //
@@ -122,9 +121,8 @@ type PatchStrategy string
 // updated separately from other fields.
 //
 // By default, a struct will be considered as a set of distinct values that
-// can be updated individually (`granular`). This default WILL NOT generate any
-// openapi extension, as this will also be interpreted as the default
-// behavior in the openapi definition.
+// can be updated individually (`granular`).
+// This default will still generate an OpenAPI extension with key: "x-kubernetes-map-type".
 //
 // This tag MUST only be used on structs, or the generation step will fail.
 //
--- a/pkg/idl/doc.go
+++ b/pkg/idl/doc.go
@@ -133,7 +133,7 @@ type PatchStrategy string
 // used on any struct.
 //
 // Using this tag will generate the following OpenAPI extension:
-//  "x-kubernetes-struct-type": "atomic"
+//  "x-kubernetes-map-type": "atomic"
 type StructType string
 
 // Union is TBD.
--- a/test/integration/README.md
+++ b/test/integration/README.md
@@ -12,14 +12,14 @@ $ go test -v .
 
 First, run the generator to create `openapi_generated.go` file which specifies
 the `OpenAPIDefinition` for each type, and generate the golden API rule
-violation report file . Note that if you do not pass a report
+violation report file. Note that if you do not pass a report
 filename (`./testdata/golden.v2.report` in the command below) to let the generator
 to print API rule violations to the file, the generator will return error to stderr
 on API rule violations.
 
 ```bash
 $ go run ../../cmd/openapi-gen/openapi-gen.go \
-  -i "k8s.io/kube-openapi/test/integration/testdata/listtype,k8s.io/kube-openapi/test/integration/testdata/maptype,k8s.io/kube-openapi/test/integration/testdata/structtype,k8s.io/kube-openapi/test/integration/testdata/dummytype,k8s.io/kube-openapi/test/integration/testdata/uniontype" \
+  -i "k8s.io/kube-openapi/test/integration/testdata/custom,k8s.io/kube-openapi/test/integration/testdata/listtype,k8s.io/kube-openapi/test/integration/testdata/maptype,k8s.io/kube-openapi/test/integration/testdata/structtype,k8s.io/kube-openapi/test/integration/testdata/dummytype,k8s.io/kube-openapi/test/integration/testdata/uniontype" \
   -o pkg \
   -p generated \
   -O openapi_generated \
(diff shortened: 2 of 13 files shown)


sttts:
What is struct type? In CRDs we only have list-type and map-type.

mariantalla:
`structType` seems to have been added as part of the [server-side apply work](https://github.com/kubernetes/enhancements/blob/f8b41632a7f655371d6e426c8674301b6691e967/keps/sig-api-machinery/0006-apply.md) (for native objects). Unless I got this wrong 🤔

mariantalla:
/hold
there's a mistake in the implementation with structType in any case - `+structType` should also convert to `x-kubernetes-map-type`

pkg/idl/doc.go
```diff
@@ -133,7 +133,7 @@ type PatchStrategy string
 // used on any struct.
 //
 // Using this tag will generate the following OpenAPI extension:
-//  "x-kubernetes-struct-type": "atomic"
+//  "x-kubernetes-map-type": "atomic"
```
apelisse: Above the comment says: 
> It may have only one value: "atomic".

Clearly that's wrong :-)

apelisse: The entire description might be worth fixing.

apelisse: You should probably also verify if the description for maptype and listtype etc are also correct

test/integration/README.md
```diff
@@ -13,26 +13,26 @@ $ go test -v .
 First, run the generator to create `openapi_generated.go` file which specifies
 the `OpenAPIDefinition` for each type, and generate the golden API rule
 violation report file . Note that if you do not pass a report
-filename (`./testdata/golden.report` in the command below) to let the generator
+filename (`./testdata/golden.v2.report` in the command below) to let the generator
```
apelisse: Why are all of these v2 now?

apelisse: Oh I see, you're fixing the doc that is incorrect, thanks.

test/integration/testdata/maptype/atomic-map.go
```diff
@@ -0,0 +1,7 @@
+package maptype
+
+// +k8s:openapi-gen=true
+type AtomicMap struct {
```
apelisse: Are these types supposed to generate something in some golden data? I don't see an update to some golden data

apelisse:
Looks good to me, a few super small nits. @sttts PTAL

pkg/idl/doc.go
mariantalla: > Clearly that's wrong :-)
Oh dear 🤦‍♀ fixed it now (hopefully).

Clarified mapType too, as it had a similar problem (I think I got inspired by it to begin with:) ).

test/integration/README.md
mariantalla: 👍

test/integration/testdata/maptype/atomic-map.go
mariantalla: Hm, good point. Not sure. The steps in the integration tests README don't change the golden data. The only thing that edits `golden.v2.json` is the `builder` script 🤔 I'll dig into it a bit as I don't know this codebase that well - let me know if it's something obvious I'm missing!

apelisse: I'm specifically concerned since the tests don't pass for me but they pass for the CI ...

apelisse: OK I can reproduce the failure on this pull-request while having master work.

apelisse: wait no it passes now ...

apelisse: Ahah it's absolutely terrible. You have to add your objects to:

https://github.com/kubernetes/kube-openapi/blob/master/test/integration/builder/main.go#L105-L117

We should have a way to fail if this is not done somehow ...

mariantalla: Thanks! Added them with https://github.com/kubernetes/kube-openapi/pull/178/commits/a0384dd483d95391e59ea2171ff8c76ac4fb6c69 - though if I'm honest I only partially understand what they're testing 🤔

apelisse: The test generates the openapi corresponding to the types that you have. Now we can look at the `report.v2.json` and see that it generates what we expect from the tags.

apelisse:
Looks good to me now!

mariantalla:
Is there anything else to be done here?

/cc @sttts

test/integration/testdata/golden.v2.json
```diff
@@ -403,6 +479,70 @@
...
+    "required": [
+     "KeyValue"
+    ],
+    "properties": {
+     "KeyValue": {
+      "type": "object",
+      "additionalProperties": {
+       "type": "string"
+      },
+      "x-kubernetes-map-type": "granular"
```
apelisse: The doc above says that we don't generate an openapi extension on the granular case?

test/integration/testdata/structtype/granular-struct.go
```diff
@@ -0,0 +1,8 @@
+package structtype
+
+// +k8s:openapi-gen=true
+type GranularStruct struct {
+	// +structType=granular
```
apelisse: is this in the right place? the structType is annotating a string. Should it fail some sort of validation?
(and same problem above on the `AtomicStruct`)

sttts:
Looks good overall. Leaving lgtm to @apelisse.

test/integration/testdata/structtype/granular-struct.go
mariantalla: Yeah, the annotation is applied to the string; "correctly" (in that `Field` gets the right tag in the produced API spec) but it's super confusing the way I've written it. I'll change it so that the marker annotates a "proper" struct.

mariantalla: Indeed interesting that the validation didn't fail though when using `structType` on a string 🤔 I'll experiment with it and see if there's a problem here 🤔

test/integration/testdata/golden.v2.json
mariantalla: Which doc are you referring to? 🤔 (happy to change it, just to make sure I understand)

pkg/idl/doc.go
```diff
@@ -79,13 +79,15 @@ type ListType string
...
+// considered as a whole; actors that wish to update the map can only
+// entirely replace it. `granular` means that specific values in the map can be
+// updated separately from other fields.
 //
 // By default, a map will be considered as a set of distinct values that
-// can be updated individually. This default WILL NOT generate any
-// openapi extension, as this will also be interpreted as the default
-// behavior in the openapi definition.
+// can be updated individually (i.e. the equivalent of `granular`).
+// This default WILL NOT generate any openapi extension, as this will also
```
apelisse: > Which doc are you referring to?

This doc :-)

apelisse: We talked about it during the meeting, but making my comment "official". I think it's strictly better to have the "granular" explicitly in the openapi than rely on the missing value.

mariantalla: Done 👌 (for `structType` too).

mariantalla:
Let me know when all looks good and I'll squash.

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -25,42 +25,30 @@ package generated
 import (
 	spec "github.com/go-openapi/spec"
 	common "k8s.io/kube-openapi/pkg/common"
-	custom "k8s.io/kube-openapi/test/integration/testdata/custom"
 )
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
-		"k8s.io/kube-openapi/test/integration/testdata/custom.Bac":              common.EmbedOpenAPIDefinitionIntoV2Extension(custom.Bac{}.OpenAPIV3Definition(), custom.Bac{}.OpenAPIDefinition()),
```
apelisse: Why are these gone?

apelisse:
I think this is good, I'm just a little concerned about some of these types disappearing?

test/integration/pkg/generated/openapi_generated.go
mariantalla: huh, no idea. Will re-run the generation and see if I can either reproduce or fix it that way.

mariantalla: I reset the changes in `openapi_generated.go` as the README in test/integration suggests. However it's still strange that the generation omits some types. Taking a look now to see if anything stands out... 🤔

mariantalla: Fixed (I think!) - problem was that the script that generates the API violations report did not get `testdata/custom` as input. User error 🙃

test/integration/README.md
```diff
@@ -12,27 +12,27 @@ $ go test -v .
...
-filename (`./testdata/golden.report` in the command below) to let the generator
+violation report file. Note that if you do not pass a report
+filename (`./testdata/golden.v2.report` in the command below) to let the generator
 to print API rule violations to the file, the generator will return error to stderr
 on API rule violations.
 
 ```bash
 $ go run ../../cmd/openapi-gen/openapi-gen.go \
-  -i "k8s.io/kube-openapi/test/integration/testdata/listtype,k8s.io/kube-openapi/test/integration/testdata/dummytype,k8s.io/kube-openapi/test/integration/testdata/uniontype" \
+  -i "k8s.io/kube-openapi/test/integration/testdata/custom,k8s.io/kube-openapi/test/integration/testdata/listtype,k8s.io/kube-openapi/test/integration/testdata/maptype,k8s.io/kube-openapi/test/integration/testdata/structtype,k8s.io/kube-openapi/test/integration/testdata/dummytype,k8s.io/kube-openapi/test/integration/testdata/uniontype" \
```
apelisse: OK So it looks like it shouldn't have worked before? These tests are requiring people to pay too much attention, we should improve that ...

apelisse:
That looks really good, thank you!
/lgtm
/approve

PR #183, opened 2020-01-16
Bump structured-merge-diff dependency to version 3

jennybuckley:
This is so we can also bump the version used in k/k.

/cc @apelisse


go.mod
```diff
@@ -6,33 +6,30 @@ require (
...
+	github.com/stretchr/testify v1.3.0
 	golang.org/x/net v0.0.0-20170114055629-f2499483f923 // indirect
 	golang.org/x/sys v0.0.0-20170830134202-bb24a47a89ea // indirect
 	golang.org/x/text v0.0.0-20160726164857-2910a502d2bf // indirect
 	golang.org/x/tools v0.0.0-20181011042414-1f849cf54d09 // indirect
 	gopkg.in/yaml.v2 v2.2.1
 	k8s.io/gengo v0.0.0-20190128074634-0689ccc1d7d6
 	k8s.io/klog v0.0.0-20181102134211-b9b56d5dfc92
-	sigs.k8s.io/structured-merge-diff v0.0.0-20190525122527-15d366b2352e
+	sigs.k8s.io/structured-merge-diff/v3 v3.0.0-20200116222232-67a7b8c61874
```
apelisse: This tag doesn't exist, does it? Because to be honest, v3 has not stabilized yet :-)

jennybuckley: No, there is no v3.0.0 tag yet, this was generated when I changed the import line and called `go mod tidy`

Issue #185, opened 2020-01-29
Items fields in List types do not need annotations


lavalamp:
https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/rules/idl_tag.go#L25

This code needs to make an exception for List types (served as a result of a LIST verb and never a target for a mutating operation).

If:
* There's a `Metadata v1meta.ListMeta` sibling field AND
* The field is named "Items" and is a list of something

Then we should actually require that people NOT include a listType= annotation. (and ideally, suppress emitting the tag in the resulting openapi spec)

Why? Otherwise people have to add an exception for this, and I don't want people doing that for nothingburgers, it's bad karma.

mariantalla:
Closing this, let's reopen if needed.

/close

lavalamp:
Do we need to bump the version in k/k to start giving people the new error
messages?

apelisse:
Yeah, I suspect that's also going to force us to remove a lot of the "exceptions", which is always good.

apelisse:
@mariantalla Do you want to take care of it? (not today ;-)

PR #186, opened 2020-02-04
Prevent listType annotation on Items fields in List types

mariantalla:
xref: https://github.com/kubernetes/kube-openapi/issues/185

In the Kubernetes API, Lists are modeled as structs that contain
* an Items field, of type slice
* a ListMeta field.

For this category of slices (i.e. the ones that hold the contents of a
list), we want to require that API developers do NOT specify a +listType
annotation

Signed-off-by: Maria Ntalla <mntalla@pivotal.io>

### DIFF
--- a/pkg/generators/rules/idl_tag.go
+++ b/pkg/generators/rules/idl_tag.go
@@ -24,7 +24,16 @@ func (l *ListTypeMissing) Validate(t *types.Type) ([]string, error) {
 	switch t.Kind {
 	case types.Struct:
 		for _, m := range t.Members {
-			if m.Type.Kind == types.Slice && types.ExtractCommentTags("+", m.CommentLines)[ListTypeIDLTag] == nil {
+			hasListType := types.ExtractCommentTags("+", m.CommentLines)[ListTypeIDLTag] != nil
+
+			if m.Name == "Items" && m.Type.Kind == types.Slice && hasNamedMember(t, "ListMeta") {
+				if hasListType {
+					fields = append(fields, m.Name)
+				}
+				continue
+			}
+
+			if m.Type.Kind == types.Slice && !hasListType {
 				fields = append(fields, m.Name)
 				continue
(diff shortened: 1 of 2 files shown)


mariantalla:
2 questions:

For now:
* we've only checked if there's a "sibling" of `Items` called `ListMeta`, but haven't checked on `ListMeta`'s type. Is that enough? Should we also expect that `ListMeta` is an embedded type? In general, how opinionated do we want to be on the structure of the list struct?
* we've added a requirement that `Items` is of type Slice in the [implementation](https://github.com/kubernetes/kube-openapi/pull/186/files#diff-b320481c072b23d4d7b13fc4911244c4R29) but it's not covered in tests. Thinking about it a bit more, attempting to apply `+listType` on a type that is not slice _is_ an error, but due to a different reason. Thoughts?

lavalamp:
Yeah this is a big improvement. No one should be embedding any of our List types inside other types, right? If so we have a problem because namespace & name (the keys for the list-typed map) are nested inside metadata.

/lgtm
/approve

pkg/generators/rules/idl_tag.go
```diff
@@ -24,13 +24,30 @@ func (l *ListTypeMissing) Validate(t *types.Type) ([]string, error) {
 	switch t.Kind {
 	case types.Struct:
 		for _, m := range t.Members {
-			if m.Type.Kind == types.Slice && types.ExtractCommentTags("+", m.CommentLines)[ListTypeIDLTag] == nil {
+			hasListType := types.ExtractCommentTags("+", m.CommentLines)[ListTypeIDLTag] != nil
+
+			if m.Name == "Items" && m.Type.Kind == types.Slice && hasNamedMember(t, "ListMeta") {
```
caesarxuchao, line 29: We should change the error message in this case. Currently the message is "list_type_missing", which is the opposite.

Issue #117, opened 2018-11-10
Remove klog dependency


roycaihw:
The error surfacing in this package should be self evident. 

A self-contained system like Kubernetes which imports kube-openapi may use logging module (glog/klog) to do V-style logging. However this package don't have a good reason to use glog/klog, which interferes/makes assumption about logging mechanism used by downstream systems.

ref: https://github.com/kubernetes/kube-openapi/pull/115

/good-first-issue
/help

palash25:
Hi @roycaihw new contributor here :wave: would like to take a crack at this issue. Just need clarification on one thing, if we don't want logging what would you want it to be replaced by? `print` statements?

devdattakulkarni:
@roycaihw Can you elaborate on what would it take to resolve this issue?

PR #115, opened 2018-11-04
Switch from glog to klog

dims:
Change-Id: Ie5d0f09425c945da9bf24474a212adf3148daf6f

PR #187, opened 2020-02-13
Remove klog dependency

Dean-Coakley:
Fixes: https://github.com/kubernetes/kube-openapi/issues/117

Remove `klog` dep as mentioned in https://github.com/kubernetes/kube-openapi/pull/115

Quite unsure if I did this correctly. Happy to take feedback.
Is this the change ye intended?: @roycaihw  @lavalamp

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	golang.org/x/tools v0.0.0-20181011042414-1f849cf54d09 // indirect
 	gopkg.in/yaml.v2 v2.2.1
 	k8s.io/gengo v0.0.0-20190128074634-0689ccc1d7d6
-	k8s.io/klog v0.0.0-20181102134211-b9b56d5dfc92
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -29,7 +29,7 @@ require (
 	golang.org/x/tools v0.0.0-20181011042414-1f849cf54d09 // indirect
 	gopkg.in/yaml.v2 v2.2.1
 	k8s.io/gengo v0.0.0-20190128074634-0689ccc1d7d6
-	k8s.io/klog v0.0.0-20181102134211-b9b56d5dfc92
+	k8s.io/klog v0.0.0-20181102134211-b9b56d5dfc92 // indirect
```
roycaihw: Thanks for your contribution! As https://github.com/kubernetes/kube-openapi/pull/134#issuecomment-459689388 pointed out, klog is a transitive dependency from gengo. Without changing gengo, this does help us get rid of the dep

/close

PR #189, opened 2020-03-31
update gnostic dependency

liggitt:
Starts the chain of updates required to get k8s.io/kubernetes on the 0.4.1 release


test/integration/pkg/generated/openapi_generated.go
```diff
@@ -244,8 +249,10 @@ func schema_test_integration_testdata_listtype_MapList(ref common.ReferenceCallb
 					"Field": {
 						VendorExtensible: spec.VendorExtensible{
 							Extensions: spec.Extensions{
-								"x-kubernetes-list-map-keys": "port",
-								"x-kubernetes-list-type":     "map",
+								"x-kubernetes-list-map-keys": []interface{}{
```
apelisse, line 252: This is correct, but why did it change?

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -325,6 +332,142 @@ func schema_test_integration_testdata_listtype_UntypedList(ref common.ReferenceC
 	}
 }
 
+func schema_test_integration_testdata_maptype_AtomicMap(ref common.ReferenceCallback) common.OpenAPIDefinition {
```
apelisse, line 335: Same for this? Did we forget to commit that before? I also remember that somewhere it says we shouldn't commit updates to this file, but that doesn't really make sense ...

liggitt: yes, it was stale on master... needs a presubmit check

apelisse: I'm pretty sure somewhere it says it shouldn't be committed, we could also remove that comment ...

apelisse: https://github.com/kubernetes/kube-openapi/blob/master/test/integration/README.md

> After the golden spec is generated, please clean up the generated file pkg/generated/openapi_generated.go before you commit. It's an intermediate product that doesn't need to be updated in kube-openapi repository. The checked-in file is kept minimum to make sure that test/integration/builder compiles. Please run:
> ```
> $ git checkout pkg/generated/openapi_generated.go
> ```
> to discard any local change.

PR #191, opened 2020-04-03

liggitt:
Resolves an error trying to validate data containing an integer literal, unmarshaled using the integer-preserving https://github.com/kubernetes/apimachinery/blob/v0.18.0/pkg/util/json/json.go#L41-L43 method, against a schema field defined to accept numbers (floats or ints)

See https://swagger.io/docs/specification/data-models/data-types/#numbers:

> OpenAPI has two numeric types, number and integer, where number includes both integer and floating-point numbers.

### DIFF
--- a/pkg/util/proto/testdata/swagger.json
+++ b/pkg/util/proto/testdata/swagger.json
@@ -6,6 +6,621 @@
   },
   "paths": {},
   "definitions": {
+    "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceColumnDefinition": {
+      "description": "CustomResourceColumnDefinition specifies a column for server side printing.",
+      "properties": {
+        "description": {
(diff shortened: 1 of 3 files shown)


pkg/util/proto/testdata/swagger.json
```diff
@@ -6,6 +6,621 @@
   },
   "paths": {},
   "definitions": {
+    "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceColumnDefinition": {
```
apelisse, line 9: Was this copied from somewhere or generated from something in that repo?

apelisse: Or did you type all these characters with your keyboard?

liggitt: copied from k/k master swagger.json

PR #193, opened 2020-04-09

liggitt:
Pick of https://github.com/kubernetes/kube-openapi/pull/191 for 1.18

/cc @apelisse


apelisse:
I can't believe you typed all these characters again!
/lgtm
/approve

PR #194, opened 2020-04-09

liggitt:
Pick of https://github.com/kubernetes/kube-openapi/pull/191 for 1.17

/cc @apelisse


apelisse:
Ah, you edited that 1.18 before I could comment!
/lgtm
/approve

PR #195, opened 2020-04-09

liggitt:
Pick of https://github.com/kubernetes/kube-openapi/pull/191 for 1.16

/cc @apelisse


liggitt:
kube PR passed with this - https://github.com/kubernetes/kubernetes/pull/90022

PR #201, opened 2020-08-07
fix handling of inlined embedded struct pointer fields

mandelsoft:
Inlined embedded struct pointer fields are ignored during open api generation:

```go
// Blah demonstrate a struct with embedded inline struct pointer field.
type Nested struct {
  // A simple string
  String string
}

type Blah struct {
  // An embedded inline struct pointer field
  *Nested `json:",inline,omitempty"`
}
```

The reason is that types of inlined types are directly searched for members.
The fix just follows pointer types until a non-pointer type before examining the type for members.

There are no tests for inline types. Those are added, including one test with an inlined pointer type,

### DIFF
--- a/pkg/generators/openapi_test.go
+++ b/pkg/generators/openapi_test.go
@@ -347,12 +347,13 @@ func TestNestedStruct(t *testing.T) {
 	callErr, funcErr, assert, callBuffer, funcBuffer := testOpenAPITypeWriter(t, `
 package foo
 
-// Blah demonstrate a struct with struct field.
+// Nested is used as struct field
 type Nested struct {
   // A simple string
   String string
 }
 
+// Blah demonstrate a struct with struct field.
 type Blah struct {
   // A struct field
   Field Nested
@@ -370,6 +371,7 @@ type Blah struct {
 return common.OpenAPIDefinition{
 Schema: spec.Schema{
 SchemaProps: spec.SchemaProps{
+Description: "Blah demonstrate a struct with struct field.",
 Type: []string{"object"},
 Properties: map[string]spec.Schema{
 "Field": {
@@ -394,12 +396,13 @@ func TestNestedStructPointer(t *testing.T) {
 	callErr, funcErr, assert, callBuffer, funcBuffer := testOpenAPITypeWriter(t, `
 package foo
 
-// Blah demonstrate a struct with struct pointer field.
+// Nested is used as struct pointer field
 type Nested struct {
   // A simple string
   String string
 }
 
+// Blah demonstrate a struct with struct pointer field.
 type Blah struct {
   // A struct pointer field
   Field *Nested
@@ -417,6 +420,7 @@ type Blah struct {
 return common.OpenAPIDefinition{
 Schema: spec.Schema{
 SchemaProps: spec.SchemaProps{
+Description: "Blah demonstrate a struct with struct pointer field.",
 Type: []string{"object"},
 Properties: map[string]spec.Schema{
 "Field": {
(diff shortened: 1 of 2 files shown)


rfranzke:
@lavalamp @jpbetz @mariantalla @windayski could you take a look if you have some time, please?

ialidzhikov:
@sts @lavalamp @liggitt could you please help checking this PR?

/kind bug
/priority important-soon

ialidzhikov:
Similar PR for go-to-protobuf - https://github.com/kubernetes/kubernetes/pull/90296.

lavalamp:
Apologies, I didn't see this until now...

/assign @apelisse

pkg/generators/openapi_test.go
```diff
@@ -343,6 +343,239 @@ Type: []string{"object"},
 `, funcBuffer.String())
 }
 
+func TestNestedStruct(t *testing.T) {
+	callErr, funcErr, assert, callBuffer, funcBuffer := testOpenAPITypeWriter(t, `
+package foo
+
+// Blah demonstrate a struct with struct field.
```
apelisse, line 350: Comment seems wrong, this is not "Blah"

pkg/generators/openapi_test.go
```diff
@@ -343,6 +343,239 @@ Type: []string{"object"},
...
+}
+
+`, funcBuffer.String())
+}
+
+func TestNestedStructPointer(t *testing.T) {
+	callErr, funcErr, assert, callBuffer, funcBuffer := testOpenAPITypeWriter(t, `
+package foo
+
+// Blah demonstrate a struct with struct pointer field.
```
apelisse, line 397: Again, comment is not at the right place

pkg/generators/openapi_test.go
```diff
@@ -343,6 +343,239 @@ Type: []string{"object"},
...
+	}
+	assert.Equal(`"base/foo.Blah": schema_base_foo_Blah(ref),
+`, callBuffer.String())
+	assert.Equal(`func schema_base_foo_Blah(ref common.ReferenceCallback) common.OpenAPIDefinition {
+return common.OpenAPIDefinition{
+Schema: spec.Schema{
+SchemaProps: spec.SchemaProps{
+Type: []string{"object"},
+Properties: map[string]spec.Schema{
+"Nested": {
```
apelisse, line 469: is this correct? I don't think Nested should be the name of a field?

ialidzhikov:
@mandelsoft , did you managed to check the suggestions?

ialidzhikov:
ping @mandelsoft

pkg/generators/openapi_test.go
mandelsoft: Because it is not indicated as inlined the name of the embedded type should be the name of the field. That's consistent with the usage in the go language, where fields of embedded fields can be denoted by the the type name followed by the desired sub field

pkg/generators/openapi_test.go
mandelsoft: There are several such locations. The comment was intended for the complete test referring to the included type `Blah`. In this sense the comment is correct.  But you are completely right. In GO such a comment is always related to the directly following type declaration, which is misleading here. I'll fix this.

pkg/generators/openapi_test.go
mandelsoft: That's still the old behaviour, I just added tests for all cases, including the already correctly handled ones.

apelisse: Oh yes, that's probably because it doesn't have `json:",inline"`.

pkg/generators/openapi_test.go
```diff
@@ -343,6 +343,249 @@ Type: []string{"object"},
...
+Required: []string{"String"},
+},
+},
+}
+}
+
+`, funcBuffer.String())
+}
+
+func TestEmbeddedInlineStructPointer(t *testing.T) {
```
apelisse, line 541: I'm guessing this test is the one that would break before the fix?

apelisse (commented):
/lgtm
/approve

Thanks!

PR #204, opened 2020-08-11
Fix openAPI x-kubernetes-map-type enum to be "granular"

jpbetz:
Mistakenly set it to "separable" in https://github.com/kubernetes/kubernetes/pull/93901. It needs to be "granular".

see https://github.com/kubernetes/kubernetes/pull/93901#discussion_r468848080

/cc @apelisse @liggitt 
/sig api-machinery
/priority important-soon

### DIFF
--- a/pkg/schemaconv/smd.go
+++ b/pkg/schemaconv/smd.go
@@ -318,7 +318,7 @@ func (c *convert) VisitKind(k *proto.Kind) {
 		switch val {
 		case "atomic":
 			a.Map.ElementRelationship = schema.Atomic
-		case "separable":
+		case "granular":
 			a.Map.ElementRelationship = schema.Separable
 		default:
 			c.reportError("unknown map type %v", val)
@@ -401,7 +401,7 @@ func (c *convert) VisitMap(m *proto.Map) {
 		switch val {
 		case "atomic":
 			a.Map.ElementRelationship = schema.Atomic
-		case "separable":
+		case "granular":
 			a.Map.ElementRelationship = schema.Separable
 		default:
 			c.reportError("unknown map type %v", val)


apelisse:
To completely remove the inconsistency and confusion, I was expecting you to just completely get rid of the "separable" word?

jpbetz:
> To completely remove the inconsistency and confusion, I was expecting you to just completely get rid of the "separable" word?

That's an independent change from this though, right? That needs to be changed in SMD and then picked up here.

apelisse:
Right, yes, I thought we were in SMD ...

jpbetz:
> Right, yes, I thought we were in SMD ...

And we might cherrypick so keeping the change to the minimum required here will make that more contained..

liggitt:
it's slightly concerning that no test failed with this change, and no test failed in https://github.com/kubernetes/kubernetes/pull/93901

I would have expected us to exercise the explicit "granular" value in a way that would have been caught by a test somewhere

apelisse:
Yeah, that feature was not implemented and not function before that, that's what Joe is actually working on.

liggitt:
> Yeah, that feature was not implemented and not functional before that, that's what Joe is actually working on.

it seems the kube-openapi code that honors x-kubernetes-map-type is not being exercised in tests with a schema containing `x-kubernetes-map-type: granular`, which surprised me in kubernetes/kubernetes

liggitt:
to be confident that code does the right thing, we should be sure to exercise it with explicit "granular" schemas

apelisse:
Yes, we're using the kubernetes OpenAPI as the main source of tests and we don't have such an occurence of `granular` there, so that's probably why it didn't fail here.

PR #210, opened 2020-10-09
Generate OpenAPI default field from +default comment marker

apelisse:


pkg/generators/openapi.go
```diff
@@ -513,6 +515,19 @@ func (g openAPITypeWriter) validatePatchTags(m *types.Member, parent *types.Type
...
 }
 
+func (g openAPITypeWriter) generateDefault(m *types.Member) error {
+	tag, err := getSingleTagsValue(m.CommentLines, tagDefault)
+	if tag == "" {
+		return err
+	}
+	var i interface{}
+	if err := json.Unmarshal([]byte(tag), &i); err != nil {
+		return fmt.Errorf("failed to unmarshal default: %v", err)
```
jpbetz, line 525: nit: also include the member name and tag that failed to parse in error? Would be great if the type name the member is from could be included too.

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -67,6 +69,80 @@ func schema_test_integration_testdata_custom_Bah(ref common.ReferenceCallback) c
...
+					"OtherField": {
+						SchemaProps: spec.SchemaProps{
+							Default: 0,
+							Type:    []string{"integer"},
+							Format:  "int32",
+						},
+					},
+					"list": {
+						SchemaProps: spec.SchemaProps{
+							Default: []interface{}{"foo", "bar"},
```
sttts, line 94: Do we make sure that these are not published to clients?

Before vendoring this into kube this must be put in place.

apelisse: Curious why we wouldn't publish that to clients?

sttts: To avoid bad behaviour in the ecosystem: like clients that take action on default values, e.g. default objects themselves instead of on the server.

apelisse: We can't prevent people from doing the wrong thing. I think the more description the API is the better.

sttts: I don't agree. Information hiding is a good thing if you don't want to create inofficially used APIs.

sttts: also I bet that OpenAPI compatible/generated clients will apply defaults. And our default are not what they think they are. They have non-standard semantics.

apelisse: I'm 100% sure that the clients will NOT apply these defaults because that's precisely AGAINST the definition of the field, which is merely to document what the server will do:

> The default value is the one that the server uses if the client does not supply the parameter value in the request.

We want to be able to describe programmatically how our APIs work.

apelisse: But yeah, I agree that we don't want clients to apply the defaults for us.

pkg/generators/openapi.go
apelisse: Good point, I'll add that.

apelisse: Just ran a test to see what it looks like:
```
2020/10/16 09:39:26 OpenAPI code generation error: Failed executing generator: some packages had errors:
failed to generate default: failed to unmarshal default: invalid character 'd' in literal false (expecting 'a')
```

That's not great, let me see what I can do.

apelisse: Actually, I missed the line before that, which indicates the name of the field. More context can still be added, runs like this now:
```
E1016 09:47:40.459671   78052 openapi.go:309] Error when generating: sub, sub *k8s.io/kube-openapi/test/integration/testdata/defaults.SubStruct
2020/10/16 09:47:40 OpenAPI code generation error: Failed executing generator: some packages had errors:
failed to generate default in k8s.io/kube-openapi/test/integration/testdata/defaults.Defaulted: failed to unmarshal default: invalid character 'd' in literal false (expecting 'a')
```

pkg/common/common.go
```diff
@@ -105,28 +105,34 @@ type Config struct {
...
-	"float32":     {"number", "float"},
-	"bool":        {"boolean", ""},
-	"time.Time":   {"string", "date-time"},
-	"string":      {"string", ""},
-	"integer":     {"integer", ""},
-	"number":      {"number", ""},
-	"boolean":     {"boolean", ""},
-	"[]byte":      {"string", "byte"}, // base64 encoded characters
-	"interface{}": {"object", ""},
+type openAPIInfo struct {
```
sttts, line 108: `type typeInfo`

pkg/common/common.go
```diff
@@ -173,7 +179,17 @@ func GetOpenAPITypeFormat(typeName string) (string, string) {
 	if !ok {
 		return "", ""
 	}
-	return mapped[0], mapped[1]
+	return mapped.name, mapped.format
+}
+
+// Returns the zero-value for the given type along with true if the type
+// could be found.
+func GetOpenAPIZeroValue(typeName string) (interface{}, bool) {
```
sttts, line 187: s/Get//

apelisse: I don't know, the function above is called `GetOpenAPITypeFormat`, I just followed the same pattern.

pkg/common/common.go
apelisse: Fixed!

pkg/common/common.go
sttts: My standard comment when I see GetFoo in golang code: the idiomatic function name for those mappings without state is `Foo`.

apelisse: Yes, the question is: do we want to be correct or consistent? :-)

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,92 @@
...
+import "github.com/go-openapi/spec"
+
+// PruneDefaults remove all the defaults recursively from all the
+// schemas in the definitions, and does not modify the definitions in
+// place.
+func PruneDefaults(definitions spec.Definitions) spec.Definitions {
+	if definitions == nil {
+		return nil
+	}
+	newDefinitions := spec.Definitions{}
```
sttts, line 28: make with same capacity as the old length

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,92 @@
...
+
+	for name, definition := range definitions {
+		newDefinitions[name] = *prune_schema_defaults(&definition)
+	}
+
+	return newDefinitions
+}
+
+func prune_schemalist_defaults(schemas []spec.Schema) []spec.Schema {
```
sttts, line 37: I would prefer to see some on-demand allocation of a new list, i.e. turn all the signatures here into `func prune_*(*Foo) (bool, *Foo)` with the bool false if nothing changed.

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,92 @@
...
+	if schemas == nil {
+		return nil
+	}
+	newSchemas := []spec.Schema{}
```
sttts, line 41: make with capacity

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,92 @@
...
+		newSchemas = append(newSchemas, *prune_schema_defaults(&schema))
+	}
+	return newSchemas
+}
+
+func prune_schemamap_defaults(schemas map[string]spec.Schema) map[string]spec.Schema {
+	if schemas == nil {
+		return nil
+	}
+	newMap := map[string]spec.Schema{}
```
sttts, line 52: make with capacity

pkg/handler/internal/default_pruning_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"encoding/json"
+	"reflect"
+	"testing"
+
+	"k8s.io/kube-openapi/pkg/handler/internal"
+
+	"github.com/go-openapi/spec"
+)
+
+func TestDefaultPruning(t *testing.T) {
```
sttts, line 29: not sure about the necessary effort: one could 
1. fuzz a definition, call your pruning func, marshal
2. call marshal on the original definition and use string replacement to remove defaults
3. compare

sttts: Also with some clever logic you can check that your pruning func does not change things in-place.

pkg/handler/internal/default_pruning.go
apelisse: I've tried to come-up with a reasonable algorithm here, but that seems complicated compared to the possible benefits.

The problem is that we don't know if we need to allocate the list/map until we know if something has changed in the list. And we can't check the list and then do it without causing squared complexity over the depth of the tree.

AFAICT, we'd have to initially assume that none are going to change, and if we find one, allocate the list, copy the items we've already seen and continue with that new list, which again, seems a little overkill here.

But maybe I'm missing something?

pkg/handler/internal/default_pruning_test.go
apelisse: Yeah, I'm happy with that level of testing, especially considering that the timeline is a little short.

pkg/handler/internal/default_pruning.go
sttts: Compare https://github.com/kubernetes/kube-openapi/blob/d8b685cd2daadc40121e864b5e87ca8ededd9eed/pkg/aggregator/mutating_walker.go#L33. Something similar for defaults.

pkg/common/common.go
```diff
@@ -105,28 +105,34 @@ type Config struct {
...
-	"[]byte":      {"string", "byte"}, // base64 encoded characters
-	"interface{}": {"object", ""},
+type typeInfo struct {
+	name   string
+	format string
+	zero   interface{}
+}
+
+var schemaTypeFormatMap = map[string]typeInfo{
+	"uint":        {"integer", "int32", 0.},
```
lavalamp, line 115: Why are all the number zero values floating point?

pkg/generators/openapi_test.go
```diff
@@ -785,7 +811,73 @@ type Blah struct {
...
+		},
+		{
+			definition: `
+package foo
+
+type Blah struct {
+	// +default={"foo": 5}
+	Struct struct{
+		foo int
+        }
```
lavalamp, line 841: nit: spacing is messed up

pkg/generators/openapi_test.go
```diff
@@ -131,6 +131,7 @@ type Blah struct {
 	WithStructTagExtension string `+"`"+`patchStrategy:"merge" patchMergeKey:"pmk"`+"`"+`
 	// a member with a list type
 	// +listType=atomic
+	// +default=["foo", "bar"]
```
lavalamp, line 134: You tested the failure cases below, but I didn't see tests for the success cases like this one?

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,206 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+import "github.com/go-openapi/spec"
+
+// PruneDefaults remove all the defaults recursively from all the
```
lavalamp, line 21:
```suggestion
// PruneDefaults removes all the defaults recursively from all the
```

pkg/handler/internal/default_pruning.go
```diff
@@ -0,0 +1,206 @@
...
+// schemas in the definitions, and does not modify the definitions in
+// place.
+func PruneDefaults(definitions spec.Definitions) spec.Definitions {
+	definitionsCloned := false
+	for k, v := range definitions {
+		if s := pruneSchema(&v); s != &v {
```
lavalamp, line 27: What if you were to always copy and just e.g. cache results, so this only happens once per schema? (I'm assuming this runs on CRD schemas and can happen repeatedly)

The test didn't convince me that this copy happens iff it's needed and that the original isn't modified at all. :)

pkg/handler/internal/default_pruning_test.go
lavalamp: Can you at least add an already-correct schema and verify that it isn't changed further?

apelisse: Yeah, that's a good point!

pkg/common/common.go
apelisse: Because `json.Unmarshal` turns all numbers into floats.

pkg/generators/openapi_test.go
apelisse: good catch :-) spaces vs tabs

test/integration/pkg/generated/openapi_generated.go
lavalamp: @sttts can you ack this point? I don't want to merge if there's a pending disagreement, Antoine's answer sounds convincing to me.

test/integration/testdata/golden.v2.report
```diff
@@ -1,4 +1,13 @@
+API rule violation: list_type_missing,k8s.io/kube-openapi/test/integration/testdata/defaults,Defaulted,List
```
lavalamp, line 1: Is there a reason to have the example violate the rules? :)

pkg/handler/internal/default_pruning.go
lavalamp: Test seems better now.

test/integration/pkg/generated/openapi_generated.go
apelisse: I ended-up removing the defaults as Stefan recommended. We'll revisit that holistically later.

PR #212, opened 2020-10-31
Converts OpenAPI defaults to structured-merge-diff schema

apelisse:
Take the defaults from OpenAPI and push them into the smd schema, only for struct fields since we only care about these.
Also, update smd to its latest version since we need the most recent changes.


apelisse:
OK I think that's ready and does the right thing. I'll try to assemble everything we have right now in a WIP Kubernetes PR.

jpbetz:
This looks pretty straightforward. I didn't see anything that needs changing.

/lgtm

PR #213, opened 2020-11-05
Fix unexpected error values

gautierdelorme:
Porting  https://github.com/go-openapi/errors/pull/21 and https://github.com/go-openapi/validate/pull/126 to fix https://github.com/kubernetes/kubernetes/issues/90128 (see https://kubernetes.slack.com/archives/C0EG7JC6T/p1592287631273100 for additional context).

Signed-off-by: Gautier Delorme <gautier.delorme@gmail.com>


pkg/validation/errors/schema.go
```diff
@@ -483,7 +485,7 @@ func EnumFail(name, in string, value interface{}, values []interface{}) *Validat
 }
 
 // Required error for when a value is missing
-func Required(name, in string) *Validation {
+func Required(name, in string, value interface{}) *Validation {
```
sttts, line 488: does this need the argument?

gautierdelorme: It's not super important for `Required` but it might help differentiate `""` vs `nil` for example.

dims:
/lgtm

thanks @gautierdelorme

pkg/validation/errors/schema.go
sttts: For CRDs "" and null are both values that fulfill required

gautierdelorme: Fixed and I also removed the unused `stringValidator` logic in https://github.com/kubernetes/kube-openapi/pull/213/commits/dc2906fe7cba3e29a77142b46d47f05c439aa781 as discussed in Slack.

PR #214, opened 2020-11-06
validation: add OWNERS

sttts:
This code is sensitive because it is both reasonable complex and has influence on the API. Hence, we add an OWNERS file to restrict the set of people as reviewers+approvers who are deeply familiar with the validation logic.

### DIFF
--- a/pkg/validation/OWNERS
+++ b/pkg/validation/OWNERS
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+options:
+  no_parent_owners: true
+approvers:
+- sttts
+- liggitt
+reviewers:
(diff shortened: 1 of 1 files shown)


pkg/validation/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+options:
+  no_parent_owners: true
+filters:
```
liggitt, line 4: filters doesn't work for approvers, should just be 

```
approvers:
...
reviewers:
...
```

sttts: oh, right. Copy&paste from api repo. But there it was restricted to different file types.

Issue #206, opened 2020-08-18
Update to gnostic v0.5.1?


eclipseo:
I'd like to use kube-openapi with latest gnostic (0.5.x) but there is an incompatibility between yaml v2 used by kube-openapi and yaml v3 used by gnostic since version 0.5.0: https://github.com/googleapis/gnostic/issues/195

May you please update kube-openapi to use the latest gnostic (0.5.1)?

PR #220, opened 2021-01-04
Update gnostic to v0.5.1

howardjohn:
Context: https://github.com/google/gnostic/issues/195
Fixes https://github.com/kubernetes/kube-openapi/issues/206

Without this, it is impossible to import this library and
controller-runtime v0.7.0


sttts:
Why does controller-runtime depend on this package?

howardjohn:
It doesn't, but it's impossible to depend on both of them without this.

cbandy:
> It doesn't, but it's impossible to depend on both of them without this.

There is `k8s.io/apimachinery` which in turn refers to `k8s.io/kube-openapi`.

https://github.com/kubernetes/kubernetes/blob/v1.20.0/staging/src/k8s.io/apimachinery/go.mod#L37
https://github.com/kubernetes/kubernetes/blob/v1.20.0/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/meta.go#L26

```
● controller-runtime (v0.7.0) head -n1 go.mod
module sigs.k8s.io/controller-runtime

● controller-runtime (v0.7.0) go mod why -m k8s.io/kube-openapi
# k8s.io/kube-openapi
sigs.k8s.io/controller-runtime/pkg/client/fake
k8s.io/client-go/testing
k8s.io/apimachinery/pkg/util/strategicpatch
k8s.io/kube-openapi/pkg/util/proto
```

howardjohn:
I also applied this patch on to our project before attempting to update to controller-runtime 0.7.0 (ie using gnostic v0.4), and it works fine. So this should be a backwards compatible change for those users, while allowing users to use gnostic v0.5

howardjohn:
Any updates? This is blocking adoption of newer Kubernetes client libraries

go.mod
```diff
@@ -12,18 +12,17 @@ require (
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/spec v0.19.3
 	github.com/go-openapi/swag v0.19.5
-	github.com/golang/protobuf v1.3.3
-	github.com/google/go-cmp v0.4.0 // indirect
+	github.com/golang/protobuf v1.4.2
```
apelisse, line 15: @yliaog Can we do that?

apelisse:
Have you tried to import this in kubernetes to see if it works? Probably by doing a `go mod replace` to your own fork?

howardjohn:
I did it in istio/istio which imports pretty much every k8s package except
k/k. Let me try k/k

go.mod
yliaog: it is used to marshal document to protobuf, so upgrading to v1.4.2 should be ok.

https://github.com/kubernetes/kube-openapi/blob/1a53ea80ebf47de9717f9a548c26f3d894905964/pkg/handler/handler.go#L188

proto.Marshal(document)

howardjohn:
I tested k/k. It works fine with one small change (mirroring the one done here):
```diff
diff --git a/staging/src/k8s.io/apiserver/pkg/util/openapi/proto.go b/staging/src/k8s.io/apiserver/pkg/util/openapi/proto.go
index 0f95a5f7946..e724e79780f 100644
--- a/staging/src/k8s.io/apiserver/pkg/util/openapi/proto.go
+++ b/staging/src/k8s.io/apiserver/pkg/util/openapi/proto.go
@@ -20,9 +20,7 @@ import (
        "encoding/json"

        "github.com/go-openapi/spec"
-       "github.com/googleapis/gnostic/compiler"
        openapi_v2 "github.com/googleapis/gnostic/openapiv2"
-       yaml "gopkg.in/yaml.v2"

        "k8s.io/kube-openapi/pkg/util/proto"
 )
@@ -34,13 +32,7 @@ func ToProtoModels(openAPISpec *spec.Swagger) (proto.Models, error) {
                return nil, err
        }

-       var info yaml.MapSlice
-       err = yaml.Unmarshal(specBytes, &info)
-       if err != nil {
-               return nil, err
-       }
-
-       doc, err := openapi_v2.NewDocument(info, compiler.NewContext("$root", nil))
+       doc, err := openapi_v2.ParseDocument(specBytes)
        if err != nil {
                return nil, err
        }
```

yliaog:
the changes for k8s.io/apiserver/pkg/util/openapi/proto.go is about yaml, i don't understand why that change is needed, or how it helps.

howardjohn:
See https://github.com/google/gnostic/issues/195 for more info. Basically, they changed from requiring yaml.v2 input to yaml.v3. The correct fix is to use their new abstracted ParseDocument instead

yliaog:
i see. so it is not caused by this PR. it is a separate issue of its own.

howardjohn:
Its caused by a (breaking) change in gnostic in 0.4 -> 0.5. This PR upgrades gnostic to 0.5, so this PR is the cause.

yliaog:
Got it. i was thinking about github.com/golang/protobuf upgrade in this PR, we'd like to know whether it would cause breakages in k/k

cbandy:
It's too bad that 0.4 didn't add a `ParseDocument()` to ease this transition.

howardjohn:
What action is needed here?

apelisse:
I'm fine moving forward with this, what do you think @yliaog ?

cmd/openapi2smd/openapi2smd.go
```diff
@@ -33,12 +33,12 @@ func main() {
 		log.Fatal("this program takes input on stdin and writes output to stdout.")
 	}
 
-	var info yaml.MapSlice
-	if err := yaml.NewDecoder(os.Stdin).Decode(&info); err != nil {
+	input, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
 		log.Fatalf("error decoding stdin: %v", err)
```
yliaog, line 38: it is not decoding stdin, simply reading stdin

yliaog:
it looks fine, PR needs rebase

howardjohn:
@yliaog PR is rebased

apelisse:
I can't oppose, I'll let @yliaog lgtm
/approve

yliaog:
do you have a pr for the small change mentioned in https://github.com/kubernetes/kube-openapi/pull/220#issuecomment-768383410 ?

howardjohn:
@yliaog I do not, I think it depends on this being merged?

yliaog:
could you please have the PR ready? refer to this PR. This PR can merge soon. Thanks.

howardjohn:
Thanks, I opened https://github.com/kubernetes/kubernetes/pull/100490

PR #222, opened 2021-01-13
Trim vendor prefix from openapi names

liggitt:
xref https://github.com/kubernetes/kubernetes/pull/98027

/cc @mikedanese @apelisse

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -80,6 +80,9 @@ func ToRESTFriendlyName(name string) string {
 // Example for vendored Go type:
 //     Original full path:  k8s.io/kubernetes/vendor/k8s.io/api/core/v1.Pod
 //     Canonical name:      k8s.io/api/core/v1.Pod
+//
+//     Original full path:  vendor/k8s.io/api/core/v1.Pod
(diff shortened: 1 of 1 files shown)


pkg/util/util.go
```diff
@@ -80,6 +80,9 @@ func ToRESTFriendlyName(name string) string {
 // Example for vendored Go type:
 //     Original full path:  k8s.io/kubernetes/vendor/k8s.io/api/core/v1.Pod
 //     Canonical name:      k8s.io/api/core/v1.Pod
+//
+//     Original full path:  vendor/k8s.io/api/core/v1.Pod
```
apelisse, line 84: It took me way longer than it should do find the difference between these two and the two above :-)

apelisse (commented):
is there any way to test that? ;-)

/lgtm
/approve

liggitt:
> is there any way to test that? ;-)

I tried, but that required making a type that lived under a vendor path, which for a repo that doesn't vendor ... was non-trivial

apelisse:
Thanks!
/unhold

Issue #225, opened 2021-02-02
hardcoded `git checkout` path at $GOPATH/src/k8s.io, is required, to pass `go test`, using github-actions


longwuyuan:
Hi,

In github-actions ci, _"go test"_ fails, unless a _"mkdir $GOPATH/src/k8s.io"_ is done  _"$GOPATH/src"_ and _"git clone"_ is executed under _$GOPATH/src/k8s.io"_  .

But _"go build"_ works without the above specific path. So I am assuming that some piece of code under $PROJECT_HOME/test needs looking into.

Details here https://github.com/kubernetes/kube-openapi/pull/226

longwuyuan:
@nikhita @dims

longwuyuan:
https://github.com/kubernetes/kube-openapi/pull/226#issuecomment-772117989

longwuyuan:
A youtube video on Go tutorial says the first GOPATH gets used for dependencies but a second GOPATH works for custom directory tree of src. Need to test and read relevance to Go version ^1.14 .

apelisse:
There's a path variable in the checkout module that lets you chose where to check out, have you tried that?

apelisse:
@liggitt Clearly that's the problem :-)

longwuyuan:
I thought i tried the path option and i thought even the path  option forced checkout path as a subdir under "/home/runner/work" but the tests only look under $GOPATH /home/runner/go even though GO111MODULES=on was used.

Checking again now

apelisse:
You should be able to use GOPATH to set-up your path accordingly?

longwuyuan:
I just forked again and hoping to use the checkout action and GOPATH as you advised. WIll update asap.

longwuyuan:
> You should be able to use GOPATH to set-up your path accordingly?

- Using github-action module checkout ;
  - build passes, test fails with reason 

     ```
      STEP: building openapi-gen
    Failure [0.001 seconds]
    [BeforeSuite] BeforeSuite 
    /home/runner/work/kube-openapi/kube-openapi/test/integration/integration_suite_test.go:68
    
      Expected error:
          <*errors.errorString | 0xc000091620>: {
              s: "$GOPATH not provided when building ../../cmd/openapi-gen/openapi-gen.go",
          }
          $GOPATH not provided when building ../../cmd/openapi-gen/openapi-gen.go
      not to have occurred
    
[...]

PR #226, opened 2021-02-02
Added github-action CI

longwuyuan:
Question: What does this PR do ?
Answer: Adds a CI pipeline to run on Github-actions, as per discussion with @dims  https://kubernetes.slack.com/archives/C0EG7JC6T/p1611674631037300

Question: Any problems ?
Answer: Yes. The integration test fails unless the src of the project is put under $GOPATH/src/k8s.io, specifically. There is a go.mod so go test should work under any path. The concern described below can be seen here https://github.com/longwuyuan/kubernetes-kube-openapi/runs/1815796018?check_suite_focus=true
- In the travis ci file called .travis.yml, on line-number 9, in the root of the project, there is a special config for import paths that looks like this
```
go_import_path: k8s.io/kube-openapi
``` 
- Travis has an option to configure something called "go_import_path" https://docs.travis-ci.com/user/languages/go/#go-import-path
- There is no well documented equivalent of this "go_import_path", in github-actions
- As a result, the package name or path "k8s-io/kube-api" https://github.com/kubernetes/kube-openapi/blob/8566a335510fbdd49b876cbf1dba523bf83d077f/.travis.yml#L9  is not being explicitly passed on to "go test ..." in the github-actions CI file
  - Integration test fails if the CI script is very staightforward simple like this ;
    ```
     name: go mod
        run: go mod tidy && git diff --exit-code

      - name: go build
        run: go build ./cmd/... ./pkg/...

      - name: go test
        run: go test ./cmd/... ./pkg/... ./test/...

    ```
  - There are 3 initial STEPS to the integration test ;
    - building openapi-gen
    - processing go idl with openapi-gen
    - writing swagger
       Seen in the screenshot below ;
    image

  - The first error is _"$GOPATH not provided to build"_,  
     Looks like this screenshot below ;
  image
 
    - If  a change is made like the "gexec.Build" to "gexec.BuildIn" like thus ;
      Line 80 of file <project-root>/test/integration/integration_test_suite.go, changed 
      from 
       ```
      gexec.Build( "../../cmd/openapi-gen/openapi-gen.go
      ```
      to
      ```
      gexec.BuildIn("/home/runner/go", "../../cmd/openapi-gen/openapi-gen.go
      ```
      as per this documentation https://onsi.github.io/gomega/#gexec-testing-external-processes 

      -  The test framework gomega/gingko gets the GOPATH, explicitly hardcoded, so gets past the error and "building the openapi-gen" STEP works.

      - This is odd & unexpected because that gexec.Build is expected to pick up the GOPATH from the environment as per documentation.

  - Now comes the cause of this concern. It is the second STEP from the 3 mentioned above. The second STEP is failing, with  reason for failure as test fails to find a file called boilerplate.go.txt. 
    - This screenshot below shows this failed to find file error ;
      image

    - Attempting a "find" command just before the test, to see where that file is. 
       Seen in this screenshot below ;
        image

    - But the test always looks for it in this specific location  _k8s.io/kube-openapi/boilerplate/boilerplate.go.txt_ and fails . I have tried to set environment specifics like "working_directory", "path" and even changed the GOPATH to the github-actions workspace folder.  But the test always looks for  _k8s.io/kube-openapi/boilerplate/boilerplate.go.txt_ and fails.
    -  Maybe because of line 42 in ./cmd/openapi-gen/args/args.go
       ```
       genericArgs.GoHeaderFilePath = filepath.Join(args.DefaultSourceTree(), "k8s.io/kube-openapi/boil    erplate/boilerplate.go.txt")
       ```
    - As mentioned earlier, the old Travis CI file has a special config for the import path _"k8s.io/kube-openapi"_ . That travis ci configuration seems somewhat related to setting a remote-url import path. That Travis CI config is documented here https://docs.travis-ci.com/user/languages/go/#go-import-path  and that page links to the likely relevance to this issue, here  https://golang.org/cmd/go/#hdr-Remote_import_paths

- I am not a developer, so apologies for any misinformation here.

@dims @nikhita

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -1,0 +1,36 @@
+name: ci-pipeline
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '^1.13'
+
+    - name: checkout
+      run: |
+        cd /home/runner/go
+        mkdir src/k8s.io -p
+        cd /home/runner/go/src/k8s.io
+        git clone https://github.com/kubernetes/kube-openapi.git
+
+    - name: build
+      run: |
+        cd /home/runner/go/src/k8s.io/kube-openapi
+        go mod tidy && git diff --exit-code
+        go build ./cmd/... ./pkg/...
+
+    - name: test
+      run: |
+        cd /home/runner/go/src/k8s.io/kube-openapi
+        GOPATH=/home/runner/go go test ./cmd/... ./pkg/... ./test/...


.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+        mkdir src/k8s.io -p
+        cd /home/runner/go/src/k8s.io
+        echo "Current Working Directory Below"
+        pwd
+        git clone https://github.com/kubernetes/kube-openapi.git
+
+    - name: Build
+      run: |
+        cd /home/runner/go/src/k8s.io/kube-openapi
+        go mod tidy
```
apelisse, line 37: Do we have the ability to run a diff to see if mod tidy changed anything, like in https://github.com/kubernetes/kube-openapi/blob/master/.travis.yml#L12

apelisse:
/lgtm

Thanks! I'll let someone else approve.

longwuyuan:
@apelisse  I was wondering if you or anyone else has any comment about absolutely needing to manually create $GOPATH/src/k8s.io and checkout/build/test only at $GOPATH/src/k8s.io/kube-openapi.

The docs say that with a go.mod in the project's root directory, the build & test should work under any path. But that is not the case. Integration test fails if I checkout to anyplace other than $GOPATH/src/k8s.io. I have described this in the PR description. Thank you very much.

.github/workflows/go.yml
longwuyuan: added git diff just like the travis ci and rebased. thnx

apelisse:
Yeah, we SHOULD fix that, it's because we depend on a file outside the repository, and some assumptions are made about the structure and location of this repository. This is WRONG, we should fix it, but that seems unrelated to it. If you want, I can help you look at it and fix it (in this PR or another).

longwuyuan:
@apelisse thanks tons. I have been trying for 3 days without a clue. To avoid clutter of commits, I am trying to use https://github.com/longwuyuan/kubernetes-kube-openapi . Kindly advise if we can use that and get it working. I added you there

apelisse:
OK, It looks a little more complicated than I hoped. Let's try to fix it later but it's not obvious how.

longwuyuan:
ok that settles it for now then. Closest I came to see is 

```
/cmd/openapi-gen/args/args.go: genericArgs.GoHeaderFilePath = filepath.Join(args.DefaultSourceTree(), "k8s.io/kube-openapi/boilerplate/boilerplate.go.txt")
```
Some of the integration test framework chooses the github-action dedault workspace at _/home/runner/work/kube-openapi-kube-openapi_ as the "env" provided "DefaultSourceTree"  equivalent.  But in this case its choosing $GOPATH/src/..... 

I am not a developer so apologies for rant.

So, settled for now, that  we just live with the manual created directory tree under $GOPATH/src .

longwuyuan:
@sttts please also advise if we need
```
go-versions: [1.13.x, 1.14.x, 1.15.x]
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.15
+
+    - name: Checkout SRC
+      run: |
+        go env
+        pwd
+        env
```
apelisse, line 26: Do we need these?

.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+        cd /home/runner/go
+        mkdir src/k8s.io -p
+        cd /home/runner/go/src/k8s.io
+        echo "Current Working Directory Below"
+        pwd
```
apelisse, line 31: Maybe remove this?

apelisse (commented):
Looks good I think, always hard to test these things.

dims:
@longwuyuan looks like some debug stuff is left like @apelisse mentioned, let's please clean that up and get this ready to merge.

.github/workflows/go.yml
longwuyuan: removed

.github/workflows/go.yml
longwuyuan: removed

longwuyuan:
@dims @apelisse , I cleaned up as suggested. Some long-term topics I have listed in https://github.com/kubernetes/kube-openapi/issues/225

apelisse:
Thanks,
/lgtm
/approve

.github/workflows/ci_pipeline.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: '^1.13'
+
+    - name: checkout
+      run: |
+        cd /home/runner/go
+        mkdir src/k8s.io -p
+        cd /home/runner/go/src/k8s.io
+        git clone https://github.com/kubernetes/kube-openapi.git
```
apelisse, line 25: This is checking out the master branch, and not the pull-request. So it's not testing that the pull-request works, but that master works. Unfortunately, once master is broken, we can't merge pull-requests anymore :-(

Can we use this instead: https://github.com/actions/checkout ?

longwuyuan: @apelisse @dims
I am working on using  https://github.com/actions/checkout .
Reason for not using was #225

PR #230, opened 2021-03-04
Remove incorrect zero value check

liggitt:
This check never returned true prior to go1.16, because using DeepEqual on reflect.Value
return values checks their internal fields, and reflect.Zero included a unique unsafe pointer.

In go1.16, reflect.Zero started returning shared zero values, and this check started returning true.

However, requiring nullable on zero values is not correct. A zero value string "" should be valid even
if the schema does not allow the field to be null.

xref https://github.com/go-openapi/validate/pull/138 and https://github.com/golang/go/issues/43993#issuecomment-790948302

/cc @sttts @apelisse


liggitt:
cc @BenTheElder @deads2k

BenTheElder:
/lgtm
CI is unhappy about go.sum though

deads2k:
reasoning looks sound to me.

PR #235, opened 2021-05-24
feat(embedded_struct): replace struct type by single embedded field type

fenngwd:
If there is only single embedded struct field, use it to replace parent
type when generating default values.


fenngwd:
Fixes kubernetes/kubernetes#101716

apelisse:
Thanks, I think that's fine.
/lgtm
/approve

PR #237, opened 2021-05-26
Use Deduced schema for Arbitrary types

apelisse:
Arbitrary types shouldn't be treated as a one-level map of atomic
fields, but as a deduced type that can be made of anything.

### DIFF
--- a/pkg/schemaconv/smd.go
+++ b/pkg/schemaconv/smd.go
@@ -449,10 +449,7 @@ func (c *convert) VisitPrimitive(p *proto.Primitive) {
 }
 
 func (c *convert) VisitArbitrary(a *proto.Arbitrary) {
-	*c.top() = untypedDef.Atom
-	if c.preserveUnknownFields {
-		*c.top() = deducedDef.Atom
-	}
+	*c.top() = deducedDef.Atom
 }
 
 func (c *convert) VisitReference(proto.Reference) {
--- a/pkg/schemaconv/testdata/new-schema.yaml
+++ b/pkg/schemaconv/testdata/new-schema.yaml
@@ -9410,8 +9410,8 @@ types:
     elementRelationship: atomic
   map:
     elementType:
-      namedType: __untyped_atomic_
-    elementRelationship: atomic
+      namedType: __untyped_deduced_
+    elementRelationship: separable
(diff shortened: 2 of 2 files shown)


apelisse:
If we merge this pull-request, it means that objects with "arbitrary" types will now track all the map layers rather than just the first layer.

seans3 (commented):
This field pruning documentation describes interaction with `preserveUnknownFields` (and how they differ between `v1beta` and `v1`). Is this documentation still accurate after this change?

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

lavalamp:
Does the deduced type actually work well? I think it'd be nice if it noticed `name` or `id` fields in a list and marked the list as being associative.

apelisse:
> Does the deduced type actually work well? I think it'd be nice if it noticed name or id fields in a list and marked the list as being associative.

I don't know if now is the best time to do that. 

We're going GA soon, and CRDs v1 is going to become the norm, forcing structural schemas (and non-deduced). It doesn't mean people can't just avoid the problem by creating an empty allow all schemas though.

Also, the problem with arbitrary is that we would have to change the rules constantly, and I don't think the tracking would really like it. What if some entries don't have name but some do? What if we have multiple entries that have the same name? Would we have to go back-and-forth between atomic/associative-by-name?

So yeah, I think it's working reasonably well enough :-)

lavalamp:
I guess, let me put it this way: will be we able to make improvements or at least fix bugs in the deduced schema in the future? Or will someone *cough*Jordan*cough* prevent that for the sake of backwards compatibility?

apelisse:
> will be we able to make improvements or at least fix bugs in the deduced schema in the future?

I guess we can fix SOME bugs, I don't know about improvements :-)

apelisse:
cc @janetkuo

lavalamp:
Thinking more, I guess if it does the wrong thing users can self-rescue by upgrading their CRD, which they have to anyway because v1beta1 is going away.

/lgtm

apelisse:
> This field pruning documentation describes interaction with preserveUnknownFields (and how they differ between v1beta and v1). Is this documentation still accurate after this change?

Thanks @seans, yeah I think it's still accurate!

PR #239, opened 2021-06-25
Add preserve unknown fields test case

jpbetz:
Adding this primarily to document the mapping between openAPI and SMD schema for x-kubernetes-preserve-unknown-fields


apelisse:
Thanks, I *think* that's the behavior I would have expected, thanks!

jpbetz:
I just realized  this is still open. Anyone willing to approve?

PR #242, opened 2021-08-03
Enum type generator

jiahuif:
This PR is the implementation of https://github.com/kubernetes/enhancements/issues/2887 in k/kube-openapi
This PR implements generation of enum type definition in the schema.

- Detect enum types by `+enum`
- Adds all possible values to [Enum](https://github.com/kubernetes/kube-openapi/blob/95288971da7e1e3d736b57c826a238a14330e0a8/pkg/validation/spec/schema.go#L180) field
- Add a description of all enum values, as a Markdown list, to the description.

An example:
```go
package foo

// EnumType is the enumType
// +enum
type EnumType string

// EnumA is a
const EnumA EnumType = "a"
// EnumB is b
const EnumB EnumType = "b"

// Blah is a test.
// +k8s:openapi-gen=true
// +k8s:openapi-gen=x-kubernetes-type-tag:type_test
type Blah struct {
  // Value is the value
	Value EnumType
}
```

generated:
```go
common.OpenAPIDefinition{
	Schema: spec.Schema{
			Description: "Blah is a test.",
			Type:        []string{"object"},
				"Value": {
						Description: "Value is the value EnumType:\n - `a`: is a\n - `b`: is b",
						Default:     "",
						Type:        []string{"string"},
						Format:      "",
						Enum:        []interface{}{"a", "b"}},
				},
			},
			Required: []string{"Value"},
		},
			Extensions: spec.Extensions{
				"x-kubernetes-type-tag": "type_test",
			},
		},
	},
}

```


jiahuif:
@apelisse Is it okay to bump `k8s.io/gengo` version? The current version is very old and causing issues with my work.

apelisse:
Sure!

jiahuif:
Also
/cc @jpbetz

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
jiahuif, line 1: Note: this is added through `go tidy`

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+		v = append(v, fmt.Sprintf("%q", value.Value))
+	}
+	sort.Strings(v)
+	return
+}
+
+// FormatEnumDescription returns a description of the enum type in the following
+//  format:
+// FooType:
+//  - `value1`: description 1
```
jpbetz, line 72: very minor nit: The below code writes the bullet list with `*`, not `-`

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+	sort.Strings(lines)
+	return append([]string{header}, lines...)
+}
+
+func parseEnums(c *generator.Context) (enumTypes enumMap) {
+	stringType := c.Universe.Type(types.Name{Name: "string"})
+
+	enumTypes = make(enumMap)
+	for _, p := range c.Universe {
+
```
jpbetz, line 101: nit: empty line needed here?

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+	"k8s.io/gengo/generator"
+	"k8s.io/gengo/types"
+)
+
+const tagEnumType = "enum"
+
+type enumValue struct {
+	Name    string
+	Value   string
+	Comment string
```
jpbetz, line 34: An alternative to keeping track of the comments of each enum value would be to keep track of the comment for the entire enum? E.g.:

`go
type enumType struct {
  Values []*enumValue
  Description string
}

type enumValue struct {
  Name string
  Value string
}
`

The advantage being that the description only needs to be constructed once.

pkg/generators/openapi_test.go
```diff
@@ -1570,3 +1570,60 @@ map[string]interface{}{
 
 `, funcBuffer.String())
 }
+
+func TestEnum(t *testing.T) {
+	callErr, funcErr, assert, _, funcBuffer := testOpenAPITypeWriter(t, `
+package foo
+
+// EnumType is the enumType
```
jpbetz, line 1578: I was expecting this enum type comment to be included along with the enum value comments.

jpbetz (changes requested):

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// IsKnownEnumType checks if a given type is a known enum type
+func (ec *enumContext) IsKnownEnumType(t *types.Type) bool {
+	_, ok := ec.enumTypes[t.Name]
+	return ok
+}
+
+// FormatEnumValues returns all possible values of the enum type
+//  the results are sorted and quoted as Go literals.
```
apelisse, line 55: Nit: A lot of your comments have a weird structure/indentation/spaces inserted.

pkg/generators/enum.go
apelisse: By the way, I think I like dash more than star, but I think that's completely subjective so mostly ignore me ;-)

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+//  - `value1`: description 1
+//  - `value2`: description 2
+func (ec *enumContext) FormatEnumDescription(t *types.Type) []string {
+	values, ok := ec.enumTypes[t.Name]
+	if !ok {
+		return nil
+	}
+
+	var lines []string
+	header := t.Name.Name + ":"
```
apelisse, line 81: I don't know about the Go type name, since that's an implementation detail

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, value := range values {
+		comment := strings.TrimSpace(value.Comment)
+
+		// The comment should starts with the type name, trim it first.
+		comment = strings.TrimPrefix(comment, value.Name)
+		comment = strings.TrimSpace(comment)
+		// The comment may be multiline, cascade all consecutive whitespaces.
+		comment = whitespaceRegex.ReplaceAllString(comment, " ")
```
apelisse, line 89: Maybe isolate all of this in a separate function, I found it a little confusing as I was reading the code.

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+		lines = append(lines, fmt.Sprintf(" * `%s`: %s", value.Value, comment))
+	}
+	sort.Strings(lines)
+	return append([]string{header}, lines...)
+}
+
+func parseEnums(c *generator.Context) (enumTypes enumMap) {
```
apelisse, line 96: I'm not generally a big fan of named returned values, but that's probably just my preference.

apelisse (commented):
Looks good overall. I can see now why you're asking about generating v3 OpenAPI, let's talk about that.

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -276,6 +278,34 @@ func schema_test_integration_testdata_dummytype_Waldo(ref common.ReferenceCallba
...
+			SchemaProps: spec.SchemaProps{
+				Type: []string{"object"},
+				Properties: map[string]spec.Schema{
+					"content": {
+						SchemaProps: spec.SchemaProps{
+							Description: "\n\nFruitType:\n * `apple`: is the Apple\n * `banana`: is the Banana\n * `onigiri`: is the Rice ball that does not seem to belong to a fruits basket but has a long comment that spans multiple lines",
+							Default:     "",
+							Type:        []string{"string"},
+							Format:      "",
+							Enum:        []interface{}{"apple", "banana", "onigiri"}},
```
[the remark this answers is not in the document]
apelisse: Ah, if they exist in v2 (which they seem to do: https://swagger.io/docs/specification/2-0/enums/), then we should totally add them right away and distinguish that effort from v3? Or am I missing something?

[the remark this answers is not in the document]
jiahuif: As long as nobody is relying on the enum fields being empty.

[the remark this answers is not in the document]
jiahuif: Reusable enums is another improvement of v3 over v2. If we are to use this feature, the generation part would need checking the version anyways.

Ref:  https://swagger.io/docs/specification/data-models/enums/

pkg/generators/enum.go
jiahuif: Misconfiguration of the editor (with my own workstation). Fixed.

pkg/generators/enum.go
jiahuif: Good point. I am rethinking how this header should be structured.

An example from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#containerport-v1-core
Field | Description
-- | --
protocol string Protocol for port. Must be UDP, TCP, or SCTP. Defaults to "TCP".

test/integration/pkg/generated/openapi_generated.go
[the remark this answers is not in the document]
apelisse: > As long as nobody is relying on the enum fields being empty.

Yeah, I'm not concerned by that.

> Reusable enums is another improvement of v3 over v2. If we are to use this feature, the generation part would need checking the version anyways.

Yeah, we could consider these, but we don't have too at that time. I think we're fine just duplicating things wherever needed for now.

pkg/generators/enum.go
jiahuif: was a removed comment, I will polish it a bit.

pkg/generators/enum.go
jiahuif: The comment of entire enum will be used for at most once. I think a getter would be better

pkg/generators/enum.go
jiahuif: I replace the header with a simple `Possible enum values:`

pkg/generators/enum.go
jiahuif: There are cases where multiple return values can be more easily identified with named returned values. I left these cases but removed the rest.

pkg/generators/openapi_test.go
jiahuif: The comment on the enum type, like the name of the enum type, is usually kind of implementation detail. For example.

```go
// Protocol defines network protocols supported for things like container ports.
type Protocol string
```

later
```go
// ServicePort represents the port on which the service is exposed
type ServicePort struct {
	// Optional if only one ServicePort is defined on this service: The
	// name of this port within the service.  This must be a DNS_LABEL.
	// All ports within a ServiceSpec must have unique names.  This maps to
	// the 'Name' field in EndpointPort objects.
	Name string

	// The IP protocol for this port.  Supports "TCP", "UDP", and "SCTP".
	Protocol Protocol
```
The comment on the field should be good enough.

pkg/generators/enum.go
```diff
@@ -0,0 +1,152 @@
...
+		// use "%q" format to generate a Go literal of the string const value
+		values = append(values, fmt.Sprintf("%q", value.Value))
+	}
+	sort.Strings(values)
+	return values
+}
+
+// DescriptionLines returns a description of the enum in this format:
+//
+// FooType:
```
apelisse, line 76: extremely minor nit, but that comment is out of date now

pkg/generators/enum.go
```diff
@@ -0,0 +1,152 @@
...
+//  -	`value1`: description 1
+//  - `value2`: description 2
+func (et *enumType) DescriptionLines() []string {
+	var lines []string
+	for _, value := range et.Values {
+		lines = append(lines, value.Description())
```
apelisse, line 82: Thanks, this function is much easier to read!

apelisse (commented):
Beside a minor comment nit,  I don't have anything against that. We will need an approved KEP to merge it I think.

pkg/generators/openapi_test.go
jpbetz: I agree that this is often not a particularly valuable comment.

But if a developer does add some really important information in this and we choose to omit it, and then as a result the API is misused, I don't think I could defend our decision to omit the the comment. I think we responsible for including everything author included in the godoc of the type in the openapi documentation.

jiahuif:
> Beside a minor comment nit, I don't have anything left to say. We will need an approved KEP to merge it I think.

KEP is there. Let's restart this.

pkg/generators/enum.go
jiahuif: Good catch, thanks!

test/integration/pkg/generated/openapi_generated.go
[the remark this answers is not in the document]
jiahuif: This discussion has been done in the KEP. Resolving here.

pkg/generators/enum.go
```diff
@@ -0,0 +1,139 @@
...
+	Value   string
+	Comment string
+}
+
+// enumMap is a map of enum type name to all its values
+type enumMap map[types.Name][]*enumValue
+
+type enumContext struct {
+	enumTypes enumMap
+}
```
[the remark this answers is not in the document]
jiahuif: actually, reworked this part.

pkg/generators/enum.go
jiahuif: I did a quick count and find out `*` is rarely used in our code base. So I switched to dash.

Jefftree:
@jiahuif it might be useful to have an additional PR to show how the k/k native OpenAPI changes as a result of this.

apelisse:
> @jiahuif it might be useful to have an additional PR to show how the k/k native OpenAPI changes as a result of this.

Totally, I'd love to see a WIP pull-request in kubernetes that integrates this change locally and adds the enum marker in a few places to see what it does.

jiahuif:
@Jefftree @apelisse PoC sent as https://github.com/kubernetes/kubernetes/pull/105018

test/integration/testdata/enumtype/enum.go
```diff
@@ -0,0 +1,25 @@
...
+
+// FruitBanana is the Banana
+const FruitBanana FruitType = "banana"
+
+// FruitRiceBall is the Rice ball that does not seem to belong to
+// a fruits basket but has a long comment that is so long that it spans
+// multiple lines
+const FruitRiceBall FruitType = "onigiri"
+
+// FruitsBasket is the type that contains the emun type.
```
Jefftree, line 18: s/emun/enum

Jefftree (commented):
Thanks, this looks great! One nit, otherwise lgtm

test/integration/testdata/enumtype/enum.go
jiahuif: fixed. Thanks!

PR #243, opened 2021-08-16
Make ipv4/ipv6/cidr format validation backwards compatible

liggitt:
xref https://github.com/kubernetes/kubernetes/issues/100895

keeps openapi format validation for ipv4 and cidr backwards compatible even when built with go1.17+

- [x] Dropped unused travis CI config, updated github CI to test go 1.13, latest released go, and go 1.17rc2 (https://github.com/kubernetes/kube-openapi/pull/243/checks?sha=5663be5eac51e9ee4d83bff54377e3f73c5d498c)
- [x] Added test cases for ipv4 and cidr with leading 0's (test passed on go1.13 and go1.16.x and failed on go1.17rc2 as expected - https://github.com/kubernetes/kube-openapi/pull/243/checks?sha=8928e5317693d9d805f01223b7fa3d1515b0f7a4)
- [x] Switched to use k8s.io/utils compatible IP/CIDR parsing, tests passed on all go versions (https://github.com/kubernetes/kube-openapi/pull/243/checks?sha=6a51b86f9e27ab4f713c38c9a88b902d9fe2517b)


pkg/validation/strfmt/default.go
```diff
@@ -154,13 +156,13 @@ func init() {
 	Default.Add("hostname", &hn, IsHostname)
 
 	ip4 := IPv4("")
-	Default.Add("ipv4", &ip4, govalidator.IsIPv4)
+	Default.Add("ipv4", &ip4, isIPv4)
```
aojea, line 159: You have one replacement here `IsIPv4String`
https://github.com/kubernetes/utils/blob/efc7438f0176a4a29c62bac8989d8dd183054fa5/net/net.go#L145

pkg/validation/strfmt/default.go
```diff
@@ -154,13 +156,13 @@ func init() {
...
 
 	ip6 := IPv6("")
 	Default.Add("ipv6", &ip6, govalidator.IsIPv6)
```
aojea, line 162: IsIPv6 https://github.com/kubernetes/utils/blob/efc7438f0176a4a29c62bac8989d8dd183054fa5/net/net.go#L122

aojea: nevermind, the leading zeros is for IPv4 only

pkg/validation/strfmt/default.go
liggitt: mm... the implementation isn't quite the same... I'm not confident it would handle `::1.2.3.4` addresses the same way

aojea: then better to be on the safe side, and not change anything, I thought it was simpler but is not

PR #244, opened 2021-08-20
CI go version mapping upgrade

jiahuif:
A previous PR introduce builds on go 1.17.0-rc, this PR update go to its latest stable version.

Also this PR removed go 1.13 because 
- go 1.13 is long out of maintenance
- no supported Kubernetes versions still use go 1.13. oldest go version of all supported Kubernetes release branch is 1.16
- integration test fails randomly when running with 1.13

As a result, CI runs with the following version:
- go 1.16 stable
- go 1.x stable (currently 1.17.x)
- go 1.x unstable (current 1.17.x because there is no unstable go version)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.13', '1.x' ]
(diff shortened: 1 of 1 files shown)


apelisse:
/lgtm
/approve

Thanks for looking at it!

Issue #245, opened 2021-09-14
current version of json-iterator incompatible with go1.18


liggitt:
json-iterator just cut a new release to address incompatibilities with the current go development version (will be go1.18)

xref:
* https://github.com/kubernetes/kubernetes/pull/104949
* https://github.com/kubernetes/kubernetes/issues/104947
* https://github.com/json-iterator/go/releases/tag/v1.1.12
* https://github.com/modern-go/reflect2

PR #246, opened 2021-09-14
Switch to stdlib json encoding

liggitt:
Fixes https://github.com/kubernetes/kube-openapi/issues/245, drops memory-unsafe json encoder


mkumatag:
@liggitt are we going to take the same direction for the k/k repo as well? https://github.com/kubernetes/kubernetes/issues/104947

liggitt:
I would like to (e.g. https://github.com/kubernetes/kubernetes/pull/105030), but I'm not sure of the timeframe for that

PR #247, opened 2021-09-16
Support for OpenAPI v3

Jefftree:
Most of work around the spec was from https://github.com/p0lyn0mial/kube-openapi/tree/spec3/pkg/spec3.

This PR provides the structs for OpenAPI v3, the builder3 package for building OpenAPI v3, and handler3 for serving OpenAPI v3 separated by group/version.

See https://github.com/kubernetes/kubernetes/pull/105945 for k/k integration


pkg/spec3/security_scheme_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"encoding/json"
+	"testing"
+
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"github.com/google/go-cmp/cmp"
+
+	"k8s.io/kube-openapi/pkg/spec3"
+)
+
+func TestSecuritySchemaJSONSerialization(t *testing.T) {
```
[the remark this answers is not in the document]
Jefftree: Ended up adding tests for all the serialization files above. As for the canonical round-trip test, we can use a subset of the k/k openapi spec for a specific group. I'm still not sure whether that'd be better to include in this repo or in k/k.

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+const (
+	OpenAPIVersion = "3.0"
+)
+
+type openAPI struct {
+	config      *common.Config
+	spec        *spec3.OpenAPI
+	definitions map[string]common.OpenAPIDefinition
```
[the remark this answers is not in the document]
Jefftree: There's a GetDefinitions function interface that we still rely on (https://github.com/kubernetes/kube-openapi/blob/master/pkg/common/common.go#L87), and I thought it might be good to keep consistency. This struct technically holds "OpenAPIDefinition" objects that have not been converted to their respective components yet so I think the naming is still accurate?

p0lyn0mial:
Would it make sense to generate a v3 spec for k/k repo and use a validator to validate the generated spec?
I think last time I did https://github.com/p0lyn0mial/kube-oasv3-poc

Jefftree:
@p0lyn0mial I've been using the [kin-openapi](https://github.com/getkin/kin-openapi) repo to validate the OpenAPI v3 in the [integration tests](https://github.com/kubernetes/kube-openapi/blob/226a557fcf5fde9b995a2e1aa69109f435d0f836/test/integration/builder3/main.go#L68-L79) in kube-openapi.

The PR for k/k https://github.com/kubernetes/kubernetes/pull/105945 also generates the static `swagger.json` equivalent for all v3 group/versions and I've manually validated a couple of the generated specs. Is there a specific validation tool you'd like to see integrated?

Jefftree:
@p0lyn0mial I've ran your suggested script on the [set of files generated in the k/k PR](https://github.com/kubernetes/kubernetes/pull/105945/commits/5bad38820911f0f2521b84c24eae06e223b56883) per group.

`for f in api/openapi-spec/v3/*; do pajv validate -s ~/tmp/schema.json -d "$f"; done`

and all the group/version paths passed the validation.

p0lyn0mial:
> @p0lyn0mial I've been using the kin-openapi repo to validate the OpenAPI v3 in the integration tests in kube-openapi.

sgtm, we should validate a spec generated directly from this repo


> The PR for k/k kubernetes/kubernetes#105945 also generates the static swagger.json equivalent for all v3 group/versions and I've manually validated a couple of the generated specs. Is there a specific validation tool you'd like to see integrated?

nothing specific, I just wanted to make sure the spec generated from k/k repo is a valid OAS v3 spec.

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,356 @@
...
+	response := &spec3.Response{
+		ResponseProps: spec3.ResponseProps{
+			Description: description,
+			Content:     make(map[string]*spec3.MediaType),
+		},
+	}
+
+	s, err := o.toSchema(util.GetCanonicalTypeName(model))
+	if err != nil {
+		return response, err
```
apelisse, line 59: minor, should this `return nil, err`?

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,356 @@
...
+				},
+			},
+		},
+	}
+
+	var err error
+
+	// Build responses
+	for _, resp := range route.ResponseErrors {
+		ret.Responses.StatusCodeResponses[resp.Code], err = o.buildResponse(resp.Model, resp.Message)
```
apelisse, line 86: not failing on errors?

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,356 @@
...
+
+func (o *openAPI) buildParameter(restParam restful.ParameterData, bodySample interface{}) (ret *spec3.Parameter, err error) {
+	ret = &spec3.Parameter{
+		ParameterProps: spec3.ParameterProps{
+			Name:        restParam.Name,
+			Description: restParam.Description,
+			Required:    restParam.Required,
+		},
+	}
+	switch restParam.Kind {
```
apelisse, line 265: Just curious as I'm comparing to v2, I can see that you dropped `BodyParameterKind`, do you know what happened to it? :-)

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,356 @@
...
+			}
+			for k, v := range extensions {
+				schema.Extensions[k] = v
+			}
+		}
+		// if v, ok := item.Schema.Extensions[common.ExtensionV2Schema]; ok {
+		// 	if v2Schema, isOpenAPISchema := v.(spec.Schema); isOpenAPISchema {
+		// 		schema = v2Schema
+		// 	}
+		// }
```
apelisse, line 316: Yeah, I'm not sure what this was to be honest, do we need to keep it commented, what is this doing here?

pkg/handler3/handler.go
```diff
@@ -0,0 +1,248 @@
...
+	"k8s.io/kube-openapi/pkg/common"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+const (
+	jsonExt = ".json"
+
+	mimeJson = "application/json"
+	// TODO(mehdy): change @68f4ded to a version tag when gnostic add version tags.
+	mimePb   = "application/com.github.googleapis.gnostic.OpenAPIv2@68f4ded+protobuf"
```
apelisse, line 45: is this one also right?

pkg/handler3/handler.go
```diff
@@ -0,0 +1,248 @@
+/*
```
apelisse, line 1: I think that's fine for now since this is alpha and we're not going to cache the OpenAPI unless the feature is enabled (right?!), but this absolutely needs to use @DangerOnTheRanger's cache fix for beta! thanks.

pkg/builder3/openapi.go
Jefftree: Thanks for the catch! Added. This was quite an important parameter

```
       },
       "post": {
         "description": "create a ConfigMap",
+        "parameters": [
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/components/schemas/io.k8s.api.core.v1.ConfigMap"
+            }
+          },
```

pkg/handler3/handler.go
Jefftree: Definitely agree. I plan to gather some performance numbers for beta with the cache as well

pkg/builder3/openapi.go
```diff
@@ -0,0 +1,340 @@
...
+			// used at the end of the path and it is not recognized by OpenAPI.
+			if strings.HasSuffix(path, ":*}") {
+				path = path[:len(path)-3] + "}"
+			}
+			if pathsToIgnore.HasPrefix(path) {
+				continue
+			}
+
+			// Aggregating common parameters make API spec (and generated clients) simpler
+			inPathCommonParamsMap, err := o.findCommonParameters(routes)
```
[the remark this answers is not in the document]
Jefftree: oops sorry I missed this comment earlier. Yes, it finds all parameters in the route's different operations and collects them together (and their spec) to publish.

test/integration/testdata/golden.v2.json
```diff
@@ -7,6 +7,9 @@
   "paths": {
    "/test/custom/bac": {
     "get": {
+     "produces": [
+      "application/json"
```
Jefftree, line 11: The old test was incomplete and swagger recommends having this produces field the response includes a schema. Since we are sharing the input structs for v2 and v3, I added this field for v2 tests as well. See [link](https://swagger.io/specification/v2/#responseObject)

"A definition of the response structure. It can be a primitive, an array or an object. If this field does not exist, it means no content is returned as part of the response. As an extension to the Schema Object, its root type value may also be "file". This SHOULD be accompanied by a relevant produces mime-type."

apelisse:
Looks good to me, feel free to tag @sttts, thanks!

PR #251, opened 2021-09-28
Lazy marshaling for OpenAPI v2 spec

DangerOnTheRanger:
This PR makes the marshaling process lazy for OpenAPI v2.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -33,6 +33,7 @@ import (
 	openapi_v2 "github.com/googleapis/gnostic/openapiv2"
 	"github.com/munnerz/goautoneg"
 	"gopkg.in/yaml.v2"
+	klog "k8s.io/klog/v2"
 	"k8s.io/kube-openapi/pkg/builder"
 	"k8s.io/kube-openapi/pkg/common"
 	"k8s.io/kube-openapi/pkg/validation/spec"
@@ -55,13 +56,40 @@ type OpenAPIService struct {
 
 	lastModified time.Time
 
-	specBytes []byte
-	specPb    []byte
-	specPbGz  []byte
+	jsonCache  cache
+	protoCache cache
+}
 
-	specBytesETag string
-	specPbETag    string
-	specPbGzETag  string
+type cache struct {
+	BuildCache func() ([]byte, error)
+	once       sync.Once
+	bytes      []byte
+	etag       string
+	err        error
+}
+
+func (c *cache) Get() ([]byte, string, error) {
+	c.once.Do(func() {
+		bytes, err := c.BuildCache()
+		// if there is an error updating the cache, there can be situations where
+		// c.bytes contains a valid value (carried over from the previous update)
+		// but c.err is also not nil; the cache user is expected to check for this
+		c.err = err
+		if c.err == nil {
+			// don't override previous spec if we had an error
+			c.bytes = bytes
+			c.etag = computeETag(c.bytes)
+		}
+	})
+	return c.bytes, c.etag, c.err
+}
+
+func (c *cache) New(cacheBuilder func() ([]byte, error)) cache {
+	return cache{
+		bytes:      c.bytes,
+		etag:       c.etag,
+		BuildCache: cacheBuilder,
+	}
 }
 
 func init() {
@@ -71,6 +99,9 @@ func init() {
 }
 
 func computeETag(data []byte) string {
+	if data == nil {
+		return ""
+	}
 	return fmt.Sprintf("\"%X\"", sha512.Sum512(data))
 }
 
@@ -83,51 +114,40 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
 	return o, nil
 }
 
-func (o *OpenAPIService) getSwaggerBytes() ([]byte, string, time.Time) {
+func (o *OpenAPIService) getSwaggerBytes() ([]byte, string, time.Time, error) {
 	o.rwMutex.RLock()
 	defer o.rwMutex.RUnlock()
-	return o.specBytes, o.specBytesETag, o.lastModified
+	specBytes, etag, err := o.jsonCache.Get()
+	if err != nil {
+		return nil, "", time.Time{}, err
+	}
+	return specBytes, etag, o.lastModified, nil
 }
 
-func (o *OpenAPIService) getSwaggerPbBytes() ([]byte, string, time.Time) {
+func (o *OpenAPIService) getSwaggerPbBytes() ([]byte, string, time.Time, error) {
 	o.rwMutex.RLock()
 	defer o.rwMutex.RUnlock()
-	return o.specPb, o.specPbETag, o.lastModified
-}
-
-func (o *OpenAPIService) getSwaggerPbGzBytes() ([]byte, string, time.Time) {
-	o.rwMutex.RLock()
-	defer o.rwMutex.RUnlock()
-	return o.specPbGz, o.specPbGzETag, o.lastModified
+	specPb, etag, err := o.protoCache.Get()
+	if err != nil {
+		return nil, "", time.Time{}, err
+	}
+	return specPb, etag, o.lastModified, nil
 }
 
 func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
-	specBytes, err := json.Marshal(openapiSpec)
-	if err != nil {
-		return err
-	}
-	specPb, err := ToProtoBinary(specBytes)
-	if err != nil {
-		return err
-	}
-	specPbGz := toGzip(specPb)
-
-	specBytesETag := computeETag(specBytes)
-	specPbETag := computeETag(specPb)
-	specPbGzETag := computeETag(specPbGz)
-
-	lastModified := time.Now()
-
 	o.rwMutex.Lock()
 	defer o.rwMutex.Unlock()
-
-	o.specBytes = specBytes
-	o.specPb = specPb
-	o.specPbGz = specPbGz
-	o.specBytesETag = specBytesETag
-	o.specPbETag = specPbETag
-	o.specPbGzETag = specPbGzETag
-	o.lastModified = lastModified
+	o.jsonCache = o.jsonCache.New(func() ([]byte, error) {
+		return json.Marshal(openapiSpec)
+	})
+	o.protoCache = o.protoCache.New(func() ([]byte, error) {
+		json, _, err := o.jsonCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		return ToProtoBinary(json)
+	})
+	o.lastModified = time.Now()
 
 	return nil
 }
@@ -206,7 +226,7 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 	accepted := []struct {
 		Type           string
 		SubType        string
-		GetDataAndETag func() ([]byte, string, time.Time)
+		GetDataAndETag func() ([]byte, string, time.Time, error)
 	}{
 		{"application", "json", o.getSwaggerBytes},
 		{"application", "com.github.proto-openapi.spec.v2@v1.0+protobuf", o.getSwaggerPbBytes},
@@ -230,7 +250,15 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 					}
 
 					// serve the first matching media type in the sorted clause list
-					data, etag, lastModified := accepts.GetDataAndETag()
+					data, etag, lastModified, err := accepts.GetDataAndETag()
+					if err != nil {
+						klog.Errorf("Error in OpenAPI handler: %s", err)
+						// only return a 503 if we have no older cache data to serve
+						if data == nil {
+							w.WriteHeader(http.StatusServiceUnavailable)
+							return
+						}
+					}
 					w.Header().Set("Etag", etag)
 					// ServeContent will take care of caching using eTag.
 					http.ServeContent(w, r, servePath, lastModified, bytes.NewReader(data))
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	json "encoding/json"
+	"errors"
 	"io/ioutil"
 	"math"
 	"net/http"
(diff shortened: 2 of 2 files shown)


pkg/handler/handler.go
```diff
@@ -55,13 +55,24 @@ type OpenAPIService struct {
...
+	etag       string
+	err        error
+}
 
-	specBytesETag string
-	specPbETag    string
-	specPbGzETag  string
+func (c *cache) Get() ([]byte, string, error) {
+	c.once.Do(func() {
+		c.bytes, c.etag, c.err = c.BuildCache()
```
apelisse, line 73: Maybe something like this:
```suggestion
		c.bytes, c.err = c.BuildCache()
		if c.err != nil {
		    c.etag = computeETag(c.bytes)
		}
```

pkg/handler/handler.go
```diff
@@ -86,47 +97,72 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
...
+	o.rwMutex.Lock()
+	defer o.rwMutex.Unlock()
+	o.jsonCache = cache{
+		BuildCache: func() ([]byte, string, error) {
+			bytes, err := jsoniter.ConfigCompatibleWithStandardLibrary.Marshal(openapiSpec)
+			if err != nil {
+				return nil, "", err
+			}
+			etag := computeETag(bytes)
+			return bytes, etag, nil
```
apelisse, line 137:
```suggestion
			return jsoniter.ConfigCompatibleWithStandardLibrary.Marshal(openapiSpec)
```

pkg/handler/handler.go
```diff
@@ -86,47 +97,72 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
...
+			json, _, err := o.jsonCache.Get()
+			if err != nil {
+				return nil, "", err
+			}
+			pb, err := ToProtoBinary(json)
+			if err != nil {
+				return nil, "", err
+			}
+			etag := computeETag(pb)
+			return pb, etag, nil
```
apelisse, line 151:
```suggestion
			return ToProtoBinary(json)
```

pkg/handler/handler.go
```diff
@@ -86,47 +97,72 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
...
+	}
+	o.protoGzCache = cache{
+		BuildCache: func() ([]byte, string, error) {
+			proto, _, err := o.protoCache.Get()
+			if err != nil {
+				return nil, "", err
+			}
+			gz := toGzip(proto)
+			etag := computeETag(gz)
+			return gz, etag, nil
```
apelisse, line 162:
```suggestion
			return toGzip(proto), nil
```

apelisse (commented):
I sent a few suggestion that I think make the code simpler and less error-prone.

DangerOnTheRanger:
Computing the ETag once in `cache.Get` makes sense - fixed, thanks!

apelisse:
The code looks good. Can we see how this impacts Kubernetes though? How impact is the first request to `/openapi/v2`? Does this impact "start time"? Can we have a sort of analysis for that? I'd rather make sure that we know what we're causing to happen here. Thanks! cc @Jefftree

pkg/handler/handler.go
```diff
@@ -86,47 +97,72 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
 func (o *OpenAPIService) getSwaggerBytes() ([]byte, string, time.Time) {
 	o.rwMutex.RLock()
 	defer o.rwMutex.RUnlock()
-	return o.specBytes, o.specBytesETag, o.lastModified
+	specBytes, etag, err := o.jsonCache.Get()
+	if err != nil {
+		panic(err.Error())
```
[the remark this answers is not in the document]
apelisse: > What's the behavior on a panic?

I want to know that

Also I don't think you have to convert the error into a string with `panic`.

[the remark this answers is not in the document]
apelisse: Absolutely. I'm not completely sure what's going on here, but we should definitely return a 500.

pkg/handler/handler.go
```diff
@@ -55,13 +55,27 @@ type OpenAPIService struct {
 
 	lastModified time.Time
 
-	specBytes []byte
-	specPb    []byte
-	specPbGz  []byte
+	jsonCache    cache
+	protoCache   cache
+	protoGzCache cache
```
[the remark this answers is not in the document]
DangerOnTheRanger: Gotcha, done!

pkg/handler/handler.go
```diff
@@ -230,7 +237,11 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 					}
 
 					// serve the first matching media type in the sorted clause list
-					data, etag, lastModified := accepts.GetDataAndETag()
+					data, etag, lastModified, err := accepts.GetDataAndETag()
+					if err != nil {
+						w.WriteHeader(http.StatusInternalServerError)
```
apelisse, line 242: Should we log the error so that we know what's going on from the logs if this fails?

pkg/handler/handler.go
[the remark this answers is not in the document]
DangerOnTheRanger: Alright, 500s should be returned now.

apelisse:
Looks good to me, I'd like someone else to sign-off. Also hoping @sttts can approve :-)

pkg/handler/handler.go
```diff
@@ -230,7 +238,12 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 					}
 
 					// serve the first matching media type in the sorted clause list
-					data, etag, lastModified := accepts.GetDataAndETag()
+					data, etag, lastModified, err := accepts.GetDataAndETag()
+					if err != nil {
+						klog.Errorf("Error in OpenAPI handler: %s", err)
+						w.WriteHeader(http.StatusInternalServerError)
```
roycaihw, line 244: Instead of returning a 500, the old behavior was to return the previous spec when the server fails to serialize a new spec. I'm not sure if it ever occurs in reality, but is there a reason to change the behavior?

apelisse: @DangerOnTheRanger We can probably initialize the cache with the previous value, so that if it fails, we continue to have the value we had before, what do you think?

pkg/handler/handler_test.go
```diff
@@ -211,18 +211,27 @@ func TestCacheError(t *testing.T) {
 }
 
 func TestCachePreviousValue(t *testing.T) {
-	prevVal := []byte("previous value")
+	origVal := []byte("previous value")
 	cacheObj := cache{
```
apelisse, line 215: Maybe you should call the `New` function:
```golang
cache{}.New(func() ([]byte, error) {
...
```

DangerOnTheRanger: Gotcha, done

apelisse:
OK thanks, looks good to me!

@roycaihw we would appreciate your opinion! Thanks

Let's see how this impacts the performance on rapid CRD updates.

DangerOnTheRanger:
(forced push to fix a slightly incorrect error message in `handler_test.go` - line 241 should read `t.Fatalf("expected previous value for cache to be returned (got %s, expected %s)", value, lastGoodVal)` as opposed to `t.Fatalf("expected previous value for cache to be returned (got %s, expected %s)", lastGoodVal, value)`

pkg/handler/handler.go
```diff
@@ -55,13 +56,39 @@ type OpenAPIService struct {
...
+		bytes, err := c.BuildCache()
+		// if there is an error updating the cache, there can be situations where
+		// c.bytes contains a valid value (carried over from the previous update)
+		// but c.err is also not nil; the cache user is expected to check for this
+		c.err = err
+		if c.err == nil {
+			// don't override previous spec if we had an error
+			c.bytes = bytes
+		}
+		c.etag = computeETag(c.bytes)
```
roycaihw, line 82: nit: move this one line above. We don't need to re-calculate the etag if the bytes don't change

pkg/handler/handler.go
```diff
@@ -55,13 +56,39 @@ type OpenAPIService struct {
...
+	etag       string
+	err        error
+}
 
-	specBytesETag string
-	specPbETag    string
-	specPbGzETag  string
+func (c *cache) Get() ([]byte, string, error) {
+	c.once.Do(func() {
+		bytes, err := c.BuildCache()
```
roycaihw, line 73: I think this will panic if a query comes before the first `UpdateSpec`, given that `BuildCache` starts with a nil pointer

pkg/handler/handler.go
```diff
@@ -230,7 +250,15 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 					}
 
 					// serve the first matching media type in the sorted clause list
-					data, etag, lastModified := accepts.GetDataAndETag()
+					data, etag, lastModified, err := accepts.GetDataAndETag()
+					if err != nil {
+						klog.Errorf("Error in OpenAPI handler: %s", err)
+						// only return a 500 if we have no older cache data to serve
+						if data == nil {
+							w.WriteHeader(http.StatusInternalServerError)
```
roycaihw, line 258: Since we keep the old bytes, `data` can only be nil during bootstrap. I thought we could write an empty body. Or perhaps returning a 503 reflects the situation better. But I don't feel we need to return a 500. WDYT?

roycaihw (commented):
Thank you!! LGTM otherwise

ulucinar:
We have conducted [two experiments](https://github.com/crossplane/crossplane/issues/2649#issuecomment-956914007) on kind clusters using a node image built from this PR (8c165785553e390aaeee03baff9cff0c0e67e5fa) and another node image built from https://github.com/kubernetes/kubernetes/commit/8fd95902da533e521970edbf10108ff46739d5b9. Definitely improves the situation we reported in https://github.com/kubernetes/kubernetes/issues/105932. Thank you @DangerOnTheRanger !

pkg/handler/handler.go
```diff
@@ -71,6 +98,9 @@ func init() {
 }
 
 func computeETag(data []byte) string {
+	if data == nil {
+		return ""
```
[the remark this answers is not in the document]
DangerOnTheRanger: Passing `nil` to `sha512.Sum512` seems to return `CF83E1357EEFB8BDF1542850D66D8007D620E4050B5715DC83F4A921D36CE9CE47D0D13C5D85F2B0FF8318D2877EEC2F63B931BD47417A81A538327AF927DA3E`: https://play.golang.org/p/5EeaLx7Zkr4
I feel like returning an empty string when no data is passed seems like a much less-surprising result.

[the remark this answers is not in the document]
roycaihw: SGTM. Thanks!

pkg/handler/handler.go
DangerOnTheRanger: Makes sense - fixed!

pkg/handler/handler.go
DangerOnTheRanger: Good point - what do you think should be the correct behavior here? If we haven't called `UpdateSpec` yet we don't have a spec to marshal, if I'm following the logic here correctly. It seems like returning a 503 might be the way to go here.

apelisse: I agree that 503 is probably the way to go. Hard to do from here. Maybe always handle errors with 503 (when bytes is nil) and make sure you return an error if `BuildCache` is not set-up yet, WDYT?

roycaihw: To avoid the panic, I think we can initialize the caches with a `return nil, errors.New("...")` `BuildCache` function in `NewOpenAPIService`. Yeah resolving the error to a 503 sounds good to me

apelisse: Ah! That's a pretty clean way to do that!

DangerOnTheRanger: Now that I take a second look at the codepaths, is there actually a way to create `OpenAPIService` without passing a spec? `NewOpenAPIService` requires a spec to be passed in, and at that point we can pass the correct functions to `BuildCache`. Using `return nil, errors.New("...")` sounds like a very clean way to solve the issue if there is such a way, though!

(NewOpenAPIService for reference:)
```
// NewOpenAPIService builds an OpenAPIService starting with the given spec.
func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
	o := &OpenAPIService{}
	if err := o.UpdateSpec(spec); err != nil {
		return nil, err
	}
	return o, nil
}
```

pkg/handler/handler.go
DangerOnTheRanger: I think a 503 would fit well - from what I've been able to gather, actual errors during this point in the handler's lifecycle seem to be pretty rare/borderline nonexistent, so maybe a 503 would be a good bit better, semantically-speaking. @apelisse did you have any further thoughts?

pkg/handler/handler.go
roycaihw: Oh you're right! I think I was imagining a bootstrap scenario that never existed... Sorry for the confusion!

pkg/handler/handler.go
```diff
@@ -83,50 +114,40 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
...
+	})
+	o.protoCache = o.protoCache.New(func() ([]byte, error) {
+		json, _, err := o.jsonCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		return ToProtoBinary(json)
+	})
+	lastModified := time.Now()
 	o.lastModified = lastModified
```
apelisse, line 151: why do we create `lastModified` here?
```suggestion
	o.lastModified = time.Now()
```

pkg/handler/handler.go
apelisse: I don't have a strong opinion. 503 sounds good to me, though again as you mentioned, hopefully no one will ever see it :-)

pkg/handler/handler.go
nekto0n: Hi there! Sorry for jumping in. Was wondering won't this `lastModified` be evaluated independently on each api-server replica and result in different results? If so - can this break `Not Modified` responses from `http.ServeContent`?

DangerOnTheRanger: @apelisse Good question - I think it was left over from an earlier refactoring, and I missed cleaning it up, thanks for spotting that!

@nekto0n This is also a good question; I'm not sure what the answer is, so I hope someone else can jump in. As far as I know, the last-modified time was calculated like this before this PR, so it's consistent with previous behavior at least. It's another question entirely if that behavior was correct in the first place though, haha.

pkg/handler/handler.go
```diff
@@ -55,13 +56,40 @@ type OpenAPIService struct {
...
+	bytes      []byte
+	etag       string
+	err        error
+}
 
-	specBytesETag string
-	specPbETag    string
-	specPbGzETag  string
+func (c *cache) Get() ([]byte, string, error) {
+	c.once.Do(func() {
```
jiahuif, line 72:
```suggestion
	var err error
	c.once.Do(func() {
```

pkg/handler/handler.go
```diff
@@ -55,13 +56,40 @@ type OpenAPIService struct {
...
+		bytes, err := c.BuildCache()
+		// if there is an error updating the cache, there can be situations where
+		// c.bytes contains a valid value (carried over from the previous update)
+		// but c.err is also not nil; the cache user is expected to check for this
+		c.err = err
+		if c.err == nil {
+			// don't override previous spec if we had an error
+			c.bytes = bytes
+			c.etag = computeETag(c.bytes)
+		}
```
jiahuif, line 82:
```suggestion
		var bytes []byte
		bytes, err = c.BuildCache()
		// if there is an error updating the cache, there can be situations where
		// c.bytes contains a valid value (carried over from the previous update)
		// but err is also not nil; the cache user is expected to check for this
		if err == nil {
			// don't override previous spec if we had an error
			c.bytes = bytes
			c.etag = computeETag(c.bytes)
		}
```

pkg/handler/handler.go
```diff
@@ -55,13 +56,40 @@ type OpenAPIService struct {
...
+	})
+	return c.bytes, c.etag, c.err
```
jiahuif, line 84:
```suggestion
	return c.bytes, c.etag, err
```

pkg/handler/handler.go
```diff
@@ -55,13 +56,40 @@ type OpenAPIService struct {
...
+	jsonCache  cache
+	protoCache cache
+}
+
+type cache struct {
+	BuildCache func() ([]byte, error)
+	once       sync.Once
+	bytes      []byte
+	etag       string
+	err        error
```
jiahuif, line 68: The `err` field is a bit confusing. Readers may wonder of what process was this error produced. We can make it local. See suggestion below.

DangerOnTheRanger: Makes sense, I think only `cache.Get` was using that field. Fixed, thanks!

DangerOnTheRanger: Hmm, I'm not sure this handles the case of multiple failing calls to cache.Get(). The first call will return the error, but the second will return `nil`. https://play.golang.org/p/3RgPZBfAvCH outputs: `[SECOND CALL] expected non-nil err from cache.Get()`

jiahuif: So...should we retry or do we want to return the same error every time?

DangerOnTheRanger: I think saving `err` and returning it again is correct in the context of what we're trying to do here - an error that occurs during marshaling (which is what would get recorded here) is very likely to occur again IMO if we retry/call out to `json.Marshal` again, so it'd be best to wait for another call to `UpdateSpec` that should result in a fresh new spec to marshal.

I think in regard to [this discussion](https://github.com/kubernetes/kube-openapi/pull/251#discussion_r740640134), saving `err` makes sense as well - if we didn't save `err`, the first call would result in an HTTP 503, but any following calls would result in an HTTP 200 with an empty body.

jiahuif: I see...you can drop the latest commit to revert back

pkg/handler/handler.go
apelisse: I think that's fine. `lastModified` is used to evaluate the freshness of the cache, but never its correctness (we use the etag for that). We don't really use the freshness anywhere, but even if we did I don't think this would necessarily be a problem since it doesn't have to be exact like the etag does.

apelisse:
Thanks everyone, I think that looks awesome and I'm really excited by that change!

/lgtm
/approve

PR #252, opened 2021-10-07
Decouple REST Framework

### LINKED ISSUE #250: Decouple REST Framework
Currently, the `builder` package is tightly coupled with a 2017 version of [`go-restful`](https://github.com/emicklei/go-restful/). This dependency is exposed in three places:
* the builder's `BuildOpenAPISpec`, which takes a slice of `*restful.WebService`s
* the common config's `GetOperationIDAndTags` function, which takes a `*restful.Route`
* the handler's `BuildAndRegisterOpenAPIVersionedService`

It would be nice to be able to use this builder with newer versions of `go-restful`, but also with other frameworks. 

I propose taking ownership of the route interfaces and adding an adapter for the current version of `go-restful` to support the transition. We could add more adapters in this repo, or 3rd parties could implement them elsewhere. 

This would result in three breaking changes that I can see, with the two places the current dependency is exposed. Would this require a formal KEP?

austince:
As discussed in the #250, the `go-restful` dependency is currently exposed in a few places, making it difficult to use `kube-openapi` in projects using other frameworks (or versions of `go-restful`). This may make it easier to eventually move away from `go-restful` in k/k as well 🤷🏼.

This PR defines interfaces based on the `go-restful` types and exposes them instead. It also creates an adapter package to make it easy to use the current `go-restful` version with the new interfaces.

The interfaces are close to `go-restful`, but:
* Renames some fields to be more explicit for clarity (e.g., `ReadSample` -> `RequestPayloadSample`)
  * A few other candidates, if we want to go further, might be `Consumes` -> `ConsumesMIMETypes`, etc.
* Tries to deal with some known limitations of the `go-restful` types
  * Allows multiple status code responses w/ the same code (pain seen in: https://github.com/kubernetes/kubernetes/blob/e3ea169d7d8b2c1417ef5a71ee1015f186ca0e3c/staging/src/k8s.io/apiserver/pkg/endpoints/installer.go#L769-L770)

No breaking changes are introduced, though all places that expose the `restful` dependency instead of the new interfaces are marked as deprecated.

The PR also adds some small changes to integration tests/docs so its easier to develop. Can move that out of this PR if desired.

/cc @jpbetz @deads2k

Closes #250


pkg/common/restfuladapter/adapter.go
```diff
@@ -0,0 +1,16 @@
+package restfuladapter
+
+import (
+	"github.com/emicklei/go-restful"
```
deads2k, line 4: is this now the only package reliant on go-restful?

is there a golang way to hide this so that we don't end up with a transitive dependency for all those relying on this library?

austince:
(I haven't abandoned this PR, just waiting for https://github.com/kubernetes/kube-openapi/pull/247 to land since it is the priority and will have quite a few conflicts. :) )

austince:
/assign @sttts
/cc @jpbetz 

This current PR includes breaking changes. A couple months ago we talked about doing proper semver in this repo – wdyt, is that worth the effort? Otherwise, we could hide the breaking changes at the cost of making the `BuildOpenAPISpec` + config `GetOperationIDAndTags` funcs take an empty interface instead and do type switching..

jpbetz:
> /assign Dr. Stefan Schimanski /cc Joe Betz
> 
> This current PR includes breaking changes. A couple months ago we talked about doing proper semver in this repo – wdyt, is that worth the effort? Otherwise, we could hide the breaking changes at the cost of making the `BuildOpenAPISpec` + config `GetOperationIDAndTags` funcs take an empty interface instead and do type switching..

I'm generally in favor of not breaking backward compatibility unless we really need to. I think we can get by here without breaking changes.

How about we introduce a new `BuildOpenAPISpecFromRoutes` function and mark `BuildOpenAPISpec` as deprecated? If it's possible to have `BuildOpenAPISpec` delegate most of the implementation to `BuildOpenAPISpecFromRoutes` that would also be great, but at a quick glance that might not be easy to do here.

Same idea applies for `GetOperationIDAndTags`. Can we introduce a new function and deprecate the old one?

austince:
+1, thanks for the review & seems reasonable to me! The only side effect is that go-restful will be a dep even if unused. I'll make those changes by EOW.

austince:
@jpbetz I believe I've addressed those notes – all places that expose the `restful` dependency are deprecated and new funcs/fields are added with the new interfaces. Please have another look when you have a chance.

jpbetz:
This looks right to me.

/lgtm

jpbetz:
/assign @sttts 
Would you be willing to review for approval?

sttts:
Is there a proof PR in k/k ?

pkg/builder3/util.go
```diff
@@ -17,19 +17,19 @@ limitations under the License.
 package builder3
 
 import (
+	"k8s.io/kube-openapi/pkg/common"
```
sttts, line 20: nit: order

sttts:
Sgtm. Waiting for a proof PR before approval.

austince:
> Sgtm. Waiting for a proof PR before approval.

Is this what you're after in a proof PR? https://github.com/kubernetes/kubernetes/pull/107453

sttts:
Yes. Applied labels to see test results.

austince:
Turned up one error it seems. Fixed, working on covering the case in tests in this repo.

austince:
@sttts Proof PR tests are now all passing, please have another look. Thanks!

austince:
Thanks so much!

PR #253, opened 2021-10-12
Fork in part of github.com/google/cel-policy-templates-go

jpbetz:
Fork in the CEL OpenAPI integration code from [cel-policy-templates-go](https://github.com/google/cel-policy-templates-go) per discussion with @tristonianjones as a short term solution to get access to this code. Long term this code may become available as a library in a github repo, and we could would migrate to use it.

I've put the code under a `third_party/forked` directory (like the one that already exists in github.com/kubernetes/kubernetes) and added a README to make the origin and status of this code clear.

I decided that this code is better put here in kube-openapi rather than github.com/kubernetes/kubernetes because we need to use it in the `pkg/validation/validate`.

I've matched the versions of the existing `github.com/golang/protobuf` and transitively added `gopkg.in/yaml.v3`, `google.golang.org/genproto`, `google.golang.org/protobuf` modules to match the exact version already in github.com/kubernetes/kubernetes. This was easy to do since Kubernetes had slightly newer version of all of them.

cc @cici37 @liggitt


third_party/forked/celopenapi/model/schemas.go
```diff
@@ -0,0 +1,398 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"k8s.io/kube-openapi/pkg/validation/spec"
```
TristonianJones, line 18: Any thoughts on a possibly moving toward a `Schema` interface, or factory of sorts so we could swap the internals as needed in the future. I'm asking more from the standpoint of creating a standalone CEL library for this than I am about what to do right this moment.

go.mod
```diff
@@ -9,9 +9,10 @@ require (
 	github.com/emicklei/go-restful v0.0.0-20170410110728-ff4f55a20633
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/swag v0.19.5
-	github.com/golang/protobuf v1.4.2
+	github.com/golang/protobuf v1.5.2
+	github.com/google/cel-go v0.7.3
```
TristonianJones, line 13: We can version bump this to v0.8.0.

third_party/forked/celopenapi/model/schemas.go
```diff
@@ -0,0 +1,398 @@
...
+func SchemaDeclTypes(s *spec.Schema, maybeRootType string) (*DeclType, map[string]*DeclType) {
+	root := SchemaDeclType(s).MaybeAssignTypeName(maybeRootType)
+	types := FieldTypeMap(maybeRootType, root)
+	return root, types
+}
+
+// SchemaDeclType returns the CEL Policy Templates type name associated with the schema element.
+func SchemaDeclType(s *spec.Schema) *DeclType {
+	if len(s.Type) > 1 {
+		// TODO
```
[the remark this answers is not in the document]
jpbetz: I don't know the background on the decision to allow more than 1 type. https://github.com/kubernetes/kube-openapi/blob/b3fe75cc9b2fe5d5dc7929832b1516bb9b22362f/pkg/validation/validate/schema.go#L136 shows how the validator code in kube-openapi deals with the cardinality.

I've restructured this function slightly so that it tries all the types. I'm not 100% certain this is needed, but it should be safe.

jpbetz:
@TristonianJones feedback addressed (sorry for the delay-- I blame kubecon!). Ready for another pass if you get a moment.

go.mod
jpbetz: Bumped.

third_party/forked/celopenapi/model/schemas.go
```diff
@@ -0,0 +1,394 @@
...
+func SchemaDeclTypes(s *spec.Schema, maybeRootType string) (*DeclType, map[string]*DeclType) {
+	root := SchemaDeclType(s).MaybeAssignTypeName(maybeRootType)
+	types := FieldTypeMap(maybeRootType, root)
+	return root, types
+}
+
+// SchemaDeclType returns the CEL Policy Templates type name associated with the schema element.
+func SchemaDeclType(s *spec.Schema) *DeclType {
+	if len(s.Type) < 1 {
+		panic("no type")
```
apelisse, line 35: isn't it common for type to be missing?

third_party/forked/celopenapi/model/schemas.go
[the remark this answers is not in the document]
TristonianJones: In that case, you'll have to return a `decls.Dyn` value for now and refine the experience later.

third_party/forked/celopenapi/model/schemas.go
apelisse: That seems consistent with what we've done in the past, thanks!

PR #257, opened 2021-10-18
[WIP] Add Cel compilation

cici37:
This adds CEL compilation process which will be used by both CRD registration validation and CR validation.

This PR is based on #253 until it merges.

Support for map/list will be added in a separate commit/PR.


pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+
+// CelRule defines the format of each rule in CelRules.
+type CelRule struct {
+	Rule    string `json:"rule"`
+	Message string `json:"message"`
+}
+
+func CompileCel(schema *spec.Schema, CelRules CelRules, rootName string) ([]cel.Program, []error) {
+	if len(rootName) == 0 {
+		rootName = "__root__"
```
jpbetz, line 40: Add a comment explaining what this is? My understanding is that we're using it as the type name of the provided schema.

Also, in hindsight (I know I picked this name originally), I think a name like `__expression_scope__` might be better.

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+			} else {
+				programs[i] = prog
+			}
+		}
+	}
+
+	return programs, allErrors
+}
+
+func CompileCelForScalarElement(CelRules CelRules, schemaName string, schemaType string) ([]cel.Program, []error) {
```
jpbetz, line 89: Can this be merged with the above function? It might make sense to have a unexported utility function, but it seems like it would be cleaner to have a single entry function that does the right thing for whatever schema is passed in.  If we pass in a schema for a scalar type, then we do what this function does.

pkg/validation/validate/cel_compilation_test.go
```diff
@@ -0,0 +1,150 @@
...
+						"stringTest": {
+							SchemaProps: spec.SchemaProps{
+								Type: []string{"string"},
+							},
+						},
+					},
+				},
+			},
+			expr:      "stringTest.startsWith('s')",
+			wantError: false,
```
jpbetz, line 93: why is wantError false?

jpbetz: Also, I'd expect these tests to check that they get the right error back.

pkg/validation/validate/cel_compilation_test.go
```diff
@@ -0,0 +1,150 @@
...
+			if !tt.wantError && len(allErrors) > 0 {
+				t.Errorf("Expected no error, but got: %v", allErrors)
+			} else if tt.wantError && len(allErrors) == 0 {
+				t.Error("Expected error, but got none")
+			}
+		})
+	}
+}
+
+func TestCompileCelForScalarElement(t *testing.T) {
```
jpbetz, line 109: Same comment as above, I don't think we want to have a completely separate code path for scalar types. I don't feel as strongly about the tests, but I suspect it will be easier to do combined testing once we unify the complication.

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		allErrors = append(allErrors, err)
+		return nil, allErrors
+	}
+
+	programs := make([]cel.Program, len(CelRules))
+	for i, rule := range CelRules {
+		ast, issues := env.Compile(rule.Rule)
+		if issues != nil {
+			allErrors = append(allErrors, fmt.Errorf("compilation failed for rule: %v with message: %v", rule.Message, issues.Err()))
```
jpbetz, line 75: Is there a plan for including the field path in the schema to this error so the developer knows where to find the failing cel rule in their openapiv3 schema?

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+	return programs, allErrors
+}
+
+func CompileCelForScalarElement(CelRules CelRules, schemaName string, schemaType string) ([]cel.Program, []error) {
+	allErrors := []error{}
+
+	var celDecls []*expr.Decl
+	switch schemaType {
+	case "string":
+		celDecls = append(celDecls, decls.NewVar(schemaName, decls.String))
```
jpbetz, line 95: s/schemaName/fieldName ?

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+// CelRules defines the format of the x-kubernetes-validator schema extension.
+type CelRules []CelRule
+
+// CelRule defines the format of each rule in CelRules.
+type CelRule struct {
+	Rule    string `json:"rule"`
+	Message string `json:"message"`
+}
+
+func CompileCel(schema *spec.Schema, CelRules CelRules, rootName string) ([]cel.Program, []error) {
```
jpbetz, line 38: +godoc on all exported functions

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return programs, allErrors
+}
+
+func CompileCelForScalarElement(CelRules CelRules, schemaName string, schemaType string) ([]cel.Program, []error) {
+	allErrors := []error{}
+
+	var celDecls []*expr.Decl
+	switch schemaType {
```
jpbetz, line 93: What about "number" for float and double formats in OpenAPIv3. xref https://swagger.io/specification/#data-types

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+	case "string":
+		celDecls = append(celDecls, decls.NewVar(schemaName, decls.String))
+	case "integer":
```
jpbetz, line 96: integer can be either int32 or int64 depending on what format is set to. Do we need to set the decl differently for these?

pkg/validation/validate/cel_compilation.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	opts, err := rt.EnvOptions(env.TypeProvider())
+	if err != nil {
+		allErrors = append(allErrors, err)
+		return nil, allErrors
+	}
+
+	var propDecls []*expr.Decl
+	if root, ok := rt.FindDeclType(rootName); ok {
+		if root.IsObject() {
```
jpbetz, line 58: This is where I was expecting the other schema types to be handled (scalars, maps, arrays).

jpbetz: There should be a final else clause that returns an error if the provided schema isn't any of the types supported here.

cici37: I added checking for non supported type. For arrays/scalars/maps, it should have been declared properly

pkg/validation/validate/cel_compilation.go
jpbetz: But we still need cases in this switch statement for them, right?

pkg/util/cel/cel_compilation.go
```diff
@@ -0,0 +1,117 @@
...
+
+// CelRule defines the format of each rule in CelRules.
+type CelRule struct {
+	Rule    string `json:"rule"`
+	Message string `json:"message"`
+}
+
+// Compile is used for cel compilation.
+// rootName is used as type name of provided schema
+func Compile(schema *spec.Schema, CelRules CelRules, rootName string) ([]cel.Program, []error) {
```
jpbetz, line 40: nit: should be `celRules CelRules` to match go conventions

pkg/util/cel/cel_compilation.go
```diff
@@ -0,0 +1,117 @@
...
+		var err error
+		env, err = cel.NewEnv(
+			celext.Strings(),
+			celext.Encoders(),
+			cel.Declarations(propDecls...))
+		if err != nil {
+			allErrors = append(allErrors, fmt.Errorf("error initializing CEL environment: %w", err))
+			return nil, allErrors
+		}
+	} else {
```
jpbetz, line 70: Why have such a different environment depending on if the type is scalar or not? Can we consolidate this else if the the above if more? I'd expect the propDecls to be the only thing that differs.

pkg/util/cel/cel_compilation_test.go
```diff
@@ -0,0 +1,241 @@
...
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expr:      "size(nestedObj[0]) == 10",
```
jpbetz, line 163: show that we can dereference into the deeply nested array?

pkg/util/cel/cel_compilation_test.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+		{
+			name: "invalid type",
+			input: spec.Schema{
+				SchemaProps: spec.SchemaProps{
+					Type: []string{"nonSupported"},
+				},
+			},
+			expr:      "unsupported type check",
+			wantError: true,
```
jpbetz, line 61: Need to verify that the error is the one you expect here.

Also need to make sure to cover the various complication errors that can happen.

pkg/util/cel/cel_compilation_test.go
```diff
@@ -0,0 +1,241 @@
...
+*/
+
+package cel
+
+import (
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"testing"
+)
+
+func TestCelCompilation(t *testing.T) {
```
jpbetz, line 24: Also need map cases for type="object", Properties unset and AdditionalProperties=something

pkg/validation/validate/cel_compilation.go
cici37: I have added number in switch float/double

pkg/util/cel/cel_compilation_test.go
```diff
@@ -0,0 +1,241 @@
...
+						},
+					},
+				},
+			},
+			expr:      "size(nestedObj[0]) == 10",
+			wantError: false,
+		},
+		{
+			name: "valid nested array of array",
+			input: spec.Schema{
```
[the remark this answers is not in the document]
cici37: The input is different for each test case instead of copy-paste. So have a package level variable for schema is not gonna help here.

PR #263, opened 2021-11-05
Recursive mutating schema walker.

jiahuif:
This PR refactors `pkg/aggregator/mutating_walker.go` into a more generic schema walker.
plus fuzz testing for Schema mutations

### DIFF
--- a/pkg/schemamutation/walker.go
+++ b/pkg/schemamutation/walker.go
@@ -14,28 +14,45 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package aggregator
+package schemamutation
 
 import (
-	_ "net/http/pprof"
-
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
-// Run a walkRefCallback method on all references of an OpenAPI spec, replacing the values.
-type mutatingReferenceWalker struct {
-	// walkRefCallback will be called on each reference. Do not mutate the input, always create a copy first and return that.
-	walkRefCallback func(ref *spec.Ref) *spec.Ref
+// Walker runs callback functions on all references of an OpenAPI spec,
+// replacing the values when visiting corresponding types.
+type Walker struct {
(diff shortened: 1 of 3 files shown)


pkg/schemamutation/walker.go
```diff
@@ -48,14 +64,18 @@ func (w *mutatingReferenceWalker) walkSchema(schema *spec.Schema) *spec.Schema {
 		}
 	}
 
-	if r := w.walkRefCallback(&schema.Ref); r != &schema.Ref {
+	// Always callback on the whole schema first so that other callbacks
```
jpbetz, line 67: The callbacks are pre-order DFS? That should probably be in the godoc of the walker of interface (and the call order is needs to be part of the interface contract).

pkg/schemamutation/walker.go
```diff
@@ -14,28 +14,44 @@ See the License for the specific language governing permissions and
...
-	_ "net/http/pprof"
-
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
-// Run a walkRefCallback method on all references of an OpenAPI spec, replacing the values.
-type mutatingReferenceWalker struct {
-	// walkRefCallback will be called on each reference. Do not mutate the input, always create a copy first and return that.
-	walkRefCallback func(ref *spec.Ref) *spec.Ref
+// Walker runs callback methods on all references of an OpenAPI spec,
```
jpbetz, line 23: very minor nit: s/methods/functions

jpbetz (commented):
Minor doc comments, lgtm otherwise

pkg/schemamutation/walker.go
jiahuif: This is a DFS but callbacks are not recursive. The recursive part is with `WalkSchema` down below.

Added comments about the call ordering in the type and impl.

apelisse:
/lgtm

Thanks!

PR #265, opened 2021-11-08
cherry-picking #220 and #251 to release-1.21

muvaf:
Cherry-pick of https://github.com/kubernetes/kube-openapi/pull/251 into release-1.21 as discussed in the sig-apimachinery meeting and over Slack. Details about why it's needed are here in [this issue](https://github.com/kubernetes/kubernetes/issues/105932).

I needed to cherry-pick https://github.com/kubernetes/kube-openapi/pull/220 as well because `ToProtoBinary` function in `handler.go` uses a new method coming in `v0.5.1` of gnostic which was updated in that PR.


apelisse:
/approve
/lgtm

@jiahuif Do we have a problem with our configuration that prevents the tests from running on branches beside master?

jiahuif:
@apelisse It is just flaky. The problem was the fixed default test timeout that the integration test may or may not meet.

go.mod
```diff
@@ -11,18 +11,17 @@ require (
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/spec v0.19.3
 	github.com/go-openapi/swag v0.19.5
-	github.com/golang/protobuf v1.3.3
-	github.com/google/go-cmp v0.4.0 // indirect
+	github.com/golang/protobuf v1.4.2
```
liggitt, line 14: we should not bump dependencies this much on release branches... backporting the lazy loading is fine, but have it lazily do whatever was being done on that release branch (for 1.21 and 1.20, `json marshal → json unmarshal → map[string]interface{} → proto`) rather than bumping dependencies significantly

apelisse: @muvaf Are you able to undo that specific change?

liggitt: might be easier to revert both commits, repick just the `Add lazy marshaling to handler.go` commit, and resolve the conflict in the lazy proto marshaler to do the `json marshal → json unmarshal → map[string]interface{} → proto` bit

muvaf: Thanks for the feedback, Yes, I can do that for both 1.20 and 1.21 branches in the next hour and let you know.

muvaf: @apelisse The PRs are ready:
* release-1.20: https://github.com/kubernetes/kube-openapi/pull/268
* release-1.21: https://github.com/kubernetes/kube-openapi/pull/267

PR #268, opened 2021-11-09
revert #266 and cherry-pick only #251

muvaf:
As result of [this discussion](https://github.com/kubernetes/kube-openapi/pull/265#discussion_r746087650), here is a PR that reverts two commits and cherry-picks only the lazy marshaling PR. One thing I needed to change was to unmarshal the JSON blob into `map[string]interface{}` before giving it to `ToProtoBinary` since its argument type changed with #220 .


apelisse:
OK Thanks, 
/lgtm
/approve
For Jordan:
/hold

muvaf:
@liggitt `git diff d219536bb9fd63731fa65ca4b42f14bc2f10c61b` (the commit `release-1.20` was created with) returns no diff for `go.mod` and `go.sum`.

liggitt:
I ran unit tests locally and they passed, but I didn't see anything running gating merge

PR #269, opened 2021-11-15
OpenAPI Enums: Format update

jiahuif:
This PR updates the format of enum description so it can be more consistent with existing godoc.

BEFORE
```
Possible enum values:

- `apple`: is the Apple
- `banana`: is the Banana
- `onigiri`: is the Rice ball that does not seem to belong to a fruits basket but has a long comment that is so long that it spans multiple lines"
```

AFTER
```
Possible enum values:

- `"apple"` is the Apple
- `"banana"` is the Banana
- `"onigiri"` is the Rice ball that does not seem to belong to a fruits basket but has a long comment that is so long that it spans multiple lines"
```

### DIFF
--- a/pkg/generators/enum.go
+++ b/pkg/generators/enum.go
@@ -74,15 +74,15 @@ func (et *enumType) ValueStrings() []string {
 // DescriptionLines returns a description of the enum in this format:
 //
 // Possible enum values:
-//  - `value1`: description 1
-//  - `value2`: description 2
+//  - `"value1"` description 1
+//  - `"value2"` description 2
 func (et *enumType) DescriptionLines() []string {
 	var lines []string
 	for _, value := range et.Values {
 		lines = append(lines, value.Description())
 	}
 	sort.Strings(lines)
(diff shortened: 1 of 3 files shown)


pkg/generators/enum.go
```diff
@@ -74,15 +74,16 @@ func (et *enumType) ValueStrings() []string {
 // DescriptionLines returns a description of the enum in this format:
 //
 // Possible enum values:
-//  - `value1`: description 1
-//  - `value2`: description 2
+//
```
apelisse, line 77: Do we really want that empty line?

apelisse: I actually don't see it in the test, which is good.

apelisse (commented):
Very minor nit. Can you also squash?

apelisse:
FYI @thockin

pkg/generators/enum.go
jiahuif: Sorry straw format. Fixing & squashing

PR #270, opened 2021-11-23
Add support for aliases of enum types and constants

jiahuif:
TODO: fix gengo to properly handle alias of a type alias.

Ref: https://github.com/kubernetes/kubernetes/pull/106300


apelisse:
Before we fix that code, I really want us to describe what we're planning on doing. Writing a good test suite would be a good way to get started?

jiahuif:
Right. Now that we are in code freeze we would have time to take a better look at the problem.

sftim:
(nit)
/retitle Add support for aliases of enum types and constants

PR #271, opened 2021-12-06
go.mod: update to github.com/emicklei/go-restful/v3 v3.7.3

thaJeztah:
relates to https://github.com/kubernetes/kubernetes/pull/106841

full diff: https://github.com/emicklei/go-restful/compare/ff4f55a20633...v3.7.3


austince:
I think this is a breaking change, since restful v1 is part of the exposed API. If https://github.com/kubernetes/kube-openapi/pull/252 lands, we could support restful v3 via an adapter in this PR. WDYT?

PR #273, opened 2021-12-08
Set path for specific validators when set on parent SchemaValidator and use JSONPath syntax

### LINKED ISSUE #262: Field path reported in validation errors within slice is incorrect
I noticed when validating a custom resource using the apiextensions [NewSchemaValidator](https://github.com/kubernetes/apiextensions-apiserver/blob/4c1a29cbb037397b8bd796378443a06075a64a83/pkg/apiserver/validation/validation.go#L33) (which is using the [SchemaValidator](https://github.com/kubernetes/kube-openapi/blob/ee342a809c29ce286ab48cea8442ef9cd0ab8c1c/pkg/validation/validate/schema.go#L34) in this repo) that when a validation error occurs within a slice, the index of the slice is omitted from the fieldpath returned in the error.

YAML snippet:
```
spec:
  dependsOn:
    - provider: registry.upbound.io/crossplane/provider-aws
      version: 1
```
Output:
```
spec.dependsOn.version in body must be of type string: "integer"
```

I looked into why this is happening by traversing through the recursive validation that is performed against the schema. When we get to a the `sliceValidator` for `spec.dependsOn`, it's `Applies()` call returns `true`, causing it to be run.

Screenshot from 2021-11-05 11-51-07

Looking good so far. When the validation actually runs, we get the length for the slice and construct a new `SchemaValidator` that we will reuse to validate each item (just one item in this case).

Screenshot from 2021-11-05 11-54-40

When we first construct this nested `SchemaValidator` we set the path to match the parent validator (`spec.dependsOn`). This is not accurate though and needs to be updated on each iteration, so we call `SetPath()`. After setting the path, we can see that the validator has the proper index appended (`spec.dependsOn.0`).

Screenshot from 2021-11-05 11-58-53

Still looking good (though the format is a bit strange, more on this below). Unfortunately, the [initial call](https://github.com/kubernetes/kube-openapi/blob/ec1f4c89925aaed16de489e6daa816bed3bd2811/pkg/validation/validate/schema.go#L81) to `NewSchemaValidator()` is when each of its child validators are instantiated (e.g. `formatValidator`, `objectValidator`, etc.), and the subsequent call to `SetPath()` updates the `SchemaValidator` path, but none of the child validator paths (so now the `SchemaValidator` has path `spec.dependsOn.0`, but all child validators have path `spec.dependsOn`). This can be observed when we iterate through each of the child validators and compare their path with the `SchemaValidator`:

Screenshot from 2021-11-05 12-04-24

Unfortunately this means that if an error occurs anywhere in any descendant validator, the index of the slice is omitted from the field path (as seen at beginning of the ticket). I have also verified that this is happening in the API server validation (that consumes this repo):

> Note: this example is using a different resource.
```
$ kubectl apply -f definition.yaml --validate=false
The CompositeResourceDefinition "compositepostgresqlinstances.database.example.org" is invalid: spec.versions.referenceable: Invalid value: "string": spec.versions.referenceable in body must be of type boolean: "string"
```

I imagine most users are not catching this because the client-side validation provides a more accurate error:
```
$ kubectl apply -f definition.yaml 
error: error validating "definition.yaml": error validating data: ValidationError(CompositeResourceDefinition.spec.versions[0].referenceable): invalid type for io.crossplane.apiextensions.v1.CompositeResourceDefinition.spec.versions.referenceable: got "string", expected "boolean"; if you choose to ignore these errors, turn validation off with --validate=false
```

## Proposal

I would like to propose two changes, and I am happy to implement them:
1. Set path for child validators in `SchemaValidator` `SetPath()`

The simplest fix is to set the path for all child validators when the path is set on `SchemaValidator`. Though I believe there could be some more elegant handling of paths while performing this traversal, this does address the issue and is a lightweight change to fix this in the immediate term.
```
// SetPath sets the path for this schema validator
func (s *SchemaValidator) SetPath(path string) {
	s.Path = path
	for _, v := range s.validators {
		v.SetPath(path)
	}
}
```
I have verified that this works correctly:
```
spec.dependsOn.0.version in body must be of type string: "integer"
```

2. Use [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html) syntax for indices

To be more aligned with client-side validation, and for a more generally accepted format, it would be great to represent that index of a slice as `dependsOn[0]` instead of `dependsOn.0`.

If both of these fixes sound reasonable to folks, I'll go ahead and open PRs to address.

/assign
/bug

hasheddan:
Currently if we set the path for the SchemaValidator after creation it
does not propagate down to the specific validators embedded within. That
means that subsequent SchemaValidators that are created by the specific
validators will have the wrong path. This updates to set the path for
all specific validators when it is set on their parent SchemaValidator.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Updates to use proper JSONPath bracket syntax for indices in a path.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

/assign @apelisse 

Fixes #262

### DIFF
--- a/pkg/validation/validate/schema.go
+++ b/pkg/validation/validate/schema.go
@@ -99,6 +99,9 @@ func NewSchemaValidator(schema *spec.Schema, rootSchema interface{}, root string
 // SetPath sets the path for this schema validator
 func (s *SchemaValidator) SetPath(path string) {
 	s.Path = path
+	for _, v := range s.validators {
+		v.SetPath(path)
+	}
 }
 
 // Applies returns true when this schema validator applies
--- a/pkg/validation/validate/schema_props.go
+++ b/pkg/validation/validate/schema_props.go
@@ -41,6 +41,18 @@ type schemaPropsValidator struct {
 
 func (s *schemaPropsValidator) SetPath(path string) {
 	s.Path = path
+	for _, v := range s.anyOfValidators {
+		v.SetPath(path)
+	}
+	for _, v := range s.allOfValidators {
+		v.SetPath(path)
+	}
+	for _, v := range s.oneOfValidators {
+		v.SetPath(path)
+	}
+	if s.notValidator != nil {
+		s.notValidator.SetPath(path)
+	}
 }
 
 func newSchemaPropsValidator(path string, in string, allOf, oneOf, anyOf []spec.Schema, not *spec.Schema, deps spec.Dependencies, root interface{}, formats strfmt.Registry, options ...Option) *schemaPropsValidator {
--- a/pkg/validation/validate/object_validator.go
+++ b/pkg/validation/validate/object_validator.go
@@ -17,7 +17,6 @@ package validate
 import (
 	"reflect"
 	"regexp"
-	"strings"
 
 	"k8s.io/kube-openapi/pkg/validation/errors"
 	"k8s.io/kube-openapi/pkg/validation/spec"
@@ -49,21 +48,6 @@ func (o *objectValidator) Applies(source interface{}, kind reflect.Kind) bool {
 	r := reflect.TypeOf(source) == specSchemaType && (kind == reflect.Map || kind == reflect.Struct)
 	debugLog("object validator for %q applies %t for %T (kind: %v)\n", o.Path, r, source, kind)
 	return r
-}
-
-func (o *objectValidator) isProperties() bool {
-	p := strings.Split(o.Path, ".")
-	return len(p) > 1 && p[len(p)-1] == jsonProperties && p[len(p)-2] != jsonProperties
-}
-
-func (o *objectValidator) isDefault() bool {
-	p := strings.Split(o.Path, ".")
-	return len(p) > 1 && p[len(p)-1] == jsonDefault && p[len(p)-2] != jsonDefault
-}
-
-func (o *objectValidator) isExample() bool {
-	p := strings.Split(o.Path, ".")
-	return len(p) > 1 && (p[len(p)-1] == swaggerExample || p[len(p)-1] == swaggerExamples) && p[len(p)-2] != swaggerExample
 }
 
 func (o *objectValidator) Validate(data interface{}) *Result {
--- a/pkg/validation/validate/slice_validator.go
+++ b/pkg/validation/validate/slice_validator.go
@@ -56,7 +56,7 @@ func (s *schemaSliceValidator) Validate(data interface{}) *Result {
 	if s.Items != nil && s.Items.Schema != nil {
 		validator := NewSchemaValidator(s.Items.Schema, s.Root, s.Path, s.KnownFormats, s.Options.Options()...)
 		for i := 0; i < size; i++ {
-			validator.SetPath(fmt.Sprintf("%s.%d", s.Path, i))
+			validator.SetPath(fmt.Sprintf("%s[%d]", s.Path, i))
 			value := val.Index(i)
 			result.Merge(validator.Validate(value.Interface()))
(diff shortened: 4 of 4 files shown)


liggitt:
Does this only affect the field path in existing returned errors, or is this path also used for anything that would make this fix start returning errors in cases it previously didn't?

hasheddan:
@liggitt I believe this field path is not currently used for anything other than providing information about where an error occurred, so I do not expect that changing it to reflect the correct path would have any negative consequences.

pkg/validation/validate/schema.go
```diff
@@ -99,6 +99,9 @@ func NewSchemaValidator(schema *spec.Schema, rootSchema interface{}, root string
 // SetPath sets the path for this schema validator
 func (s *SchemaValidator) SetPath(path string) {
 	s.Path = path
+	for _, v := range s.validators {
+		v.SetPath(path)
+	}
```
sttts, line 104: would like to see a test case where this goes wrong before the PR and is fixed afterwards.

liggitt:
Is the same change needed here?

https://github.com/kubernetes/kube-openapi/blob/e816edb12b65975944ee23073f35617fd9d49215/pkg/validation/validate/schema_props.go#L42-L44

> @liggitt I believe this field path is not currently used for anything other than providing information about where an error occurred, so I do not expect that changing it to reflect the correct path would have any negative consequences.

These look like they affect logic / code paths:

https://github.com/kubernetes/kube-openapi/blob/e816edb12b65975944ee23073f35617fd9d49215/pkg/validation/validate/object_validator.go#L54-L67

pkg/validation/validate/schema.go
hasheddan: @sttts there is a pretty comprehensive write up of exactly what is happening, along with an example from before and after the fix in https://github.com/kubernetes/kube-openapi/issues/262. Are there additional cases that would be more helpful for review?

sttts: Indeed comprehensive :)

sttts:
/assign @liggitt 

Anything open from Jordan's comment?

hasheddan:
> Is the same change needed here?

@liggitt I believe so due to the fact that the `schemaPropsValidator` is embedded in the `SchemaValidator` _and_ embeds its own `SchemaValidators`. When we call `SetPath()` on the parent `SchemaValidator`, it calls `SetPath()` on the `schemaPropsValidator`, which we need to then call `SetPath()` on each of its embedded `SchemaValidators` (and so on). I have pushed a commit to address this.

> edit: I don't actually see those called anywhere, so maybe they're not used

@liggitt you are correct that they are not used -- I have now removed them to avoid confusion. Let me know if you prefer to keep them around.

pkg/validation/validate/schema_props.go
```diff
@@ -41,6 +41,18 @@ type schemaPropsValidator struct {
...
+	}
+	for _, v := range s.allOfValidators {
+		v.SetPath(path)
+	}
+	for _, v := range s.oneOfValidators {
+		v.SetPath(path)
+	}
+	if s.notValidator != nil {
+		s.notValidator.SetPath(path)
+	}
```
hasheddan, line 55: Note for onlookers: we would not panic on calling `SetPath()` on a `nil` `notValidator` except for the fact that `SetPath()` accesses the `Path` struct member and does not check if the receiver (i.e. itself) is `nil`, which would cause a panic. Therefore, we check that `notValidator` is not `nil` before calling `SetPath()`.

PR #275, opened 2022-01-18
support Deprecated field

Jefftree:
go-restful is at version v2.9.5 in k/k already: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L36

This PR supports publishing the deprecated field for an operation in both OpenAPI v2 and v3.

See https://github.com/kubernetes/kubernetes/pull/107622 for how swagger.json changes with this update. 

See https://github.com/kubernetes/kubernetes/issues/107498 for more details on the issue.

/cc @apelisse @liggitt @sttts

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/emicklei/go-restful/v3 v3.8.0
-	github.com/getkin/kin-openapi v0.76.0
+ 	github.com/getkin/kin-openapi v0.76.0
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/swag v0.19.5
 	github.com/golang/protobuf v1.5.2
@@ -14,7 +14,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.2
-	github.com/mitchellh/mapstructure v1.1.2
+ 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/munnerz/goautoneg v0.0.0-20120707110453-a547fc61f48d
 	github.com/onsi/ginkgo/v2 v2.1.4
 	github.com/onsi/gomega v1.19.0
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/emicklei/go-restful/v3 v3.8.0
- 	github.com/getkin/kin-openapi v0.76.0
+	github.com/getkin/kin-openapi v0.76.0
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/swag v0.19.5
 	github.com/golang/protobuf v1.5.2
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.2
 	github.com/googleapis/gnostic v0.5.1
+	github.com/json-iterator/go v1.1.12 // indirect
```
liggitt, line 18: this is unfortunate (xref https://github.com/kubernetes/kube-openapi/pull/246)... is this actually used?

pkg/builder3/openapi.go
```diff
@@ -74,6 +74,7 @@ func (o *openAPI) buildOperations(route common.Route, inPathCommonParamsMap map[
 	ret := &spec3.Operation{
```
liggitt, line 74: for openapi v3, what about the deprecated field on the schema object as well as the operation?

apelisse: Yes, that'd be super sweet, especially if we automatically enable new mechanisms for help with deprecation.

Jefftree: Yeah it should possible but it's not as straightforward and I thought it might be better to include in a follow up after some more discussion.

I'm still doing some investigations, but afaik we can't guarantee that a deprecated operation path would deprecate all the corresponding schemas associated with that path. An example of this is a delete on an alpha/beta resource that returns a meta.v1.Status. Is this a one-off exception or are there more examples of this?

For schema fields, there are cases where we deprecate a single schema field and not the entire schema. This is currently embedded in field descriptions, and it may be beneficial to include these programmatically as well.

liggitt: If you look at how this deprecated operation setting is being driven from k/k, it's based on the entire primary struct for a set of operations being deprecated. Even if it's in another PR, having a way to plumb the fact that that struct is deprecated into openapi would be useful.

apelisse: OK, thanks for clarifying, I completely agree.

pkg/common/interfaces.go
```diff
@@ -36,6 +36,8 @@ type Route interface {
 	// StatusCodeResponses defines a mapping of HTTP Status Codes to the specific response(s).
 	// Multiple responses with the same HTTP Status Code are acceptable.
 	StatusCodeResponses() []StatusCodeResponse
+	// Deprecated marks a route as deprecated
+	Deprecated() bool
```
austince, line 40: Will this be released as a breaking change? iirc, we were quite careful when adding this interface to not break backwards compatibility, suggest doing the same here

apelisse:
@Jefftree what's the plan here?

pkg/common/interfaces.go
Jefftree: I did a quick scan of the k/k codebase, it looks like nothing is using the particular interface and all calls of the interface are from kube-openapi itself. The GetOperationIDAndTagsFromRoute in the PR linked has not yet been integrated into k/k yet. 

So to answer your question, yes this will be a breaking change for current users of the API, but at the moment it seems like we have no users of the particular API (that we know of). I'm okay with taking the safe approach of just creating another interface, just want to gather some thoughts first.

Jefftree:
@apelisse I think this provides valuable information to the OpenAPI and we should look to get it in once the pending comment is addressed.

pkg/common/interfaces.go
austince: Thanks for your reply. I agree it isn't a huge breaking change and won't affect k/k, though will affect other projects using it (mine 😄), perhaps it's acceptable for this project to only support k/k.
Do you think it would be valuable to switch k/k to the interface, or rather wait until there's a push to get off go-restful?

Jefftree: Thanks for pointing that out, that's enough of a reason to create backwards compatibility and create a new interface. Do you have any preferences for the interface name?

I don't seem any harm in switching k/k to the interface, although that will be a decision up to the approvers.

austince: Thanks very much for the backwards compat effort :) No real preference on name, perhaps `DeprecatableRoute`?

PR #278, opened 2022-02-14
Cooldown-based eager marshaling

jiahuif:
With #251, the marshaling of our OpenAPI spec is delayed until the first request, so that the spec is not re-marshaled every time a new CRD is added, which saves a lot of marshaling time. However, I've encountered the following frustration with our current implementation.

- The API server is idle after CRDs are added and before first request comes. Considering in a real-world scenario, the user usually start adding workload after a few minutes, there is usually plenty of time for the API server to "warm-up" before first request from the user.
- In a replicated/HA control plane, there are multiple instances of API server. The first request warms up only one of the instances. In other words, for N instances, the first N requests are "warm-up"s and slow.

This PR addresses the issues by adding a cooldown. If the spec is not updated within the cooldown, build the cache by marshaling the spec before the first request comes. This way, we can avoid having to frequently re-marshal the spec while keeping first requests fast.

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
+	"math"
 	"mime"
 	"net/http"
 	"sync"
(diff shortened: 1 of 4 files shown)


apelisse:
I haven't read the code yet but thank you for the great explanation of the pull-request. I don't know if this is something we want for OpenAPI v2 since eventually we have to keep the feature enabled but we're hoping to decrease the number of users to 0, which means that the cache would NEVER have to be populated (that's the target goal). Now, we could do what you suggest for OpenAPI v3, but since it's split across group, we also don't know how often the groups are pulled vs refreshed. Do we have any specific data or reason to implement this?

PR #279, opened 2022-02-19
yaml unmarshal for OpenAPIv2 types

alexzielenski:
There is no fast way to deserialize protobuf into kube-openapi types. The obvious choice is to follow this sequence:

`PB` -> `google/gnostic` -> `JSON` -> kube-openapi

However in CLI tools this is slow. On mine and others' systems it takes over half a second to just do the  conversion, let alone the operation the user requested. Too slow for interactivity.

This PR facilitates the intermediate step necessary to perform the following faster conversion:

`PB` -> `google/gnostic` -> `yaml.v3.Node` -> kube-openapi

This method compares favorably to the choice of using JSON as an intermediary.

OpenAPI V2 Conversion Benchmarks:
These benchmarks start with `swagger.json` pulled from a running k8s cluster and measure conversions between various representations of the openapi spec. In the below tests `swagger` refers to `kube-openapi's` `pkg/validation/spec.Swagger`

Ran a benchmark to compare different conversion:
https://github.com/alexzielenski/kube-openapi-gnostic-benchmark

Using JSON:
```
BenchmarkSlowConversion/json->swagger-8         	       2	 557015227 ns/op	95739288 B/op	 1381635 allocs/op
BenchmarkSlowConversion/swagger->json-8         	       8	 128882670 ns/op	71913162 B/op	  274072 allocs/op
BenchmarkSlowConversion/json->gnostic-8         	       5	 216769974 ns/op	81453435 B/op	 1248390 allocs/op
BenchmarkSlowConversion/gnostic->pb-8           	     121	   9842726 ns/op	 2899970 B/op	       1 allocs/op
BenchmarkSlowConversion/pb->gnostic-8           	     100	  15442423 ns/op	 9480707 B/op	  123829 allocs/op
BenchmarkSlowConversion/gnostic->yaml-8         	      42	  28809063 ns/op	32855155 B/op	  264562 allocs/op

BenchmarkSlowConversion/yaml->json-8            	      20	  61439852 ns/op	25325099 B/op	  463786 allocs/op
BenchmarkSlowConversion/json->swagger#01-8      	       2	 524846406 ns/op	95787616 B/op	 1381624 allocs/op
```

Using YAML:
```
BenchmarkFastConversion/json->swagger-8         	       2	 548771326 ns/op	95736376 B/op	 1381630 allocs/op
BenchmarkFastConversion/swagger->json-8         	       8	 147762185 ns/op	67593998 B/op	  274060 allocs/op
BenchmarkFastConversion/json->gnostic-8         	       5	 228517622 ns/op	81446896 B/op	 1248395 allocs/op
BenchmarkFastConversion/gnostic->pb-8           	     100	  10217647 ns/op	 2899968 B/op	       1 allocs/op
BenchmarkFastConversion/pb->gnostic-8           	     100	  12665889 ns/op	 9480707 B/op	  123829 allocs/op
BenchmarkFastConversion/gnostic->yaml-8         	      37	  27832678 ns/op	32855294 B/op	  264562 allocs/op

BenchmarkFastConversion/yaml->swagger-8         	      21	  58512404 ns/op	23687182 B/op	  510369 allocs/op
```

`yaml->swagger` at 58.5ms is 10x faster than `yaml->json` + `json->swagger#01` at 586.2ms

OpenAPI V3 yaml patch is forthcoming in another PR

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -17,8 +17,11 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"reflect"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 )
 
 // [DEPRECATED] ToCanonicalName converts Golang package/type canonical name into REST friendly OpenAPI name.
@@ -107,4 +110,22 @@ func GetCanonicalTypeName(model interface{}) string {
 		path = strings.TrimPrefix(path, "vendor/")
 	}
 	return path + "." + t.Name()
+}
+
+// Provides a fast path for decoding YAML scalar node as a string
+// If the node's value can be simply returned directly, then it is. Otherwise,
+// the yaml.v3.Node.Decode slow path is taken
+func DecodeYAMLString(n *yaml.Node, s *string) error {
+	if n.Kind != yaml.ScalarNode {
+		return fmt.Errorf("expected scalarnode, not %v", n.Kind)
+	}
+
+	if (n.Tag == "!!str" || n.Tag == "tag:yaml.org,2002:!!string") ||
+		(n.Tag == "" || n.Tag == "!") && n.Style&(yaml.SingleQuotedStyle|yaml.DoubleQuotedStyle|yaml.LiteralStyle|yaml.FoldedStyle) != 0 {
+		*s = n.Value
+		return nil
+	}
+
+	// Use slow path if it is not a basic string
+	return n.Decode(&s)
 }
--- a/pkg/validation/spec/contact_info.go
+++ b/pkg/validation/spec/contact_info.go
@@ -18,7 +18,7 @@ package spec
 //
 // For more information: http://goo.gl/8us55a#contactObject
 type ContactInfo struct {
-	Name  string `json:"name,omitempty"`
-	URL   string `json:"url,omitempty"`
-	Email string `json:"email,omitempty"`
+	Name  string `json:"name,omitempty" yaml:"name,omitempty"`
+	URL   string `json:"url,omitempty" yaml:"url,omitempty"`
+	Email string `json:"email,omitempty" yaml:"email,omitempty"`
 }
--- a/pkg/validation/spec/info.go
+++ b/pkg/validation/spec/info.go
@@ -16,9 +16,12 @@ package spec
 
 import (
 	"encoding/json"
+	"errors"
 	"strings"
 
 	"github.com/go-openapi/swag"
+	"gopkg.in/yaml.v3"
+	"k8s.io/kube-openapi/pkg/util"
 )
 
 // Extensions vendor specific extensions
@@ -133,14 +136,43 @@ func (v *VendorExtensible) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (v *VendorExtensible) UnmarshalYAML(value *yaml.Node) error {
+	// var d map[string]interface{}
+	if value.Kind != yaml.MappingNode {
+		return errors.New("expected mappingNode")
+	} else if len(value.Content)%2 != 0 {
+		return errors.New("expected even child nodes")
+	}
+
+	for i := 0; i < len(value.Content); i += 2 {
+		var keyStr string
+		if err := util.DecodeYAMLString(value.Content[i], &keyStr); err != nil {
+			return err
+		}
+
+		if strings.HasPrefix(keyStr, "x-") || strings.HasPrefix(keyStr, "X-") {
+			if v.Extensions == nil {
+				v.Extensions = map[string]interface{}{}
+			}
+
+			var decodedVal interface{}
+			if err := value.Content[i+1].Decode(&decodedVal); err != nil {
+				return err
+			}
+			v.Extensions[strings.ToLower(keyStr)] = decodedVal
+		}
+	}
+	return nil
+}
+
 // InfoProps the properties for an info definition
 type InfoProps struct {
-	Description    string       `json:"description,omitempty"`
-	Title          string       `json:"title,omitempty"`
-	TermsOfService string       `json:"termsOfService,omitempty"`
-	Contact        *ContactInfo `json:"contact,omitempty"`
-	License        *License     `json:"license,omitempty"`
-	Version        string       `json:"version,omitempty"`
+	Description    string       `json:"description,omitempty" yaml:"description,omitempty"`
+	Title          string       `json:"title,omitempty" yaml:"title,omitempty"`
+	TermsOfService string       `json:"termsOfService,omitempty" yaml:"termsOfService,omitempty"`
+	Contact        *ContactInfo `json:"contact,omitempty" yaml:"contact,omitempty"`
+	License        *License     `json:"license,omitempty" yaml:"license,omitempty"`
+	Version        string       `json:"version,omitempty" yaml:"version,omitempty"`
 }
 
 // Info object provides metadata about the API.
@@ -171,4 +203,11 @@ func (i *Info) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	return json.Unmarshal(data, &i.VendorExtensible)
+}
+
+func (i *Info) UnmarshalYAML(value *yaml.Node) error {
+	if err := value.Decode(&i.InfoProps); err != nil {
+		return err
+	}
+	return i.VendorExtensible.UnmarshalYAML(value)
 }
--- a/pkg/validation/spec/swagger.go
+++ b/pkg/validation/spec/swagger.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 
 	"github.com/go-openapi/swag"
+	"gopkg.in/yaml.v3"
+	"k8s.io/kube-openapi/pkg/util"
 )
 
 // Swagger this is the root document object for the API specification.
@@ -30,6 +32,8 @@ type Swagger struct {
 	VendorExtensible
 	SwaggerProps
 }
+
+var _ yaml.Unmarshaler = &Swagger{}
 
 // MarshalJSON marshals this swagger structure to json
 func (s Swagger) MarshalJSON() ([]byte, error) {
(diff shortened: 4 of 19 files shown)


mengqiy:
json->spec.swagger     550 ms
yaml->spec.swagger    58 ms
pd->spec.swagger = PB -> google/gnostic -> yaml.v3.Node -> spec.swagger 99 ms

It seems the fastest way to get a spec.swagger is to deserialize yaml, though the apiserver doesn't serve openapi in yaml format. I guess when using it in a CLI tool, we can convert json or pd to yaml once and deserialize yaml when needed to get faster performance.

```go
yamlNode := &yaml.Node{
	Kind:        yaml.DocumentNode,
	Content:     []*yaml.Node{rawInfo},
	HeadComment: "",
}

var decodedSwagger spec.Swagger
err := yamlNode.Decode(&decodedSwagger)
// check error
```

Users will need to use something like above to convert yaml to `spec.Swagger`.
I guess if you can provide a method like `PraseYAMLDocument(in string) (*spec.Swagger, error)` to wrap the logic above, it would be more accessible to the users.

pkg/util/util.go
```diff
@@ -108,3 +111,21 @@ func GetCanonicalTypeName(model interface{}) string {
 	}
 	return path + "." + t.Name()
 }
+
+// Provides a fast path for decoding YAML scalar node as a string
+// If the node's value can be simply returned directly, then it is. Otherwise,
+// the yaml.v3.Node.Decode slow path is taken
+func DecodeYAMLString(n *yaml.Node, s *string) error {
```
Jefftree, line 118: What's the performance difference with this change? I'm curious if this might be better committed in the YAML library rather than here since there isn't any k8s specific logic.

/cc @apelisse

pkg/validation/spec/info.go
```diff
@@ -133,14 +136,43 @@ func (v *VendorExtensible) UnmarshalJSON(data []byte) error {
...
+		return errors.New("expected even child nodes")
+	}
+
+	for i := 0; i < len(value.Content); i += 2 {
+		var keyStr string
+		if err := util.DecodeYAMLString(value.Content[i], &keyStr); err != nil {
+			return err
+		}
+
+		if strings.HasPrefix(keyStr, "x-") || strings.HasPrefix(keyStr, "X-") {
```
Jefftree, line 153: Is the capitalization check capturing an edge case or do we already have places where `X-` is passed in?

pkg/validation/spec/swagger.go
```diff
@@ -163,6 +196,20 @@ func (s *SchemaOrStringArray) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (s *SchemaOrStringArray) UnmarhsalYAML(value *yaml.Node) error {
```
Jefftree, line 199: s/Unmarhsal/Unmarshal

pkg/validation/spec/swagger_test.go
```diff
@@ -126,3 +186,15 @@ func TestSwaggerSpec_Deserialize(t *testing.T) {
...
+	var actual Swagger
+	var nodes yaml.Node
+	err := yaml.Unmarshal([]byte(specYAML), &nodes)
+	assert.NoError(t, err)
+
+	err = nodes.Decode(&actual)
+	if assert.NoError(t, err) {
+		assert.EqualValues(t, actual, spec)
+	}
+}
```
Jefftree, line 200: Not a comment for this PR, but we should add a roundtrip test in k/k for converting to YAML and back

Jefftree (commented):
Should we also try replacing the swagger -> json -> proto marshalling to use YAML instead? (https://github.com/kubernetes/kube-openapi/blob/master/pkg/handler/handler.go#L109)

pkg/util/util.go
```diff
@@ -108,3 +111,21 @@ func GetCanonicalTypeName(model interface{}) string {
...
+	if n.Kind != yaml.ScalarNode {
+		return fmt.Errorf("expected scalarnode, not %v", n.Kind)
+	}
+
+	if (n.Tag == "!!str" || n.Tag == "tag:yaml.org,2002:!!string") ||
+		(n.Tag == "" || n.Tag == "!") && n.Style&(yaml.SingleQuotedStyle|yaml.DoubleQuotedStyle|yaml.LiteralStyle|yaml.FoldedStyle) != 0 {
```
liggitt, line 124: is this copied from something in the yaml library? I'm not familiar with what this means or if it is correct

pkg/validation/spec/contact_info.go
```diff
@@ -18,7 +18,7 @@ package spec
 //
 // For more information: http://goo.gl/8us55a#contactObject
 type ContactInfo struct {
-	Name  string `json:"name,omitempty"`
-	URL   string `json:"url,omitempty"`
-	Email string `json:"email,omitempty"`
+	Name  string `json:"name,omitempty" yaml:"name,omitempty"`
+	URL   string `json:"url,omitempty" yaml:"url,omitempty"`
+	Email string `json:"email,omitempty" yaml:"email,omitempty"`
```
liggitt, line 23: I _really_ don't think we should start adding yaml tags to our API types. Having benchmarked the yaml decoding in the past, I'm also really surprised it is faster than the json decoding (it was generally significantly slower in the past). Do we know which bits the json decoding is super slow on?

pkg/util/util.go
liggitt: +1 on not putting general-purpose yaml decoding in this lib

pkg/validation/spec/contact_info.go
alexzielenski: Yaml v3 package since 2019 operates on an AST; did you look before then? 

My benchmarks measured unmarshalling from the AST type instead of text. This is sufficient for my use case of providing a path to convert from protobuf: `protoubf -> google/gnostic -> kube-openapi`

I haven't benchmarked it, but I expect going from YAML test -> AST -> Kube-Openapi would be comparable/slower than JSON.

pkg/util/util.go
alexzielenski: I copied it from yaml decoding logic for string path:

https://github.com/go-yaml/yaml/blob/496545a6307b2a7d7a710fd516e5e16e8ab62dbc/yaml.go#L463
https://github.com/go-yaml/yaml/blob/496545a6307b2a7d7a710fd516e5e16e8ab62dbc/decode.go#L565

pkg/validation/spec/info.go
alexzielenski: This extra check for "X-" is here only to mirror the UnmarshalJSON behavior. I think you wrote that ;)

pkg/util/util.go
alexzielenski: Agree that it is not desirable long-term. I will prepare a PR for upstream. However, I do not see an alternative in the short term? It will likely take a while for the change to be merged and appear in a tagged release.

alexzielenski: @Jefftree The difference with this change is significant. Below are benchmark results for openapi v2 with this optimization removed:

```
goos: darwin
goarch: amd64
pkg: github.com/alexzielenski/parsebench
cpu: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
BenchmarkFastConversion
BenchmarkFastConversion/json->swagger
BenchmarkFastConversion/json->swagger-8         	       2	 503133747 ns/op	95739432 B/op	 1381637 allocs/op
BenchmarkFastConversion/swagger->json
BenchmarkFastConversion/swagger->json-8         	       9	 122539288 ns/op	66432463 B/op	  274050 allocs/op
BenchmarkFastConversion/json->gnostic
BenchmarkFastConversion/json->gnostic-8         	       5	 225056310 ns/op	81462545 B/op	 1248402 allocs/op
BenchmarkFastConversion/gnostic->pb
BenchmarkFastConversion/gnostic->pb-8           	     100	  11495301 ns/op	 2899970 B/op	       1 allocs/op
BenchmarkFastConversion/pb->gnostic
BenchmarkFastConversion/pb->gnostic-8           	     100	  13719325 ns/op	 9480698 B/op	  123829 allocs/op
BenchmarkFastConversion/gnostic->yaml
BenchmarkFastConversion/gnostic->yaml-8         	      40	  30027913 ns/op	32855169 B/op	  264562 allocs/op
BenchmarkFastConversion/yaml->swagger
BenchmarkFastConversion/yaml->swagger-8         	      13	  92594648 ns/op	33392264 B/op	  675799 allocs/op
```

93.5ms in this run on my machine. Over a number of runs i saw a range of 80-100ms for yaml->swagger.

This is compared to ~60ms with the optimization enabled.

alexzielenski:
> json->spec.swagger 550 ms yaml->spec.swagger 58 ms pd->spec.swagger = PB -> google/gnostic -> yaml.v3.Node -> spec.swagger 99 ms
> 
> It seems the fastest way to get a spec.swagger is to deserialize yaml, though the apiserver doesn't serve openapi in yaml format. I guess when using it in a CLI tool, we can convert json or pd to yaml once and reuse it by deserializing yaml when needed to get faster performance.
> 
> ```go
> yamlNode := &yaml.Node{
> 	Kind:        yaml.DocumentNode,
> 	Content:     []*yaml.Node{rawInfo},
> 	HeadComment: "",
> }
> 
> var decodedSwagger spec.Swagger
> err := yamlNode.Decode(&decodedSwagger)
> // check error
> ```
> 
> Users will need to use something like above to convert yaml to `spec.Swagger`. I guess if you can provide a method like `PraseYAMLDocument(in string) (*spec.Swagger, error)` to wrap the logic above, it would be more accessible to the users. @alexzielenski WDYT?

Agreed, the stanza to set up the `DocumentNode` is not obvious, and such a function is required. Ill add it in tomorrow :)

pkg/validation/spec/contact_info.go
liggitt: If we want a google/gnostic → kube-openapi path, it would make more sense to build that transformation and round-trip test it to me. Using yaml AST bits is clever, but I don't think we should decorate API types with yaml tags and push consumers in that direction

natasha41575: > I don't think we should decorate API types with yaml tags and push consumers in that direction

I'm curious why you're against having yaml tags here? The proto->gnostic->kube-openapi conversion that this facilitates is a critical performance improvement that kpt and kustomize would like to have as soon as possible.

apelisse: We have an alternative solution that is probably going to be even faster, but it will take a few weeks to implement. This is also a blocker for next code-freeze so we're trying to move fast on that!

natasha41575: Thank you so much for the update!

alexzielenski:
Given consensus that directing users to use YAML is not desirable, I am closing this PR in favor of a better alternative: direct conversion method without using YAML: #283

PR #282, opened 2022-03-09
add vendor extensions to marshal openapi v2 spec.Header

alexzielenski:
I noticed while fuzzing a type converter to gnostic that the `spec.Header` type does not include its `x-` vendor extensions when marshaled to JSON (they were already being properly unmarshaled from JSON). This PR adds that functionality.

### DIFF
--- a/pkg/validation/spec/header.go
+++ b/pkg/validation/spec/header.go
@@ -53,7 +53,11 @@ func (h Header) MarshalJSON() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return swag.ConcatJSON(b1, b2, b3), nil
+	b4, err := json.Marshal(h.VendorExtensible)
+	if err != nil {
+		return nil, err
+	}
+	return swag.ConcatJSON(b1, b2, b3, b4), nil
 }
 
 // UnmarshalJSON unmarshals this header from JSON
--- a/pkg/validation/spec/header_test.go
+++ b/pkg/validation/spec/header_test.go
@@ -16,9 +16,12 @@ package spec
 
 import (
 	"encoding/json"
+	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func float64Ptr(f float64) *float64 {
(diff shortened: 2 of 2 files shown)


pkg/validation/spec/header_test.go
```diff
@@ -88,3 +89,19 @@ func TestIntegrationHeader(t *testing.T) {
...
+// Makes sure that a Header unmarshaled from known good JSON, and one unmarshaled
+// from generated JSON are equivalent.
+func TestHeaderSerialization(t *testing.T) {
+	generatedJSON, err := json.Marshal(header)
+	require.NoError(t, err)
+
+
+	generatedJSONActual := Header{}
+	require.NoError(t, json.Unmarshal(generatedJSON, &generatedJSONActual))
+	require.EqualValues(t, header, generatedJSONActual)
```
liggitt, line 102: can this use reflect.DeepEqual instead? various comparison libraries have disconcerting ways of coercing almost-equal values for comparison (floats/ints, null vs empty lists, etc)

alexzielenski: @liggitt the existing test also uses `EqualValues` from the same library here: https://github.com/kubernetes/kube-openapi/blob/cf3229a1ee8e0089b4f80d3d53ae8ecffb8b9426/pkg/validation/spec/header_test.go#L87 I expect you would like that removed as well?

A nice benefit of using this assertion method is that in the case of failure, a readable diff is shown. I am unsure if reflect has the same functionality. An alternative might be that testify has `require.Equal` as opposed to `require.EqualValues` but its documentation not detailed, and is not explicit about a promise not to coerce types.

liggitt: yeah, reflect.DeepEqual is easier to reason about... if the test passes with that, I'd prefer switching to it. If the test doesn't pass with that, then I guess I'd want to know why and whether we care about the difference that method found

liggitt: a pattern in other unit tests is to check equality with reflect.DeepEqual, then use something like cmp.Diff to print a pretty diff in failure cases

alexzielenski: the test now follows the pattern you suggested

pkg/validation/spec/header.go
```diff
@@ -53,7 +53,11 @@ func (h Header) MarshalJSON() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return swag.ConcatJSON(b1, b2, b3), nil
+	b4, err := json.Marshal(h.VendorExtensible)
+	if err != nil {
+		return nil, err
+	}
```
liggitt, line 59: are we using json marshaling of this type anywhere in kubernetes today (in CRD validation, or openapi v2 generation, etc)? just wanting to make sure this isn't an API-facing change in current use

alexzielenski: Looking at the instances of `Response` through https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json none of the default types define headers for their responses.

In the case of CRDs, it may be possible that CRDs populate response headers for their open api spec. I am wholly unfamiliar with that part of the infrastructure, though. @apelisse do you know whether existing CRDs are able to populate `Schema.Paths.Responses.Headers`?

apelisse: CRDs don't have the ability to do anything specific to their response header, if I understand correctly. I don't think it's being used at all.

Also, I'm not sure what this would break, or how this would be an api-facing change.

liggitt: yeah, I swept k8s.io/kubernetes and couldn't find any places we construct this

> Also, I'm not sure what this would break, or how this would be an api-facing change.

just wanted to know if this was going to start including directives places we hadn't before, and think if any known clients would be unable to handle those (think `kubectl`'s fragile handling of openapi v2)

PR #284, opened 2022-03-10
Cache busting

Jefftree:
Support Cache Busting in kube-openapi.

Also fix a bug where lazy marshaling wasn't performed on OpenAPI V3.

### DIFF
--- a/pkg/handler3/handler.go
+++ b/pkg/handler3/handler.go
@@ -23,7 +23,10 @@ import (
 	"fmt"
 	"mime"
 	"net/http"
+	"net/url"
+	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -49,9 +52,20 @@ const (
 	subTypeJSON     = "json"
 )
 
+// OpenAPIV3Discovery is the format of the Discovery document for OpenAPI V3
+// It maps Discovery paths to their corresponding URLs with a hash parameter included
+type OpenAPIV3Discovery struct {
+	Paths map[string]OpenAPIV3DiscoveryGroupVersion
+}
+
+// OpenAPIV3DiscoveryGroupVersion includes information about a group version and URL
+// for accessing the OpenAPI. The URL includes a hash parameter to support client side caching
+type OpenAPIV3DiscoveryGroupVersion struct {
+	URL string
+}
+
 // OpenAPIService is the service responsible for serving OpenAPI spec. It has
 // the ability to safely change the spec while serving it.
-// OpenAPI V3 currently does not use the lazy marshaling strategy that OpenAPI V2 is using
 type OpenAPIService struct {
 	// rwMutex protects All members of this service.
 	rwMutex      sync.RWMutex
@@ -66,6 +80,7 @@ type OpenAPIV3Group struct {
 
 	pbCache   handler.HandlerCache
 	jsonCache handler.HandlerCache
+	etagCache handler.HandlerCache
 }
 
 func init() {
@@ -75,7 +90,18 @@ func init() {
 }
 
 func computeETag(data []byte) string {
-	return fmt.Sprintf("\"%X\"", sha512.Sum512(data))
+	if data == nil {
+		return ""
+	}
+	return fmt.Sprintf("%X", sha512.Sum512(data))
+}
+
+func constructURL(gvString, etag string) string {
+	u := url.URL{Path: path.Join("/openapi/v3", gvString)}
+	query := url.Values{}
+	query.Set("hash", etag)
+	u.RawQuery = query.Encode()
+	return u.String()
 }
 
 // NewOpenAPIService builds an OpenAPIService starting with the given spec.
@@ -96,10 +122,17 @@ func (o *OpenAPIService) getGroupBytes() ([]byte, error) {
 	}
 
 	sort.Strings(keys)
-	group := make(map[string][]string)
-	group["Paths"] = keys
-
-	j, err := json.Marshal(group)
+	discovery := &OpenAPIV3Discovery{Paths: make(map[string]OpenAPIV3DiscoveryGroupVersion)}
+	for gvString, groupVersion := range o.v3Schema {
+		etagBytes, err := groupVersion.etagCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		discovery.Paths[gvString] = OpenAPIV3DiscoveryGroupVersion{
+			URL: constructURL(gvString, string(etagBytes)),
+		}
+	}
+	j, err := json.Marshal(discovery)
 	if err != nil {
 		return nil, err
 	}
@@ -114,11 +147,19 @@ func (o *OpenAPIService) getSingleGroupBytes(getType string, group string) ([]by
 		return nil, "", time.Now(), fmt.Errorf("Cannot find CRD group %s", group)
 	}
 	if getType == subTypeJSON {
-		specBytes, etag, err := v.jsonCache.Get()
-		return specBytes, etag, v.lastModified, err
+		specBytes, err := v.jsonCache.Get()
+		if err != nil {
+			return nil, "", v.lastModified, err
+		}
+		etagBytes, err := v.etagCache.Get()
+		return specBytes, string(etagBytes), v.lastModified, err
 	} else if getType == subTypeProtobuf {
-		specPb, etag, err := v.pbCache.Get()
-		return specPb, etag, v.lastModified, err
+		specPb, err := v.pbCache.Get()
+		if err != nil {
+			return nil, "", v.lastModified, err
+		}
+		etagBytes, err := v.etagCache.Get()
+		return specPb, string(etagBytes), v.lastModified, err
 	}
 	return nil, "", time.Now(), fmt.Errorf("Invalid accept clause %s", getType)
 }
@@ -127,15 +168,10 @@ func (o *OpenAPIService) UpdateGroupVersion(group string, openapi *spec3.OpenAPI
 	o.rwMutex.Lock()
 	defer o.rwMutex.Unlock()
 
-	specBytes, err := json.Marshal(openapi)
-	if err != nil {
-		return err
-	}
-
 	if _, ok := o.v3Schema[group]; !ok {
 		o.v3Schema[group] = &OpenAPIV3Group{}
 	}
-	return o.v3Schema[group].UpdateSpec(specBytes)
+	return o.v3Schema[group].UpdateSpec(openapi)
 }
 
 func (o *OpenAPIService) DeleteGroupVersion(group string) {
@@ -192,7 +228,26 @@ func (o *OpenAPIService) HandleGroupVersion(w http.ResponseWriter, r *http.Reque
 			if err != nil {
 				return
 			}
-			w.Header().Set("Etag", etag)
+			// ETag must be enclosed in double quotes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
+			w.Header().Set("Etag", strconv.Quote(etag))
+
+			if hash := r.URL.Query().Get("hash"); hash != "" {
+				if hash != etag {
+					u := constructURL(group, etag)
+					http.Redirect(w, r, u, 301)
+					return
+				}
+				// The Vary header is required because the Accept header can
+				// change the contents returned. This prevents clients from caching
+				// protobuf as JSON and vice versa.
+				w.Header().Set("Vary", "Accept")
+
+				// Only set these headers when a hash is given.
+				w.Header().Set("Cache-Control", "public, immutable")
+				// Set the Expires directive to the maximum value of one year from the request,
+				// effectively indicating that the cache never expires.
+				w.Header().Set("Expires", time.Now().AddDate(1, 0, 0).Format(time.RFC1123))
+			}
 			http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
 			return
 		}
@@ -207,16 +262,28 @@ func (o *OpenAPIService) RegisterOpenAPIV3VersionedService(servePath string, han
 	return nil
 }
 
-func (o *OpenAPIV3Group) UpdateSpec(specBytes []byte) (err error) {
+func (o *OpenAPIV3Group) UpdateSpec(openapi *spec3.OpenAPI) (err error) {
 	o.rwMutex.Lock()
 	defer o.rwMutex.Unlock()
 
-	o.pbCache = o.pbCache.New(func() ([]byte, error) {
-		return ToV3ProtoBinary(specBytes)
-	})
-
 	o.jsonCache = o.jsonCache.New(func() ([]byte, error) {
-		return specBytes, nil
+		return json.Marshal(openapi)
+	})
+	o.pbCache = o.pbCache.New(func() ([]byte, error) {
+		json, err := o.jsonCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		return ToV3ProtoBinary(json)
+	})
+	// TODO: This forces a json marshal of corresponding group-versions.
+	// We should look to replace this with a faster hashing mechanism.
+	o.etagCache = o.etagCache.New(func() ([]byte, error) {
+		json, err := o.jsonCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		return []byte(computeETag(json)), nil
 	})
 	o.lastModified = time.Now()
 	return nil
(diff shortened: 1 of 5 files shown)


pkg/handler3/handler.go
```diff
@@ -96,10 +107,15 @@ func (o *OpenAPIService) getGroupBytes() ([]byte, error) {
...
-	group["Paths"] = keys
-
-	j, err := json.Marshal(group)
+	discovery := &OpenAPIV3Discovery{Paths: make(map[string]string)}
+	for k, v := range o.v3Schema {
+		_, etag, err := v.jsonCache.Get()
+		if err != nil {
+			return nil, err
+		}
+		discovery.Paths[k] = "openapi/v3/" + k + "?hash=" + unquoteETag(etag)
```
alexzielenski, line 116: While consuming this API in a test I found it a bit unwieldy. I expected `discovery.Paths` to be a `groupVersion -> ETAG` map, rather than a `groupVersion -> path` map. I ended up using `url.Parse` to extract the `hash` query param from the path since the `client-go` discovery client for OpenAPI V3 [asks for the `hash` by itself](https://github.com/Jefftree/kubernetes/pull/3/files#diff-14f541e7356b7bc7f735b24db888b48b81b81e136b5af2925e912b95889dd010R476). 

Also it is worth pointing out that in the future more query params may be added to the OpenAPI endpoints, so it may become inconvenient to deal with these paths with an assumed URL structure.

pkg/handler3/handler.go
```diff
@@ -193,6 +226,8 @@ func (o *OpenAPIService) HandleGroupVersion(w http.ResponseWriter, r *http.Reque
 				return
 			}
 			w.Header().Set("Etag", etag)
+
+			ApplyCacheBusting(w, r, etag)
 			http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
```
alexzielenski, line 231: In the case that `ApplyCacheBusting` calls `http.Redirect`, does this `http.ServeContent` conflict with the response?

pkg/handler3/handler.go
```diff
@@ -49,9 +49,12 @@ const (
 	subTypeJSON     = "json"
 )
 
+type OpenAPIV3Discovery struct {
```
alexzielenski, line 52: needs documentation on type and fields

pkg/handler3/handler.go
```diff
@@ -74,6 +77,14 @@ func init() {
 	mime.AddExtensionType(".gz", mimePbGz)
 }
 
+// unquoteEtag removes the string quotes around an ETag
+func unquoteETag(etag string) string {
```
alexzielenski, line 81: This function has similar semantics to [`strconv.Unquote`](https://pkg.go.dev/strconv#Unquote). Should that be used instead?

pkg/handler3/handler.go
alexzielenski: We may also in the future end up wanting to extend the discovery map to include additional metadata. Should the type of `discovery.Paths[k]` be a `map[string]interface{}` rather than simply a `string`? So that for example the response might be:

```json
{
    "paths": {
        "batch/v1": {
            "hash": <ETAG>
        }
    }
}
```

pkg/handler3/handler.go
apelisse: I'd even suggest we don't prematurely quote the etag so we don't have to untag it later?

pkg/handler3/handler.go
```diff
@@ -117,7 +133,11 @@ func (o *OpenAPIService) getSingleGroupBytes(getType string, group string) ([]by
 		specBytes, etag, err := v.jsonCache.Get()
 		return specBytes, etag, v.lastModified, err
 	} else if getType == subTypeProtobuf {
-		specPb, etag, err := v.pbCache.Get()
+		_, etag, err := v.jsonCache.Get()
+		if err != nil {
+			return nil, "", v.lastModified, err
+		}
+		specPb, _, err := v.pbCache.Get()
```
apelisse, line 140: I think we're doing the wrong thing with the lazy caching base-code. If we don't like how the etag is computed (since clearly we don't want one per type anymore), we should change the underlying lazy-caching code.

pkg/handler3/handler.go
apelisse: I don't disagree with that, but it should be a URL rather than the hash so it's built automatically (clients shouldn't even have to know what to do with the hash, it should just contain the URL they need to download as a relative link).

pkg/handler3/handler.go
apelisse: Note that this probably also applies to other things below.

pkg/handler3/handler.go
apelisse: > While consuming this API in a test I found it a bit unwieldy. I expected discovery.Paths to be a groupVersion -> ETAG map, rather than a groupVersion -> path map. I ended up using url.Parse to extract the hash query param from the path since the client-go discovery client for OpenAPI V3 [asks for the hash by itself](https://github.com/Jefftree/kubernetes/pull/3/files#diff-14f541e7356b7bc7f735b24db888b48b81b81e136b5af2925e912b95889dd010R476).

Interesting, why didn't you just want to download the relative path given?

Jefftree: Thanks for catching this inconsistency. 

>We may also in the future end up wanting to extend the discovery map to include additional metadata.

Good idea, I'll change the structure.

>  I ended up using url.Parse to extract the hash query param from the path since the client-go discovery client

Question for both of you: Would it make more sense for the client-go interface to directly consume the URL (eg: `OpenAPISchema(URL string)`) instead of the current `OpenAPISchema(gvPath, hash string)`. I agree we shouldn't assume an URL structure and going the route of parsing the URL may add more complexity in the future

alexzielenski: > I don't disagree with that, but it should be a URL rather than the hash so it's built automatically (clients shouldn't even have to know what to do with the hash, it should just contain the URL they need to download as a relative link).

In that case a URL could be included as one of the keys in the map I proposed

pkg/handler3/handler.go
Jefftree: I'll update the etag so it's not prematurely quoted and use strconv.Quote when we need to

pkg/handler3/handler.go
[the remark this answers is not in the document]
Jefftree: Added a helper function for constructing the URL and used the url and path libraries. I'm still not sure if it's the expected way of constructing the URL, PTAL when you get a chance.

pkg/handler3/handler.go
Jefftree: I sort of fixed this by allowing the user to pass in an optional ETag in the New function when initializing the cache. If the ETag is empty then we'll compute it. Let me know your thoughts.

pkg/handler3/handler.go
Jefftree: Changed the Discovery schema format to

```
{
    "paths": {
        "apis/batch/v1": {
            "url":  "/openapi/v3/apis/batch/v1?hash=XXX"
        }
    }
}
```

apelisse: Yeah, I don't disagree that having an object rather than a direct mapping gives us more flexibility in the future, but we shouldn't need the hash per-se.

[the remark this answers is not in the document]
Jefftree: Updated per discussion offline.

pkg/handler3/handler.go
Jefftree: Created a new cache for ETag and decoupled from byteCache.

pkg/handler3/handler.go
```diff
@@ -157,6 +193,27 @@ func (o *OpenAPIService) HandleDiscovery(w http.ResponseWriter, r *http.Request)
 	http.ServeContent(w, r, "/openapi/v3", time.Now(), bytes.NewReader(data))
 }
 
+func CacheBustingShouldRedirect(r *http.Request, etag string) bool {
```
alexzielenski, line 196: This function is only used once. I'm not sure keeping it separate it contributes much to code hygiene/readability

alexzielenski: Could be written as one-line:

```go
if hash := r.URL.Query().Get("hash); hash != "" && hash != etag {
    u := constructURL(group, etag)
	http.Redirect(w, r, u, 301)
	return
}
```

pkg/handler3/handler.go
```diff
@@ -192,7 +249,14 @@ func (o *OpenAPIService) HandleGroupVersion(w http.ResponseWriter, r *http.Reque
 			if err != nil {
 				return
 			}
-			w.Header().Set("Etag", etag)
+			w.Header().Set("Etag", strconv.Quote(etag))
```
alexzielenski, line 252: Why does the ETag HTTP header need to be quoted? Might be worth a `//` comment.

alexzielenski (commented):
Looks clean. Very minor nits

pkg/handler3/handler.go
Jefftree: Fixed

pkg/handler3/handler.go
Jefftree: Added a reference to the MDN page although I'm not sure how much value that provides since it is a standard.

pkg/handler3/handler_test.go
```diff
@@ -120,3 +122,153 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
+		if resp.StatusCode != 200 {
+			t.Errorf("Accept: Unexpected response status code, want: %v, got: %v", 200, resp.StatusCode)
+		}
+
+		if cacheControl := resp.Header.Get("Cache-Control"); cacheControl != tc.cacheControl {
+			t.Errorf("Expected Cache Control %v, got %v", tc.cacheControl, cacheControl)
+		}
+
+		if tc.expectedHash != "" {
+			if hash := resp.Request.URL.Query()["hash"]; hash[0] != tc.expectedHash {
```
apelisse, line 240: What is this actually testing?

pkg/handler3/handler.go
```diff
@@ -207,16 +266,26 @@ func (o *OpenAPIService) RegisterOpenAPIV3VersionedService(servePath string, han
...
 }
 
-func (o *OpenAPIV3Group) UpdateSpec(specBytes []byte) (err error) {
+func (o *OpenAPIV3Group) UpdateSpec(openapi *spec3.OpenAPI) (err error) {
 	o.rwMutex.Lock()
 	defer o.rwMutex.Unlock()
 
 	o.pbCache = o.pbCache.New(func() ([]byte, error) {
-		return ToV3ProtoBinary(specBytes)
+		json, err := o.jsonCache.Get()
```
apelisse, line 274: Do you know for sure what's going to be captured in that closure? `o` or `o.jsonCache`? Because if it's the latter, `o.jsonCache` is made obsolete line 280 below, so that's fishy. Can we topologically sort these instantiations just to be safe (and the intent actually looks a little better).

pkg/handler3/handler.go
```diff
@@ -96,10 +122,17 @@ func (o *OpenAPIService) getGroupBytes() ([]byte, error) {
 	}
 
 	sort.Strings(keys)
-	group := make(map[string][]string)
-	group["Paths"] = keys
-
-	j, err := json.Marshal(group)
+	discovery := &OpenAPIV3Discovery{Paths: make(map[string]OpenAPIV3DiscoveryGroupVersion)}
+	for gvString, groupVersion := range o.v3Schema {
+		etagBytes, err := groupVersion.etagCache.Get()
```
apelisse, line 127: Maybe add a TODO here (or inside the etagCache), this is forcing a json marshal of all group/versions, which is unfortunately going to be slow and we would like to replace that with a faster hashing mechanism.

pkg/handler3/handler.go
```diff
@@ -157,6 +193,21 @@ func (o *OpenAPIService) HandleDiscovery(w http.ResponseWriter, r *http.Request)
 	http.ServeContent(w, r, "/openapi/v3", time.Now(), bytes.NewReader(data))
 }
 
+func applyCacheBusting(w http.ResponseWriter, r *http.Request, etag string) {
+	hash := r.URL.Query().Get("hash")
+	// Only apply cache busting on requests where the hash is provided.
+	if hash != "" {
+		// The Vary header is required because the Accept header can change the contents returned. This prevents clients from caching protobuf as JSON and vice versa.
+		w.Header().Set("Vary", "*")
```
apelisse, line 201: `Vary: *` literally disables all caching AFAIK, so that's not going to work ;-), you need `Vary: Accept` and maybe a few other.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary#directives:
> ```
> *
>     Indicates that factors other than request headers influenced the generation of this response. Implies that the response is uncacheable.
```

File: pkg/handler3/handler.go [hunk 12]
```diff
@@ -192,7 +243,15 @@ func (o *OpenAPIService) HandleGroupVersion(w http.ResponseWriter, r *http.Reque
...
-			w.Header().Set("Etag", etag)
+			// ETag must be enclosed in double quotes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
+			w.Header().Set("Etag", strconv.Quote(etag))
+
+			if hash := r.URL.Query().Get("hash"); hash != "" && hash != etag {
+				u := constructURL(group, etag)
+				http.Redirect(w, r, u, 301)
+				return
+			}
+			applyCacheBusting(w, r, etag)
```
apelisse, line 254: Piling up on Alex's comment, a lot of the conditions here are very similar to the conditions in `applyCacheBusting`. Try to remove the `applyCacheBusting` method below and move the code here, I think you'll be able to do a few other simplifications.

apelisse: Since I don't code enough these days, I'll do it for fun. Not addressing the `Vary` comment.
```suggestion
			if hash := r.URL.Query().Get("hash"); hash != "" {
				// The Vary header is required because the Accept header can 
				// change the contents returned. This prevents clients from caching 
				// protobuf as JSON and vice versa.
				// Do you need Vary in the 301?!
				w.Header().Set("Vary", "*")
				if hash != etag {
					u := constructURL(group, etag)
					http.Redirect(w, r, u, 301)
					return
				}
				// Only set these headers when a hash is given.
				w.Header().Set("Cache-Control", "public, immutable")
				// Set the Expires directive to the maximum value of one year from the request,
				// effectively indicating that the cache never expires.
				w.Header().Set("Expires", time.Now().AddDate(1, 0, 0).Format(time.RFC1123))
			}
```

apelisse: So we're not redirecting if no hash is given?

pkg/handler3/handler_test.go
```diff
@@ -120,3 +122,153 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
+		// Incorrect hash should redirect to the page with the correct hash
+		{"application/json",
+			200,
+			"openapi/v3/apis/apps/v1?hash=" + "OUTDATEDHASH",
+			returnedJSON,
+			hash,
+			"public, immutable",
+		},
+		{"application/com.github.proto-openapi.spec.v3@v1.0+protobuf",
+			200,
```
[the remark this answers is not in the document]
Jefftree: The response comes from after the redirect/retry.

pkg/handler3/handler_test.go
apelisse: I see, so this is testing to see if the redirect worked well?

Jefftree: Checks if the hash parameter returned in the response matches the latest hash. If the client supplied the correct hash, this hash should remain the same. If the client supplied the incorrect hash, this should return the hash of the latest document.

pkg/handler3/handler.go
apelisse: I think you could have split the type docs into type docs and per field docs, but you don't have to change that.

apelisse: Do we need the json tags on the fields? (here and below?)

pkg/handler3/handler.go
```diff
@@ -192,7 +243,27 @@ func (o *OpenAPIService) HandleGroupVersion(w http.ResponseWriter, r *http.Reque
...
+				// protobuf as JSON and vice versa.
+				w.Header().Set("Vary", "Accept")
+
+				// Only set these headers when a hash is given.
+				w.Header().Set("Cache-Control", "public, immutable")
+				// Set the Expires directive to the maximum value of one year from the request,
+				// effectively indicating that the cache never expires.
+				w.Header().Set("Expires", time.Now().AddDate(1, 0, 0).Format(time.RFC1123))
+			}
+			applyCacheBusting(w, r, etag)
```
apelisse, line 266: you shouldn't call that function anymore (and you can remove it)

PR #285, opened 2022-03-10
Change references of googleapis/gnostic -> google/gnostic

Jefftree:
See https://github.com/kubernetes/kubernetes/pull/108644

k/k and k/kube-openapi need to change googleapis/gnostic -> google/gnostic at the same time to avoid type check failure.

Test failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/108644/pull-kubernetes-integration/1502033895208521728

Change references of googleapis/gnostic -> google/gnostic

Also upgrade gnostic from 0.5.5 to 0.6.6. No significant changes for openapi_v2 and openapi_v3,

- https://github.com/google/gnostic/commits/master/openapiv2
- https://github.com/google/gnostic/commits/master/openapiv3

since April 26, 2021 @ version v0.5.5

- PR 250 updates the .md file
- PR 259 Renames googleapis -> google
- PR 270 updates the protoc version

/assign @apelisse


apelisse:
OK Thanks,
/lgtm
/approve

mattrobenolt:
I think this breaks dependencies like `k8s.io/client-go` too.

I'm not entirely sure, since this is a bit over my head in how Go resolves dependencies.

But it seems that because the old version `googleapis/gnostic` is just a redirect to the newer version `google/gnostic`, I think explicitly in the dependency resolver, it doesn't treat them the same.

So what's happening is I'm getting an `indirect` dependency for both `google/gnostic v0.6.6` and `googleapis/gnostic v0.5.5` from `client-go`. I think this confuses the compiler or somewhere in that it tries to alias the `googleapis/gnostic` version _as_ `google/gnostic` and they are incompatible because different versions? I'm not exactly sure.

But the exact compilation error when updating deps now is:

```
/go/pkg/mod/k8s.io/client-go@v0.23.4/applyconfigurations/meta/v1/unstructured.go:64:38: cannot use doc (variable of type *"github.com/googleapis/gnostic/openapiv2".Document) as *"github.com/google/gnostic/openapiv2".Document value in argument to proto.NewOpenAPIData (compile)
```

I'm not exactly sure how to work around this or what the resolution is.

mattrobenolt:
I guess I worked around it by pinning my indirect dependency to a sha before this one, and it resolved it.

I wasn't able to reproduce it on a new project either, so I'm not sure the perfect storm that leads to this being a problem.

Jefftree:
@mattrobenolt Yes, we are aware there will be a version mismatch with kubernetes/kubernetes (and subsequently the client-go applyconfigurations) until https://github.com/kubernetes/kubernetes/pull/108644 is merged.

liggitt:
> I think this breaks dependencies like k8s.io/client-go too.

it shouldn't, as long as you use the level of kube-openapi referenced by client-go. did you run `go get -u` to bump all dependencies to their latest versions? client-go has many v0.x.y dependencies that will not work with

PR #286, opened 2022-03-11
Trim leading newlines in Descriptions

Jefftree:
Trim leading newlines in Descriptions. Improves the formatting when Enum types have no comments attached and any case where there are extra newlines at the beginning of descriptions.

Changes `"\n\nDescription: \"Possible enum values:...` to `"Description: \"Possible enum values:...`

/cc @apelisse 
/cc @jiahuif

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -603,7 +603,8 @@ func (g openAPITypeWriter) generateDescription(CommentLines []string) {
 		}
 	}
 
-	postDoc := strings.TrimRight(buffer.String(), "\n")
+	postDoc := strings.TrimLeft(buffer.String(), "\n")
+	postDoc = strings.TrimRight(postDoc, "\n")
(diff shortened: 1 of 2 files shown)


apelisse:
Can you please lgtm @jiahuif

pkg/generators/openapi.go
```diff
@@ -603,7 +603,8 @@ func (g openAPITypeWriter) generateDescription(CommentLines []string) {
 		}
 	}
 
-	postDoc := strings.TrimRight(buffer.String(), "\n")
+	postDoc := strings.TrimLeft(buffer.String(), "\n")
```
jiahuif, line 606: Note: After we get go 1.18 trims will be alloc-free and faster. ref: https://tip.golang.org/doc/go1.18

jiahuif (commented):
/lgtm

PR #288, opened 2022-03-16
Update gnostic to drop jsonschema dependency

Jefftree:
The short term fix is to use this specific gnostic version which includes our OpenAPI V3 bug fix and does not pull in additional dependencies.

Gnostic will be split into two repos in the future, with a dedicated library for generated models to avoid pulling in unnecessary dependencies.

/assign @apelisse 
/assign @liggitt


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/go-openapi/jsonreference v0.19.3
 	github.com/go-openapi/swag v0.19.5
 	github.com/golang/protobuf v1.5.2
-	github.com/google/gnostic v0.6.6
+	github.com/google/gnostic v0.5.7-v3refs
```
liggitt, line 14: this isn't a semantic version, which means `go mod tidy` wasn't run on this, right?

Jefftree: I did run `go mod tidy` on this, which removed the jsonschema dependencies. The version name is valid per semver https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions (pre-release).

liggitt: huh... I didn't realize go preserved that... that's fun

liggitt:
want to take a stab at the corresponding k8s.io/kubernetes bump based on this before merge? you can see what it will look like with `hack/pin-dependency.sh k8s.io/kube-openapi=github.com/Jefftree/kube-openapi gnostic-fix`

Jefftree:
Yep, I have it in https://github.com/kubernetes/kubernetes/pull/108644. The dependencies look correct and jsonschema and its tree of dependencies were removed. (github.com/jefftree/kube-openapi v0.0.8-gnostic corresponds to this branch)

liggitt:
great, that looks way better. this will let us make progress, but we'll still want to resolve https://github.com/google/gnostic/issues/317 asap, since we're effectively blocked from picking up gnostic bugfixes at this point

/lgtm

apelisse:
Agreed, that's exciting, thanks @Jefftree !
/approve

PR #289, opened 2022-03-18
add support for custom v3 oneOf types

jiahuif:
observations:
- in Kubernetes type definitions, only `oneOf` makes sense. `anyOf` or `allOf` does not [1].
- it makes no sense to have v3 `oneOf` but not v2 `type` or `format`

[1] https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/


test/integration/pkg/generated/openapi_generated.go
```diff
@@ -17,7 +17,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by openapi-gen. DO NOT EDIT.
+// Code generated by openapi-gen.go. DO NOT EDIT.
```
jiahuif, line 20: This should be a side effect of upgrading golang version.

jiahuif: never mind, reverting this

pkg/generators/openapi.go
```diff
@@ -380,6 +413,9 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 				"},\n"+
 				"}\n}\n\n", args)
 			return nil
+		case hasV3OneOfTypes:
+			// having v3 oneOf types without custom v2 type or format does not make sense.
```
apelisse, line 417: Why not?

jiahuif: Because if a type needs custom type definition, it at least should contain a "format" field. We require a type provide both "format" and "type" even if the type only wants one of them.

jiahuif: And thus, the structure can either (1) do not use any type customization (2) provide all of v2 type definition and v3 oneOf

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -315,7 +336,7 @@ func schema_test_integration_testdata_enumtype_FruitsBasket(ref common.Reference
 				Properties: map[string]spec.Schema{
 					"content": {
 						SchemaProps: spec.SchemaProps{
-							Description: "\n\n\nPossible enum values:\n - `\"apple\"` is the Apple\n - `\"banana\"` is the Banana\n - `\"onigiri\"` is the Rice ball that does not seem to belong to a fruits basket but has a long comment that is so long that it spans multiple lines",
+							Description: "Possible enum values:\n - `\"apple\"` is the Apple\n - `\"banana\"` is the Banana\n - `\"onigiri\"` is the Rice ball that does not seem to belong to a fruits basket but has a long comment that is so long that it spans multiple lines",
```
apelisse, line 339: How can we have this here?

test/integration/testdata/custom/v3.go
```diff
@@ -77,3 +77,17 @@ func (_ Bah) OpenAPISchemaType() []string {
 func (_ Bah) OpenAPISchemaFormat() string {
 	return "test-format"
 }
+
+// FooV3OneOf has an OpenAPIV3OneOfTypes method
+// +k8s:openapi-gen=true
+type FooV3OneOf struct{}
+
+func (FooV3OneOf) OpenAPIV3OneOfTypes() []string {
+	return []string{"number", "String"}
```
apelisse, line 86: nit, but maybe `string` lowercase?

jiahuif: Thanks!

test/integration/pkg/generated/openapi_generated.go
apelisse: I understand that, my concern is that the test should have failed in #286 but didn't

jiahuif: The reason is that if the integration tests fail, the expected results get updated on disk instead of reporting as an error, which is bad.

PR #290, opened 2022-03-21
do not include v2 schema when serving v3

jiahuif:
This PR adds pruning v2 schema from a v3 one

### DIFF
--- a/pkg/builder3/openapi_test.go
+++ b/pkg/builder3/openapi_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/emicklei/go-restful"
 	"github.com/stretchr/testify/assert"
+
 	openapi "k8s.io/kube-openapi/pkg/common"
 	"k8s.io/kube-openapi/pkg/spec3"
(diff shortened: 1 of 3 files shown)


pkg/builder3/openapi_test.go
```diff
@@ -90,10 +91,11 @@ func (_ TestInput) OpenAPIDefinition() *openapi.OpenAPIDefinition {
 		},
 	}
 	schema.Extensions = spec.Extensions{"x-test": "test"}
-	return &openapi.OpenAPIDefinition{
+	def := openapi.EmbedOpenAPIDefinitionIntoV2Extension(openapi.OpenAPIDefinition{
 		Schema:       schema,
 		Dependencies: []string{},
-	}
+	}, openapi.OpenAPIDefinition{})
```
apelisse, line 97: If I'm reading correctly, this would embed an empty v2extension in the test schema, and since the test is unaffected, it proves that the schema is dropped, is that correct?

jiahuif: Yes, I am adding comments to clarify this.

PR #292, opened 2022-03-22
Add OpenAPI V2 to V3 Converter

Jefftree:
- Use the CommonResponses object in OpenAPI V3 builder. (This adds the 404 Response structs)
- Create a OpenAPIV3Config struct that we can transition to in the future. Currently it is just an upsampling from the v2 Config.
- Add converter from OpenAPI V2 to V3 
- Verify converter by generating the OpenAPI v2/v3 from go structs, and attempt to convert v2 to v3 and check for equality.


test/integration/testutil/testutil.go
```diff
@@ -66,8 +66,10 @@ func CreateWebServices() []*restful.WebService {
 	addRoutes(w, buildRouteForType(w, "uniontype", "InlinedUnion")...)
 	addRoutes(w, buildRouteForType(w, "custom", "Bal")...)
 	addRoutes(w, buildRouteForType(w, "custom", "Bak")...)
-	addRoutes(w, buildRouteForType(w, "custom", "Bac")...)
-	addRoutes(w, buildRouteForType(w, "custom", "Bah")...)
+	if includeV2SchemaAnnotation {
```
Jefftree, line 69: All other tests use true for the new flag. OpenAPI v2->v3 conversion tests have this disabled because they include the x-kubernetes-v2-schema extension and there is no way to provide equality with the generated v3 schema and upsampled v2->v3.

alexzielenski: I don't understand your comment here, or what is special about the "Bac" and "Bah" routes that they would need to be excluded for new conversion tests. Could you provide some more context?

Also, I am told by @apelisse the `x-kubernetes-v2-schema` extension is being removed so maybe this does not matter any longer?

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+	klog "k8s.io/klog/v2"
+	"k8s.io/kube-openapi/pkg/spec3"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+var OpenAPIV2DefPrefix = "#/definitions/"
+var OpenAPIV3DefPrefix = "#/components/schemas/"
+
+func ConvertV2ToV3(v2Spec *spec.Swagger) *spec3.OpenAPI {
+	v3Spec := &spec3.OpenAPI{
```
alexzielenski, line 31: This conversion is not complete. For instance, `Servers` is missing. Are we explicitly excluding it due to domain knowledge that k8s does not make use of this field?

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+		if len(v2Schema.Items.Schemas) > 0 {
+			v3Schema.Items.Schemas = []spec.Schema{}
+			for _, s := range v2Schema.Items.Schemas {
+				v3Schema.Items.Schemas = append(v3Schema.Items.Schemas, *ConvertSchema(&s))
+			}
+		}
+	}
+
+	if v2Schema.AdditionalProperties != nil {
+		v3Schema.AdditionalProperties.Schema = ConvertSchema(v2Schema.AdditionalProperties.Schema)
```
alexzielenski, line 103: What is the test coverage for this function? I am not seeing where `v3Schema.AdditoinalProperties` is initialized.

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+	return securityScheme
+}
+
+func ConvertPaths(v2Paths *spec.Paths) *spec3.Paths {
+	if v2Paths == nil {
+		return nil
+	}
+	paths := &spec3.Paths{
+		VendorExtensible: v2Paths.VendorExtensible,
+		Paths:            make(map[string]*spec3.Path),
```
alexzielenski, line 145: `Paths` should be non-nil if and only if `v2Paths.Paths` is non-nil. This same reasoning also applies elsewhere: if `spec` lists the fields as `nil`, `spec3` should also use `nil`. If `spec` is an empty map/list, `spec3` should be an empty map/list.

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+			OperationId:  v2Operation.ID,
+			Parameters:   []*spec3.Parameter{},
+		},
+	}
+
+	for _, param := range v2Operation.Parameters {
+		if param.ParamProps.Name == "body" && param.ParamProps.Schema != nil {
+			operation.OperationProps.RequestBody = &spec3.RequestBody{
+				RequestBodyProps: spec3.RequestBodyProps{
+					Content: make(map[string]*spec3.MediaType),
```
alexzielenski, line 196: Does this content map ever get populated?

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+			ExternalDocs: ConvertExternalDocumentation(v2Operation.OperationProps.ExternalDocs),
+			Tags:         v2Operation.Tags,
+			Summary:      v2Operation.Summary,
+			Deprecated:   v2Operation.Deprecated,
+			OperationId:  v2Operation.ID,
+			Parameters:   []*spec3.Parameter{},
+		},
+	}
+
+	for _, param := range v2Operation.Parameters {
```
alexzielenski, line 192: `Parameter` is `Refable` but I do not see this case being handled.

pkg/common/common.go
```diff
@@ -117,6 +119,89 @@ type Config struct {
 	DefaultSecurity []map[string][]string
 }
 
+type OpenAPIV3Config struct {
```
Jefftree, line 122: Descriptions need to be updated

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+	}
+
+	if v2securityScheme.Flow != "" {
+		securityScheme.Flows = make(map[string]*spec3.OAuthFlow)
+		securityScheme.Flows[v2securityScheme.Flow] = &spec3.OAuthFlow{
+			OAuthFlowProps: spec3.OAuthFlowProps{
+				AuthorizationUrl: v2securityScheme.AuthorizationURL,
+				TokenUrl:         v2securityScheme.TokenURL,
+				Scopes:           v2securityScheme.Scopes,
```
alexzielenski, line 132: This shares a reference between the two instances on `Scopes`. Is this conversion intended to create a copy? If not, I think it should be documented that references will be shared. There are likely other places  in this conversion where references are being shared.

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+	if refString := v2Schema.Ref.String(); refString != "" {
+		if idx := strings.Index(refString, OpenAPIV2DefPrefix); idx != -1 {
+			v3Schema.Ref = spec.MustCreateRef(OpenAPIV3DefPrefix + refString[idx+len(OpenAPIV2DefPrefix):])
+		} else {
+			klog.Errorf("Error: Swagger V2 Ref %s does not contain #/definitions\n", refString)
+		}
+	}
+
+	for key, property := range v2Schema.Properties {
+		v3Schema.Properties[key] = *ConvertSchema(&property)
```
alexzielenski, line 90: Do we want to change `v3Schema.Properties` to be a `map[string]*Schema` instead of `map[string]Schema` to prevent copying? I am not sure how smart Go is about copy-on-write; especially when consumers of this map are fetching keys.

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+		ExternalDocumentationProps: spec3.ExternalDocumentationProps{
+			Description: v2ED.Description,
+			URL:         v2ED.URL,
+		},
+	}
+}
+
+func ConvertComponents(v2SecurityDefinitions spec.SecurityDefinitions, v2Definitions spec.Definitions, v2Responses map[string]spec.Response, produces []string) *spec3.Components {
+	components := &spec3.Components{
+		Schemas:         make(map[string]*spec.Schema),
```
alexzielenski, line 56: These instances should only be non-nil if their v2 counterparts are non-nil

test/integration/openapiconv/convert_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	builderv2 "k8s.io/kube-openapi/pkg/builder"
+	builderv3 "k8s.io/kube-openapi/pkg/builder3"
+	"k8s.io/kube-openapi/pkg/openapiconv"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"k8s.io/kube-openapi/test/integration/pkg/generated"
+	"k8s.io/kube-openapi/test/integration/testutil"
+)
+
+func TestConvertGolden(t *testing.T) {
```
alexzielenski, line 32: Can we also add fuzz tests?

pkg/builder3/openapi.go
```diff
@@ -154,18 +165,23 @@ func (o *openAPI) buildRequestBody(parameters []common.Parameter, bodySample int
 
 func newOpenAPI(config *common.Config) openAPI {
 	o := openAPI{
-		config: config,
+		config: common.ConvertConfigToV3(config),
```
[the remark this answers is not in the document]
Jefftree: No, we should have separate configs for OpenAPI V2 and V3. This should apply for structs as well (spec.Info, Schema, etc). 

I think making that change would be too big given our current timeline, so as a compromise I'm starting to create the V3 config but generating that from the common Config.

pkg/builder3/openapi.go
```diff
@@ -154,18 +165,23 @@ func (o *openAPI) buildRequestBody(parameters []common.Parameter, bodySample int
...
 			Components: &spec3.Components{
-				Schemas: map[string]*spec.Schema{},
+				Schemas:         map[string]*spec.Schema{},
+				SecuritySchemes: make(spec3.SecuritySchemes),
+				Responses: make(map[string]*spec3.Response),
 			},
 		},
 	}
+	for k, response := range o.config.ResponseDefinitions {
+		o.spec.Components.Responses[k] = response
```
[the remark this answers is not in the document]
Jefftree: You mean v2.Schema to v3.Schema right?

pkg/openapiconv/convert.go
Jefftree: Are you suggesting for this PR or in the future? Currently both v2 and v3 share the same Schema

pkg/builder3/openapi.go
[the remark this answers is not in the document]
alexzielenski: I mistakenly mentioned schema, but I do mean we will have to convert v2.Response to v3.Response

pkg/openapiconv/convert.go
alexzielenski: Not for this PR, was asking for future. I take your point this schema type is from the v2 openAPI. IIRC the v3 schema already corrects for this

pkg/openapiconv/convert.go
Jefftree: Added documentation. I know we probably shouldn't be sharing the references but this function is intended to be used only by the aggregator via json Unmarshal -> v2 -> v3 -> json Marshal so it may be fine to keep as is. Let me know if there's a preference

pkg/openapiconv/convert.go
Jefftree: Fixed.

pkg/openapiconv/convert.go
Jefftree: The for loop right below populates the content map.

test/integration/testutil/testutil.go
Jefftree:
```
   "custom.Bah": {
     "type": "object",
     "x-kubernetes-v2-schema": {
      "type": "test-type",
      "format": "test-format"
     }
    }
```

OpenAPI V2 will interpret this with "test-type" and OpenAPI V3 will interpret this as "object". Converting OpenAPI V2 to V3 will not magically change the type and will cause a type mismatch in our comparison. This mismatch is intended because V2 to V3 is a lossy conversion but will cause the tests to fail.

pkg/openapiconv/convert.go
Jefftree: Hm, is there a significance between nil vs empty maps? They would json marshal to the same result but I guess their representation is different in golang. I'm sure there are a bunch of other instances of this in the code, is there a reason we'd prefer one over the other?

pkg/openapiconv/convert.go
Jefftree: I'll add a note for this, there are a couple of fields that were excluded because they're not populated by the OpenAPI v2 builder

pkg/builder3/openapi.go
```diff
@@ -102,7 +104,16 @@ func (o *openAPI) buildOperations(route common.Route, inPathCommonParamsMap map[
 		}
 	}
 
-	// TODO: Default response if needed. Common Response config
+	for code, resp := range o.config.CommonResponses {
+		if _, exists := ret.Responses.StatusCodeResponses[code]; !exists {
+			ret.Responses.StatusCodeResponses[code] = resp
```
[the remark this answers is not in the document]
Jefftree: I believe so, basing the logic off of https://github.com/kubernetes/kube-openapi/blob/master/pkg/builder/openapi.go#L323-L331

pkg/builder3/openapi.go
```diff
@@ -102,7 +104,16 @@ func (o *openAPI) buildOperations(route common.Route, inPathCommonParamsMap map[
...
+		}
+	}
+
+	// If there is still no response, use default response provided.
+	if len(ret.Responses.StatusCodeResponses) == 0 {
```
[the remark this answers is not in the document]
Jefftree: hm, I was basing this code off the v2 generator, but you're right we should probably populate it regardless of whether there are other codes.

pkg/builder3/openapi.go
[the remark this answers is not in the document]
Jefftree: The conversion happens in the Config conversion

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,270 @@
...
+			klog.Errorf("Error: Swagger V2 Ref %s does not contain #/definitions\n", refString)
+		}
+	}
+
+	for key, property := range v2Schema.Properties {
+		v3Schema.Properties[key] = *ConvertSchema(&property)
+	}
+	if v2Schema.Items != nil {
+		v3Schema.Items.Schema = ConvertSchema(v2Schema.Items.Schema)
+		if len(v2Schema.Items.Schemas) > 0 {
```
apelisse, line 94: Same comment as in other places, should this be `if v2Schema.Items.Schemas != nil` so that we keep that nil if it is.

pkg/openapiconv/convert.go
apelisse: Are we 100% sure that they serialize to the same thing? There's clearly a difference between a missing map and an empty map (s/map/list/)

pkg/builder3/openapi.go
```diff
@@ -34,11 +34,12 @@ const (
 )
 
 type openAPI struct {
-	config      *common.Config
+	config      *common.OpenAPIV3Config
```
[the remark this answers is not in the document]
Jefftree: The V2 config belongs in common, any reason the V3 Config should live elsewhere?

pkg/openapiconv/convert.go
Jefftree: They serialize to the same thing but reflect.DeepEqual will fail the comparison. I've added length checks to avoid initializing 0 length maps/lists.

test/integration/openapiconv/convert_test.go
```diff
@@ -0,0 +1,63 @@
...
+		log.Fatalf("ERROR: %s", serr.Error())
+	}
+
+	openapiv3, serr := builderv3.BuildOpenAPISpec(testutil.CreateWebServices(false), config)
+	if serr != nil {
+		log.Fatalf("ERROR: %s", serr.Error())
+	}
+
+	convertedOpenAPIV3 := openapiconv.ConvertV2ToV3(openapiv2)
+	if !reflect.DeepEqual(openapiv3, convertedOpenAPIV3) {
```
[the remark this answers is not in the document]
Jefftree: Added testdata tests for the full k8s schema for apiextensions, batch, and api.

pkg/openapiconv/convert.go
Jefftree: We don't modify an array by taking the length, so if it's a nil array, the length will return 0 and no change will be applied. The array will still be nil.

pkg/builder3/openapi.go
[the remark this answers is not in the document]
alexzielenski: It is not a no-op to populate this field, since its meaning is that if a response code is returned by the API which is not part of `StatusCodeResponses`, the user of the API can assume the response body conforms to the `Default` schema. 

It is important that the spec reflects that actual behavior of the API. In this case, I don't have enough information to know what is the real behavior of the API.  My take is that `ret.Responses.Default` should only ever be non-nil if you can provide that guarantee.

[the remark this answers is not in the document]
Jefftree: Thanks for the explanation, fixed.

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,300 @@
...
+				v3Schema.Items.Schemas = append(v3Schema.Items.Schemas, *ConvertSchema(&s))
+			}
+		}
+	}
+
+	if v2Schema.AdditionalProperties != nil {
+		v3Schema.AdditionalProperties.Schema = ConvertSchema(v2Schema.AdditionalProperties.Schema)
+	}
+	if v2Schema.AdditionalItems != nil {
+		v3Schema.AdditionalProperties.Schema = ConvertSchema(v2Schema.AdditionalProperties.Schema)
```
apelisse, line 113: AdditionalItems or AdditionalProperties?

pkg/openapiconv/convert.go
apelisse: v3Schema is copied from v2Schema L86, so if it's not nil for v2, it's not nil for v3. But shouldn't it also be copied?

test/integration/openapiconv/convert_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	builderv2 "k8s.io/kube-openapi/pkg/builder"
+	builderv3 "k8s.io/kube-openapi/pkg/builder3"
+	"k8s.io/kube-openapi/pkg/openapiconv"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"k8s.io/kube-openapi/test/integration/pkg/generated"
+	"k8s.io/kube-openapi/test/integration/testutil"
+)
+
+func TestConvertGolden(t *testing.T) {
```
apelisse, line 32: Can we update this test (and the other tests) to make sure that the conversion function DOES NOT alter the openapi v2?

pkg/openapiconv/convert.go
Jefftree: Thanks for the critical catch. I went back through the generated files to see why the tests didn't catch this, it turns out every instance of this in the k/k schema had additionalitems == additionalproperties.

Jefftree:
/retest
hm, the tests all pass locally

test/integration/openapiconv/convert_test.go
Jefftree: The conversion function does alter the openapi. I want to point out that the aggregator would download the OpenAPI V2 json, unmarshal, runs the V2->V3 converter, and then directly marshal the V3 to be served. The intermediary OpenAPI v2 spec is not used for anything else.

To ensure that the conversion function does not alter the openapi v2, we need to deepcopy the openapi v2. That would result in a marshal/unmarshal in the converter. The aggregator would then marshal/unmarshal the openapi v2 multiple times to preserve our unaltered openapi v2 guarantee.

One way I can think of doing this is changing the function signature to accept the marshalled swagger instead of the swagger itself. That will save us on a json round trip if we want to guarantee that v2 schema is unaltered.

apelisse: That's awesome thanks

pkg/openapiconv/convert.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if v2Schema.Items != nil {
+		v3Schema.Items = new(spec.SchemaOrArray)
+		v3Schema.Items.Schema = ConvertSchema(v2Schema.Items.Schema)
+		if v2Schema.Items.Schemas != nil {
+			v3Schema.Items.Schemas = []spec.Schema{}
+			for _, s := range v2Schema.Items.Schemas {
+				v3Schema.Items.Schemas = append(v3Schema.Items.Schemas, *ConvertSchema(&s))
+			}
+		}
```
apelisse, line 115: Not something I specifically want you to change, but you could have written this like:
```
v3Schema.Items := spec.SchemaOrArray {
    Schema: ConvertSchema(v2schema.Items.Schema),
    Schemas: ConvertSchemaList(v2.schema.Items.Schemas), // Move the for loop into its own function
}
```

Jefftree: You could have just said you want me to change it :)

Updated.

apelisse (commented):
OK Thanks, I'm happy with it. @sttts Do you want to take a look?

apelisse:
Feel free to squash I gues?

PR #293, opened 2022-03-22
Implement proto.Models for OpenAPI v3

alexzielenski:
This PR adds the ability to parse an gnostic `openapi_v3.Document` into types conforming to the interface required by kubectl to power the `explain` feature.

This is a temporary measure as we prepare to replace usage of the gnostic types with the Kube-OpenAPI's `spec3` types in a future release.


pkg/util/proto/openapi_test.go
```diff
@@ -28,6 +28,7 @@ import (
 
 var fakeSchema = testing.Fake{Path: filepath.Join("testdata", "swagger.json")}
 var fakeSchemaNext = testing.Fake{Path: filepath.Join("testdata", "swagger_next.json")}
+var fakeSchemaV300 = testing.Fake{Path: filepath.Join("testdata", "openapi_v3_0_0")}
```
apelisse, line 31: 🤖 😎

pkg/util/proto/document_v3.go
```diff
@@ -0,0 +1,329 @@
...
+
+	components := doc.GetComponents()
+	if components == nil {
+		return &definitions, nil
+	}
+
+	schemas := components.GetSchemas()
+	if schemas == nil {
+		return &definitions, nil
+	}
```
apelisse, line 43: Maybe I'm wrong, but if gnostic still works the same, it usually works well when you call methods on nil pointers. So you can shorten that to:
```suggestion
	schemas := components.GetComponents().GetSchemas()
	if schemas == nil {
		return &definitions, nil
	}
```

pkg/util/proto/testing/openapi.go
```diff
@@ -52,6 +56,35 @@ func (f *Fake) OpenAPISchema() (*openapi_v2.Document, error) {
 	return f.document, f.err
 }
 
+func (f *Fake) OpenAPIV3Discovery() (*handler3.OpenAPIV3Discovery, error) {
```
apelisse, line 59: I noticed that you haven't used this, how do you plan on using this?

pkg/util/proto/testing/openapi.go
```diff
@@ -52,6 +56,35 @@ func (f *Fake) OpenAPISchema() (*openapi_v2.Document, error) {
...
+	_, err := os.Stat(f.Path)
+	if err != nil {
+		return nil, err
+	}
+	spec, err := ioutil.ReadFile(filepath.Join(f.Path, groupVersion+".json"))
+	if err != nil {
+		return nil, err
+	}
+	return openapi_v3.ParseDocument(spec)
+}
```
apelisse, line 86: This one doesn't cache the result, which is sort of the promise of the `Fake` object, maybe we need a `FakeV3`?

pkg/util/proto/document_v3.go
```diff
@@ -0,0 +1,329 @@
...
+	"reflect"
+	"strings"
+
+	openapi_v3 "github.com/google/gnostic/openapiv3"
+	"gopkg.in/yaml.v3"
+)
+
+// Temporary parse implementation to be used until gnostic->kube-openapi conversion
+// is possible.
+func NewOpenAPIV3Data(doc *openapi_v3.Document) (Models, error) {
```
apelisse, line 30: I'm not excited to see more of this, but hopefully we can get rid of it altogether soon.

pkg/util/proto/testing/openapi.go
alexzielenski: I added caching to this method of the `Fake`.

On a slightly related note,  I did end up adding the following mock class to `client-go` which I found more useful in my testing:

https://github.com/alexzielenski/kubernetes/blob/cache-busting/staging/src/k8s.io/client-go/util/testing/fake_openapi_handler.go#L40

Maybe it is more appropriate to add to kube-openapi, but I did not use it there. I use this mock server so far in tests for `client-go`, `kubectl`, and `kube-aggregator`

pkg/util/proto/testing/openapi.go
```diff
@@ -32,6 +34,10 @@ type Fake struct {
 	once     sync.Once
 	document *openapi_v2.Document
 	err      error
+
+	v3DocumentsLock sync.Mutex
```
apelisse, line 38: You have members of that struct that only make sense for v3 and members that only make sense for v2. Can you split into two different structs?

alexzielenski: yeah fair enough

apelisse:
Thanks
/lgtm
/approve

pkg/util/proto/document_v3.go
liubog2008: Does this conversion is ready now ?

PR #295, opened 2022-03-23
Fix json encoding for OpenAPI v3 Discovery

Jefftree:
Address some comments from https://github.com/kubernetes/kubernetes/pull/108895#discussion_r833561682

### DIFF
--- a/pkg/handler3/handler.go
+++ b/pkg/handler3/handler.go
@@ -55,13 +55,14 @@ const (
 // OpenAPIV3Discovery is the format of the Discovery document for OpenAPI V3
 // It maps Discovery paths to their corresponding URLs with a hash parameter included
 type OpenAPIV3Discovery struct {
-	Paths map[string]OpenAPIV3DiscoveryGroupVersion
+	Paths map[string]OpenAPIV3DiscoveryGroupVersion `json:"paths"`
 }
 
 // OpenAPIV3DiscoveryGroupVersion includes information about a group version and URL
 // for accessing the OpenAPI. The URL includes a hash parameter to support client side caching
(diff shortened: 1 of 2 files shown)


pkg/handler3/handler.go
```diff
@@ -55,13 +55,14 @@ const (
...
 type OpenAPIV3Discovery struct {
-	Paths map[string]OpenAPIV3DiscoveryGroupVersion
+	Paths map[string]OpenAPIV3DiscoveryGroupVersion `json:"paths"`
 }
 
 // OpenAPIV3DiscoveryGroupVersion includes information about a group version and URL
 // for accessing the OpenAPI. The URL includes a hash parameter to support client side caching
 type OpenAPIV3DiscoveryGroupVersion struct {
-	URL string
+	// Path is an absolute path of an OpenAPI V3 document in the form of /openapi/v3/<group-version-string>?hash=<hash>
```
liggitt, line 64: spell out the group version, and make it an example, not a promised format:
```suggestion
	// Path is an absolute path of an OpenAPI V3 document. For example: /openapi/v3/apps/v1?hash=014fbff9a07c
```

liggitt: if it's going to optionally contain a query string, I'm not sure what the best name for this is. on the other hand, it is an item inside a list named `paths`... so....

apelisse: Could it be a `RelativeURL`?

liggitt: `ServerRelativeURL` (json `serverRelativeURL`) , I guess, since we're starting with `/`, not relative to /openapi/v3/

that could work as well

Jefftree: Updated.

PR #297, opened 2022-03-24
Return valid field in Validation struct

### LINKED ISSUE #296: Inconsistent in error `Value`
"kube-openapi is super inconsistent about what it places in Value for these errors... in some cases, it's the value, in some cases, it's the actual length, in some cases, it's the max length. The unit tests will reveal if propagating value into the field error makes for a sensible or weird error... " from https://github.com/kubernetes/kubernetes/pull/108859#discussion_r834585755

cici37:
Due to inconsistent in returned err value field, add `valid` for Kubernetes to better handle err.
TooFewProperties/TooManyProperties didn't align well with other err but it might make sense since we don't want return large object back especially when it is rarely used.

Close #296

### DIFF
--- a/pkg/validation/errors/schema.go
+++ b/pkg/validation/errors/schema.go
@@ -156,31 +156,33 @@ func PropertyNotAllowed(name, in, key string) *Validation {
 }
 
 // TooFewProperties an error for an object with too few properties
-func TooFewProperties(name, in string, n int64) *Validation {
-	msg := fmt.Sprintf(tooFewProperties, name, in, n)
+func TooFewProperties(name, in string, minProperties int64) *Validation {
+	msg := fmt.Sprintf(tooFewProperties, name, in, minProperties)
 	if in == "" {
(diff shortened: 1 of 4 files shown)


pkg/validation/errors/schema.go
```diff
@@ -166,6 +166,7 @@ func TooFewProperties(name, in string, n int64) *Validation {
 		Name:    name,
 		In:      in,
 		Value:   n,
+		Valid:   n,
```
liggitt, line 169: change the name of n to make it clear what it should be (minProperties), and make sure callers are setting it as expected

pkg/validation/errors/schema.go
```diff
@@ -181,6 +182,7 @@ func TooManyProperties(name, in string, n int64) *Validation {
...
```
liggitt, line 185: change the name of n to make it clear what it should be (maxProperties), and make sure callers are setting it as expected

PR #298, opened 2022-03-25
Wrap Refs with AllOf

Jefftree:
Refs should not have sibling elements other than description: https://github.com/kubernetes/kubernetes/issues/106387. Wrap Refs in AllOf to preserve gnostic round trip.

### DIFF
--- a/pkg/builder3/openapi.go
+++ b/pkg/builder3/openapi.go
@@ -20,12 +20,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 
 	restful "github.com/emicklei/go-restful"
 
 	"k8s.io/kube-openapi/pkg/common"
 	"k8s.io/kube-openapi/pkg/common/restfuladapter"
+	"k8s.io/kube-openapi/pkg/schemamutation"
 	"k8s.io/kube-openapi/pkg/spec3"
 	"k8s.io/kube-openapi/pkg/util"
 	"k8s.io/kube-openapi/pkg/validation/spec"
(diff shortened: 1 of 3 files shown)


pkg/builder3/openapi.go
```diff
@@ -436,3 +438,30 @@ func (o *openAPI) toSchema(name string) (_ *spec.Schema, err error) {
...
+	walker := schemamutation.Walker{
+		SchemaCallback: func(schema *spec.Schema) *spec.Schema {
+			orig := schema
+			clone := func() {
+				if orig == schema {
+					schema = new(spec.Schema)
+					*schema = *orig
+				}
+			}
+			if schema.Ref.String() != "" && (schema.Nullable != false || schema.Default != nil || len(schema.Extensions) > 0) {
```
Jefftree, line 455: I'm questioning if we have a better way to do these checks.

apelisse: you could check if, when removing the ref, you have an empty schema. Or you could copy the ref to an empty schema and compare them? I'm assuming that what you're actually checking is that there is no other field along with the ref, your condition will need to be updated anytime we add such a new thing in our openapi.

pkg/builder3/openapi.go
```diff
@@ -436,3 +438,30 @@ func (o *openAPI) toSchema(name string) (_ *spec.Schema, err error) {
...
+				clone()
+				newSchema := new(spec.Schema)
+				newSchema.Ref = schema.Ref
+				schema.Ref = spec.Ref{}
+				schema.AllOf = []spec.Schema{*newSchema}
```
apelisse, line 460: nit, maybe call the new schema "refSchema" for clarity?

apelisse (commented):
I'm mostly concerned about the condition for creating the ref, but looks good beside that.

PR #299, opened 2022-03-28
Optimize GetOpenAPIDefinitions to avoid generating the map for all definitions on every BuildOpenAPISpec invocation

Jefftree:
Optimize GetOpenAPIDefinitions to avoid generating the map for all definitions on every BuildOpenAPISpec invocation.

Use a pre-existing generated Definitions object if possible to avoid going through the generator process to generate the map of all Definitions.

Significantly reduces memory allocation footprint.

Please only look at latest commit. Branches off of https://github.com/kubernetes/kube-openapi/pull/292

This PR only optimizes for V3, but we can apply the optimization later for V2 as well. 

/assign @apelisse

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -94,6 +94,10 @@ type Config struct {
 	// or any of the models will result in spec generation failure.
 	GetDefinitions GetOpenAPIDefinitions
 
+	// Provides the definition for all models used by routes. One of GetDefinitions or Definitions must be defined to generate a spec.
+	// This takes precedent over the GetDefinitions function
+	Definitions map[string]OpenAPIDefinition
+
 	// GetOperationIDAndTags returns operation id and tags for a restful route. It is an optional function to customize operation IDs.
 	//
 	// Deprecated: GetOperationIDAndTagsFromRoute should be used instead. This cannot be specified if using the new Route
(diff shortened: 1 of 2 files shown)


pkg/common/common.go
```diff
@@ -176,12 +185,13 @@ func ConvertConfigToV3(config *Config) *OpenAPIV3Config {
 		GetOperationIDAndTags:          config.GetOperationIDAndTags,
 		GetOperationIDAndTagsFromRoute: config.GetOperationIDAndTagsFromRoute,
 		GetDefinitionName:              config.GetDefinitionName,
+		Definitions:                    config.Definitions,
```
apelisse, line 188: I must be mis-reading, but what sets that definitions? I can only see it read, I can't find where it's written.

apelisse:
cc @jiahuif @alexzielenski @seans3 

How much of an improvement is this?

pkg/common/common.go
Jefftree: k/k config setup

apelisse: Ah, yeah, right of course

Jefftree:
Reduces memory allocs by the GetDefinitions function from ~5mb x 60 group versions = ~300mb to 5mb (shared struct across all gvs). In use memory is unaffected after the openapi is built.

apelisse:
LGTM for the second commit.

PR #301, opened 2022-04-25
Rename gnostic to gnostic-models

Jefftree:
gnostic-models (https://github.com/google/gnostic-models) is a condensed version of gnostic with the openapi models portion extracted. The dependency graph is much smaller than gnostic.


Jefftree:
> Do you think this will help solve issues such as [kubernetes/client-go#1075 (comment)](https://github.com/kubernetes/client-go/issues/1075#issuecomment-1088820330) ?

Replied in the issue

Jefftree:
cc @apelisse @liggitt for visibility.

https://github.com/kubernetes/kubernetes/pull/109922 is the corresponding k/k diff

PR #304, opened 2022-06-11
Switch to released version of v3.8.0 - github.com/emicklei/go-restful/v3

dims:
Tested in https://github.com/kubernetes/kubernetes/pull/110518, let's pick a released tag which was recently cut.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


austince:
I think this could have been accomplished without breaking changes with the semi-recently added interfaces, as suggested in past PRs (see https://github.com/kubernetes/kube-openapi/pull/271#issuecomment-1007530040). Would anyone be open to talking about proper versioning for the library?

PR #307, opened 2022-06-21
Add required job that depends on matrix jobs

apelisse:


apelisse:
@BenTheElder That hack should fix it, shouldn't it? :D

BenTheElder:
let's find out!

I haven't used github actions for required tests yet, only niche informing signal
/lgtm
/approve

PR #308, opened 2022-06-22
add ability to override structType with field-level annotation

alexzielenski:
Requires https://github.com/kubernetes-sigs/structured-merge-diff/pull/220 to be merged as a prerequisite

Part of solution to https://github.com/kubernetes/kubernetes/pull/110495

Implements functionality so that extensions on struct fields are converted over to the SMD `TypeRef`. This makes it se putting `+structType` on a field is no longer a no-op. The field level structType would take precedence over any `structType` declaration on the referred struct


alexzielenski:
@apelisse test added

apelisse:
Looks good, thanks
/lgtm
/approve

PR #310, opened 2022-08-01
update to smd v5

alexzielenski:
required before updating k8s dep tree

also includes removal of orphaned/unused private function so dep on `github.com/davecgh/go-spew` is now just indirect


alexzielenski:
major version bump no longer necessary

PR #317, opened 2022-09-14
Add openapi server to serve from files

apelisse:
Creates a "fake" openapi server that serves OpenAPI v2 and v3 loaded from files rather than from an actual apiserver.
There are plans to load files from an actual apiserver (as a proxy), and also have hard-coded patches to the openapi.

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+
+	"k8s.io/kube-openapi/pkg/handler"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+func main() {
+	swagger := spec.Swagger{}
+	if len(os.Args) != 2 {
+		fmt.Println("usage: %v <path_to_openapi_v2>", os.Args[0])
+		os.Exit(1)
+	}
+	content, err := os.ReadFile(os.Args[1])
+	if err != nil {
+		panic(fmt.Errorf("Failed to load file: %v", err))
+	}
+	err = json.Unmarshal(content, &swagger)
+	if err != nil {
+		panic(fmt.Errorf("Failed to parse json: %v", err))
+	}
+	o, err := handler.NewOpenAPIService(&swagger)
+	if err != nil {
+		panic(fmt.Errorf("Failed to create OpenAPI service: %v", err))
+	}
+	mux := http.NewServeMux()
+	err = o.RegisterOpenAPIVersionedService("/openapi/v2", mux)
+	if err != nil {
+		panic(fmt.Errorf("Failed to register OpenAPI service: %v", err))
+	}
+	fmt.Println("Listening on :8080")
+	panic(http.ListenAndServe(":8080", mux))
+}
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -6,33 +6,54 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/spf13/cobra"
+	"k8s.io/kube-openapi/pkg/common"
 	"k8s.io/kube-openapi/pkg/handler"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
-func main() {
+func serve_openapi_v2(v2_path string, h common.PathHandler) error {
 	swagger := spec.Swagger{}
-	if len(os.Args) != 2 {
-		fmt.Println("usage: %v <path_to_openapi_v2>", os.Args[0])
-		os.Exit(1)
-	}
-	content, err := os.ReadFile(os.Args[1])
+	content, err := os.ReadFile(v2_path)
 	if err != nil {
-		panic(fmt.Errorf("Failed to load file: %v", err))
+		return fmt.Errorf("Failed to load file: %v", err)
 	}
 	err = json.Unmarshal(content, &swagger)
 	if err != nil {
-		panic(fmt.Errorf("Failed to parse json: %v", err))
+		return fmt.Errorf("Failed to parse json: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


apelisse:
cc @Jefftree @alexzielenski

cmd/proxy/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if version, ok := t["swagger"]; ok {
+		if version == "2.0" {
+			return true, nil
+		}
+		return false, fmt.Errorf(`invalid "swagger" version: %v`, version)
+	}
+	return false, errors.New(`invalid openapi/swagger, no version included`)
+}
+
+func parse_and_serve_files(mux common.PathHandler, args ...string) error {
```
alexzielenski, line 124: what_is_this_naming_convention_why_are_we_larping_as_rust_devs_who_do_you_think_you_are?

cmd/proxy/main.go
```diff
@@ -0,0 +1,173 @@
...
+
+func parse_v2(swagger_json []byte) (*spec.Swagger, error) {
+	swagger := spec.Swagger{}
+	if err := json.Unmarshal(swagger_json, &swagger); err != nil {
+		return nil, fmt.Errorf("failed to parse json: %v", err)
+	}
+	return &swagger, nil
+}
+
+type XKubernetesGroupVersion struct {
```
alexzielenski, line 53: Does this type exist because we cant import k8s types directly from kube-openapi?

cmd/proxy/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if err := json.Unmarshal(openapi_json, &openapi); err != nil {
+		return XKubernetesGroupVersion{}, spec3.OpenAPI{}, fmt.Errorf("failed to parse json: %v", err)
+	}
+	// Find the groupVersion.
+	if openapi.Paths == nil {
+		return XKubernetesGroupVersion{}, spec3.OpenAPI{}, fmt.Errorf("empty OpenAPI paths")
+	}
+	var gv *XKubernetesGroupVersion = nil
+	for path := range openapi.Paths.Paths {
+		matches := resourcePathRe.FindStringSubmatch(path)
```
alexzielenski, line 87: why use regex instead of directly searching for a specific path?

cmd/proxy/main.go
```diff
@@ -0,0 +1,173 @@
...
+	return nil
+}
+
+func serve_v3(v3 map[XKubernetesGroupVersion]spec3.OpenAPI, h common.PathHandler) error {
+	o, err := handler3.NewOpenAPIService(nil)
+	if err != nil {
+		return fmt.Errorf("failed to create OpenAPI v3 service: %v", err)
+	}
+
+	h.Handle("/openapi/v3", http.HandlerFunc(o.HandleDiscovery))
```
alexzielenski, line 36: whats wrong with https://github.com/kubernetes/kube-openapi/blob/352e363eb79bf637a9a7aa5e1db7c67215fd0414/pkg/handler3/handler.go#L260-L264

alexzielenski (commented):
I have a few questions but otherwise very clean

cmd/proxy/main.go
apelisse: it's uses a "handler" interface that implements the `HandlePrefix` function, which exists in k8s, but not in kube-openapi. I don't really need/want to import something more complicated than the old `http/net` mux.

cmd/proxy/main.go
apelisse: For the first occurence, we need the regex, after that we could be searching for the path, sure.

cmd/proxy/main.go
apelisse: I don't think anything currently parses the XKubernetesGroupVersion directly from the openapi extension? What type did you have in mind?

cmd/proxy/main.go
apelisse: https://www.rust-lang.org/

cmd/proxy/main.go
alexzielenski: Not really sure how this type does anything special to parse the openapi exnteion. It seems like a copy of https://github.com/kubernetes/kubernetes/blob/a9593d634c6a053848413e600dadbf974627515f/staging/src/k8s.io/apimachinery/pkg/runtime/schema/group_version.go#L166-L183 except with a `Path()` method.

but you wouldn't be able to use that due to import cycle

apelisse: Yeah, it's mostly the same thing, it's also only a few lines of code, whatever?

cmd/proxy/main.go
```diff
@@ -0,0 +1,173 @@
...
+			return XKubernetesGroupVersion{}, spec3.OpenAPI{}, fmt.Errorf("Multiple GV operations: %v/%v", *gv, newGV)
+		}
+	}
+	if gv == nil {
+		return XKubernetesGroupVersion{}, spec3.OpenAPI{}, fmt.Errorf("no group-version detected in OpenAPI")
+	}
+	return *gv, openapi, nil
+}
+
+func is_v2(content []byte) (bool, error) {
```
Jefftree, line 104: Seems to be a very loose check for the version. Maybe we can do what we did in the tests and run a validator? https://github.com/kubernetes/kube-openapi/blob/3a31a646d5158c26e77bebb8285a6f3194cda7e2/test/integration/builder3/main.go#L68-L75

cmd/proxy/main.go
Jefftree: > Does this type exist because we cant import k8s types directly from kube-openapi?

Is this a bigger issue that we should look to resolve? That's the same reason we can't use the HandlePrefix in this repo.

cmd/proxy/main.go
apelisse: To be fair also, I'm not planning on keeping this code super long :-)

PR #318, opened 2022-09-23
Add a go.mod for integration test

dims:
To prevent leaking test only dependencies into the root go.mod file

gets rid of 3 things from go.mod/sum

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


dims:
see graph below on how we end up with `getkin/kin-openapi` in kubernetes/kubernetes

image

lavalamp:
Nice change!

/lgtm
/approve

apelisse:
Nice, thanks!

PR #320, opened 2022-09-27
fix json unmarshal bug when extension used with Responses

alexzielenski:
discovered while implementing https://github.com/kubernetes/kube-openapi/pull/319

spec allows vendor extensions, but if they are used the parsing would fail and leave both `Default` and `StatusCodeResponses` empty

code now decodes each response individually so vendor extensions can be filtered out

### DIFF
--- a/pkg/validation/spec/responses.go
+++ b/pkg/validation/spec/responses.go
@@ -96,7 +96,9 @@ func (r *ResponsesProps) UnmarshalJSON(data []byte) error {
 	}
 	if v, ok := res["default"]; ok {
 		value := Response{}
-		json.Unmarshal(v, &value)
+		if err := json.Unmarshal(v, &value); err != nil {
+			return err
+		}
 		r.Default = &value
 		delete(res, "default")
 	}
--- a/pkg/validation/spec/responses.go
+++ b/pkg/validation/spec/responses.go
@@ -92,7 +92,7 @@ func (r ResponsesProps) MarshalJSON() ([]byte, error) {
 func (r *ResponsesProps) UnmarshalJSON(data []byte) error {
 	var res map[string]json.RawMessage
 	if err := json.Unmarshal(data, &res); err != nil {
-		return nil
+		return err
 	}
 	if v, ok := res["default"]; ok {
 		value := Response{}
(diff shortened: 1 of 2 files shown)


pkg/validation/spec/responses.go
```diff
@@ -90,20 +90,28 @@ func (r ResponsesProps) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON unmarshals responses from JSON
 func (r *ResponsesProps) UnmarshalJSON(data []byte) error {
-	var res map[string]Response
+	var res map[string]json.RawMessage
 	if err := json.Unmarshal(data, &res); err != nil {
 		return nil
```
apelisse, line 95: why do we return nil here?

apelisse: Also, I've tried to switch that to `return err`, but it still doesn't fail even when it should. Are there other occurences of errors not being propagated properly?

apelisse:
/lgtm
/approve

Looks good, I left one question but it's only slightly related.

pkg/validation/spec/responses.go
alexzielenski: > but it still doesn't fail even when it should. 

I'm not sure what you've tried, but it should fail if provided with a non-map element i.e. `["a", "b", "c"]`, or `123`.

you are right that this should not return nil

pkg/validation/spec/responses.go
```diff
@@ -90,20 +90,28 @@ func (r ResponsesProps) MarshalJSON() ([]byte, error) {
...
-	var res map[string]Response
+	var res map[string]json.RawMessage
 	if err := json.Unmarshal(data, &res); err != nil {
-		return nil
+		return err
 	}
 	if v, ok := res["default"]; ok {
-		r.Default = &v
+		value := Response{}
+		json.Unmarshal(v, &value)
```
liggitt, line 99: don't we need to check for `err` here and return if this errors?

alexzielenski: whoops. thanks for catching that

pkg/validation/spec/responses.go
```diff
@@ -90,20 +90,30 @@ func (r ResponsesProps) MarshalJSON() ([]byte, error) {
...
 	}
 	for k, v := range res {
+		// Take all integral keys
 		if nk, err := strconv.Atoi(k); err == nil {
 			if r.StatusCodeResponses == nil {
 				r.StatusCodeResponses = map[int]Response{}
 			}
-			r.StatusCodeResponses[nk] = v
+			value := Response{}
+			err := json.Unmarshal(v, &value)
```
apelisse, line 112: Why not consistent?!

PR #322, opened 2022-09-29
Consistently find enum values from separate packages in openapi-gen.

### LINKED ISSUE #321: openapi-gen output is nondeterministic for enum values declared outside the package of the enum type
Given an enum type in package "foo":

```
package foo

// +enum
type Protocol string

const (
  ProtocolTCP Protocol = "TCP"
  ProtocolUDP Protocol = "UDP"
)
```

and values declared in another package "bar":

```
package bar

import "foo"

const ProtocolSCTP foo.Protocol = "SCTP"
```

instances of `foo.Protocol` in the resulting schema will have either `["TCP", "UDP"]` or `["TCP", "UDP", "SCTP"]` as the value of `enum`, depending on whether "foo" or "bar" is processed first during a [map iteration](https://github.com/kubernetes/kube-openapi/blob/b7a9e6fe3bb6adf5263a4086de91f1c567a6d9a8/pkg/generators/enum.go#L94).

benluddy:
When an enum value resides in a package other than the one containing the enum type, the output of openapi-gen varies depending on the order packages are processed.

Doing two passes -- first to discover all enum types, then to find all enum values -- makes openapi-gen always include values from all packages.

Fixes #321.

### DIFF
--- a/pkg/generators/enum.go
+++ b/pkg/generators/enum.go
@@ -90,9 +90,9 @@ func parseEnums(c *generator.Context) enumMap {
 	// First, find the builtin "string" type
 	stringType := c.Universe.Type(types.Name{Name: "string"})
 
+	// find all enum types.
 	enumTypes := make(enumMap)
 	for _, p := range c.Universe {
-		// find all enum types.
 		for _, t := range p.Types {
 			if isEnumType(stringType, t) {
 				if _, ok := enumTypes[t.Name]; !ok {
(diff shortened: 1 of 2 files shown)


pkg/generators/enum.go
```diff
@@ -102,7 +102,10 @@ func parseEnums(c *generator.Context) enumMap {
 				}
 			}
 		}
-		// find all enum values from constants, and try to match each with its type.
+	}
+
+	// find all enum values from constants, and try to match each with its type.
+	for _, p := range c.Universe {
```
sttts, line 108: there is no difference to the previous code, is there?

sttts: hmm, looks like you claim there is :) Order changes. Constants can be defined before their type. Makes sense.

PR #325, opened 2022-10-28
Use "jsonv2"/go-json-experimental to marshal OpenAPI v2

### LINKED ISSUE #324: `json.Marshal` allocates a lot of unnecessary memory
This is a follow-up of #315/#319. Now that `Unmarshaling` has been fixed, we need to worry about the next big thing, `Marshaling`.

As we can see from the pprof heap profile (alloc_space) below, `json.Marshal` is allocating a lot of memory, certainly because it serializes in buffers that are quickly discarded as these buffers are "concatenated" (e.g. https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/spec/header.go#L44-L63, https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/spec/schema.go#L435-L466)
profile002

There's clearly no great way to do that with json.Marshal v1, since when you have multiple embedded types (which we have everywhere in the spec), only one of the custom marshaller is called. That's also true for json.Marshal v2, I don't see how it could be different anyways.

I've been looking at go-json-experiment, but I can't find a good way to "embed" multiple values, all the methods require that you serialize exactly one json "value", but in most cases, we want to serialize a list of fields for each of these embedded sub-structures. I think we could give each of these types a "MarshalKeysAndValues", and construct the object itself, but then we will have to rewrite a lot of logic to extract the keys  and values.

Anyways, I'm stuck. Suggestions welcome @dsnet @alexzielenski @liggitt

apelisse:
Performance boost is also pretty impressive:
```
name                                    old time/op    new time/op    delta
SwaggerSpec_ExperimentalMarshal/json-8     102ms ± 1%      25ms ± 1%  -75.12%  (p=0.016 n=4+5)

name                                    old alloc/op   new alloc/op   delta
SwaggerSpec_ExperimentalMarshal/json-8    52.4MB ± 4%    19.9MB ± 2%  -62.02%  (p=0.008 n=5+5)

name                                    old allocs/op  new allocs/op  delta
SwaggerSpec_ExperimentalMarshal/json-8      210k ± 0%       76k ± 0%  -63.82%  (p=0.008 n=5+5)
```

Mostly because this removes the need to deserialize json into buffers that are then thrown away when the jsons are concatenated together (due to lots of embedded objects).

@dsnet @alexzielenski @liggitt 

Fixes #324


pkg/validation/spec/swagger_test.go
```diff
@@ -255,5 +267,44 @@ func BenchmarkSwaggerSpec_ExperimentalUnmarshal(b *testing.B) {
...
+	b.Run("jsonv1", func(b2 *testing.B) {
+		for i := 0; i < b2.N; i++ {
+			if _, err = json.Marshal(swagger); err != nil {
+				b2.Fatal(err)
+			}
+		}
+	})
 
+	// Our UnmarshalJSON implementation which defers to jsonv2 causes the
+	// text to be parsed/validated twice. This costs a significant amount of time.
```
apelisse, line 302: Bad copy/paste, I need to remove this.

apelisse:
Running this in Kubernetes to see if it works, already found a couple of changes that need to be updated in this PR.

https://github.com/kubernetes/kubernetes/pull/113445

apelisse:
Updated with WIP that breaks tests because marshal doesn't serialize in the same order. There's also a slight different change in how unicodes are encoded apparently which is causing the length to not always be strictly equal (though unmarshaling seems to restore to the previous state).

Also, so far I've only found one location in kubernetes/kubernetes that need to be updated: staging/src/k8s.io/apiserver/pkg/util/openapi/proto.go

dsnet:
If you're not already, you could consider a semantic comparison of JSON objects with:
```go
var got, want []byte = ... // where x and y are raw JSON values
if d := cmp.Diff(got, want, cmp.Transform(in []byte) (out any) {
    // Note that integers larger than 2^53 will lose precision
    // since they will be represented using a Go float64.
    // There are workarounds for this, but calling it out.
    if err := json.Unmarshal(in, &out); err != nil {
        return in
    }
    return out
}); d != "" {
    t.Errorf("JSON mismatch (-got +want):\n%s", d)
}
```

dsnet:
In v1, the characters in "<>&" were escaped for dubious security reasons. In v2, we stop doing that, and follow RFC recommendations to use the minimal escaped string output.

apelisse:
Tests should be passing, it's a little bit yucky how the structs are duplicated (more than once), suggestions welcome, ready to merge other than that. Thanks!

pkg/validation/spec/header.go
```diff
@@ -62,6 +62,43 @@ func (h Header) MarshalJSON() ([]byte, error) {
 	return swag.ConcatJSON(b1, b2, b3, b4), nil
 }
 
+func (h Header) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	type SimpleSchemaOmitZero struct {
```
apelisse, line 66: There's a lot of repetition in these types (SimpleSchemaOmitZero is defined at least a couple of times), and even the `x` structs are re-use between MarshalNext and UnmarshalNext. Should we name them (uncapitalized) next to the types they copy so that we can re-use?

dsnet: `omitzero` is ignored by the v1 "json" marshaler, we could just put those on original types. That will probably remove most of the duplication.

apelisse: Actually, I copied these types because I couldn't get a combination that behaves the same way with both v1 and v2.
v1 requires omitempty, v2 requires omitzero without omitempty.

dsnet: Couldn't you put both "omitzero,omitempty" for some set of these? The "omitzero" will still be ignored by v1, but that's okay. For example, that would work for all the cases with boolean and number types.

apelisse: Yeah, for boolean and number types, I only needed to add omitzero to existing types. I had to copy the whole thing to have omitzero without omitempty for pointers to structs (since omitempty is now removing structs that are serialized to `{}`)

dsnet: Ah, got it. I'll make a note of this for v1-to-v2 compatibility issues.

apelisse: To be honest, it'd have been very nice if we could get a way to have the flags to give me something compatible. I basically wrote something like this: https://go.dev/play/p/4yckWam8Pw1 to try to find a combination that works with both, but there was no way to get the same output in both v1 and v2 without duplicating the structure.

dsnet: It's possible we'll add flags to v2 to have some more v1-like behavior. Getting the v1-to-v2 compatibility story is possibly the hardest part of the puzzle.

pkg/validation/spec/swagger.go
```diff
@@ -46,6 +46,21 @@ func (s Swagger) MarshalJSON() ([]byte, error) {
 	return swag.ConcatJSON(b1, b2), nil
 }
 
+func (s Swagger) MarshalNext() ([]byte, error) {
+	return jsonv2.Marshal(s)
+}
```
liggitt, line 51: is there a reason not to follow the same pattern as Unmarshal, like 

```
func (s Swagger) MarshalJSON() ([]byte, error) {
	if internal.UseOptimizedJSONMarshaling {
		return jsonv2.Marshal(s)
	}
  ...
}
```

then all the call-sites could switch to calling `Swagger#MarshalJSON()` rather than calling `MarshalNext()` directly, which I assume we don't really want long-term

pkg/validation/spec/swagger.go
```diff
@@ -46,6 +46,21 @@ func (s Swagger) MarshalJSON() ([]byte, error) {
...
+
+// MarshalJSON marshals this swagger structure to json
+func (s Swagger) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	var x struct {
+		Extensions
+		SwaggerProps
+	}
+	x.Extensions = s.Extensions
+	x.SwaggerProps = s.SwaggerProps
```
liggitt, line 60: curious if there's any alloc difference between making the fields of these unified types pointers so they only need to take the address of the subfield we're marshaling, e.g.

```go
var x struct {
  *Extensions
  *SwaggerProps
}
x.Extensions = &s.Extensions
x.SwaggerProps = &s.SwaggerProps
```

(if that improves allocs, the same comment might apply other places we're doing this)

pkg/util/jsontesting/json_roundtrip.go
```diff
@@ -22,9 +22,28 @@ import (
...
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
+func JsonCompare(got, want []byte) error {
+	if d := cmp.Diff(got, want, cmp.Transformer("JSONBytes", func(in []byte) (out interface{}) {
+		// Note that integers larger than 2^53 will lose precision
+		// since they will be represented using a Go float64.
+		// There are workarounds for this, but calling it out.
+		if err := json.Unmarshal(in, &out); err != nil {
```
liggitt, line 37: `sigs.k8s.io/json` has `UnmarshalCaseSensitivePreserveInts` which will preserve ints... do we want to use that?

pkg/validation/spec/header.go
```diff
@@ -62,6 +62,43 @@ func (h Header) MarshalJSON() ([]byte, error) {
...
+		Enum             []interface{} `json:"enum,omitempty"`
+	}
+	var x struct {
+		CommonValidationsOmitZero
+		SimpleSchemaOmitZero
+		Extensions
+		HeaderProps
+	}
+	x.CommonValidationsOmitZero = CommonValidationsOmitZero(h.CommonValidations)
+	x.SimpleSchemaOmitZero = SimpleSchemaOmitZero(h.SimpleSchema)
```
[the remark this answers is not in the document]
apelisse: Changing the type, removing a variable or adding a variable triggers the same compile-time error:
```
pkg/validation/spec/header.go:96:58: cannot convert h.CommonValidations (variable of type CommonValidations) to type CommonValidationsOmitZero
```

pkg/validation/spec/swagger.go
apelisse: I don't know what to think:
```
name                                      old time/op    new time/op    delta
SwaggerSpec_ExperimentalMarshal/jsonv2-8    31.5ms ± 1%    31.6ms ± 1%     ~     (p=0.743 n=8+9)

name                                      old alloc/op   new alloc/op   delta
SwaggerSpec_ExperimentalMarshal/jsonv2-8    22.4MB ± 0%    19.1MB ± 0%  -14.70%  (p=0.000 n=8+9)

name                                      old allocs/op  new allocs/op  delta
SwaggerSpec_ExperimentalMarshal/jsonv2-8     91.3k ± 0%    105.7k ± 0%  +15.73%  (p=0.000 n=7+10)
```

pkg/util/jsontesting/json_roundtrip.go
apelisse: I don't have a strong opinion, this is strictly with-in the test and we're not having any problems with it?

pkg/validation/spec/header.go
apelisse: Using reflection and all? Between the compilation failing and the round-trip tests, I don't see this as necessary nor useful.

pkg/validation/spec/swagger.go
liggitt: eh, probably not worth it then

pkg/validation/spec/header.go
liggitt: I thought we might be vulnerable to case-insensitivity in the json tag names, but it looks like the tests are comparing the actual json marshaled bytes, so I agree.

package-private `___OmitZero` types next to the existing types _would_ let you avoid declaring `SimpleSchemaOmitZero` and `CommonValidationsOmitZero` three times

pkg/validation/spec/swagger.go
apelisse: Done!

pkg/validation/spec/swagger.go
apelisse: 🤷

pkg/util/jsontesting/json_roundtrip.go
apelisse: done

apelisse:
OK I think that's good, PTAL!

pkg/validation/spec/operation.go
```diff
@@ -36,12 +36,29 @@ type OperationProps struct {
 	Summary      string                 `json:"summary,omitempty"`
 	ExternalDocs *ExternalDocumentation `json:"externalDocs,omitempty"`
 	ID           string                 `json:"operationId,omitempty"`
-	Deprecated   bool                   `json:"deprecated,omitempty"`
+	Deprecated   bool                   `json:"deprecated,omitempty,omitzero"`
```
liggitt, line 39: revert this line if we have a dedicated operationPropsOmitZero type?

pkg/validation/spec/paths.go
```diff
@@ -114,7 +114,9 @@ func (p *Paths) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Deco
 				p.Paths[k] = pi
 			default:
 				_, err := dec.ReadValue() // skip value
-				return err
+				if err != nil {
+					return err
+				}
```
liggitt, line 119: should this merge first for backporting? also, should there be a testcase to exercise this?

pkg/validation/spec/paths.go
```diff
@@ -142,3 +147,14 @@ func (p Paths) MarshalJSON() ([]byte, error) {
...
 }
+
+func (p Paths) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	m := make(map[string]interface{}, len(p.Extensions)+len(p.Paths))
+	for k, v := range p.Extensions {
+		m[k] = v
+	}
+	for k, v := range p.Paths {
+		m[k] = v
+	}
```
liggitt, line 158: Paths#MarshalJSON() filters to paths starting with `/`; should this?

pkg/util/jsontesting/json_roundtrip.go
```diff
@@ -22,9 +22,26 @@ import (
...
 
+	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
+	kjson "sigs.k8s.io/json"
+
+	"github.com/go-openapi/jsonreference"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
+func JsonCompare(got, want []byte) error {
```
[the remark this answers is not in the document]
liggitt: Talked offline with @apelisse, the new json library explicitly does not marshal maps in sorted order. I think we need our json marshaling to remain deterministic, so I don't think this should merge until we can upstream an optional "output maps in sorted-key order" behavior (which I think is pretty important, actually) or modify our invocation of the marshaling to accomplish the same thing (in a way that doesn't modify the internal copy of the json-experimental lib).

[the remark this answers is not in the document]
apelisse: @dsnet any chance we can have a marshal option to get maps sorted? I have a tentative unoptimized piece of code that does the sorting and it's still a massive improvement compared to json v1.

pkg/validation/spec/swagger_test.go
```diff
@@ -185,6 +220,17 @@ func TestSwaggerSpec_ExperimentalUnmarshal(t *testing.T) {
 
 	actual := Swagger{}
 	internal.UseOptimizedJSONUnmarshaling = true
+
+	// Serialize into JSON again
```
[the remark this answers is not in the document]
apelisse: Thanks, that's a left-over from when we had to different methods.

pkg/util/jsontesting/json_roundtrip.go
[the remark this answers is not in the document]
dsnet: This is a use case we need at @tailscale as well, so I might be able to implement it in O(weeks).

pkg/validation/spec/info.go
```diff
@@ -89,6 +89,32 @@ func (e Extensions) GetObject(key string, out interface{}) error {
 	return nil
 }
 
+// sanitized either uses:
+// - the map that was passed as an argument
+// - or, if the map is nil, returns the extension map if all the extensions were valid
+// - or allocates a new map.
+// In other words, if you specify a nil map as a parameter, don't edit the returned map.
+func (e Extensions) sanitized(m map[string]any) Extensions {
```
apelisse, line 97: I feel like this signature is too convoluted, I should probably split that into `sanitized` and `sanitizedWithMap` ... But I don't have a strong enough opinion ...

pkg/internal/third_party/go-json-experiment/json/arshal_any.go
```diff
@@ -4,7 +4,10 @@
 
 package json
 
-import "reflect"
```
apelisse, line 7: @dsnet FYI Here and below is my very sad attempt at getting deterministic behavior. It's definitely not great performance-wise (see below), but I'm assuming a few of these arrays could be re-used via a pool.

```
> benchstat unsorted sorted
name                                                    old time/op    new time/op    delta
SwaggerSpec_ExperimentalMarshal/jsonv2_via_jsonv1-8       47.8ms ± 6%    51.7ms ± 8%   +8.21%  (p=0.001 n=9+10)
SwaggerSpec_ExperimentalMarshal/jsonv2-8                  31.7ms ± 4%    33.3ms ± 2%   +5.23%  (p=0.000 n=10+9)
SwaggerSpec_ExperimentalMarshal/jsonv2-canonicalized-8    49.9ms ± 4%    49.6ms ± 5%     ~     (p=0.481 n=10+10)

name                                                    old alloc/op   new alloc/op   delta
SwaggerSpec_ExperimentalMarshal/jsonv2_via_jsonv1-8       36.9MB ±11%    40.0MB ±12%     ~     (p=0.075 n=10+10)
SwaggerSpec_ExperimentalMarshal/jsonv2-8                  22.1MB ± 2%    24.7MB ± 2%  +11.78%  (p=0.000 n=10+10)
SwaggerSpec_ExperimentalMarshal/jsonv2-canonicalized-8    29.7MB ±16%    29.3MB ± 0%     ~     (p=0.220 n=10+6)

name                                                    old allocs/op  new allocs/op  delta
SwaggerSpec_ExperimentalMarshal/jsonv2_via_jsonv1-8        91.3k ± 0%    111.8k ± 0%  +22.40%  (p=0.000 n=10+10)
SwaggerSpec_ExperimentalMarshal/jsonv2-8                   91.3k ± 0%    111.8k ± 0%  +22.41%  (p=0.000 n=10+10)
SwaggerSpec_ExperimentalMarshal/jsonv2-canonicalized-8     92.9k ± 0%    113.4k ± 0%  +22.08%  (p=0.000 n=10+8)
```

pkg/util/jsontesting/json_roundtrip.go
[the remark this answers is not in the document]
dsnet: My apologies for the delay. I've been on paternity leave. My return date in January 30th, so I'll start looking at these things more in depth then.

pkg/internal/third_party/go-json-experiment/json/arshal_any.go
dsnet: My apologies for the delay. I just merged https://github.com/go-json-experiment/json/commit/540f014424240312547dcccddf11a8229ca0f463, which adds `MarshalOptions.Deterministic`.

apelisse: Thank you so much! That's absolutely great

apelisse:
OK, tests pass, should be working now, pretty good improvement considering this is definitely the worst thing going on in an empty api-server right now.

apelisse:
@Jefftree @alexzielenski either of you has something to add? If not, feel free to cancel the hold, thanks!

pkg/util/jsontesting/json_roundtrip.go
```diff
@@ -39,7 +55,12 @@ type RoundTripTestCase struct {
 	ExpectedUnmarshalError string
 }
 
-func (t RoundTripTestCase) RoundTripTest(example json.Unmarshaler) error {
+type MarshalerUnmarsaheler interface {
```
Jefftree, line 58: nit: typo

pkg/validation/spec/operation.go
```diff
@@ -42,6 +42,23 @@ type OperationProps struct {
...
+type operationPropsOmitZero struct {
+	Description  string                 `json:"description,omitempty"`
+	Consumes     []string               `json:"consumes,omitempty"`
+	Produces     []string               `json:"produces,omitempty"`
+	Schemes      []string               `json:"schemes,omitempty"`
+	Tags         []string               `json:"tags,omitempty"`
+	Summary      string                 `json:"summary,omitempty"`
+	ExternalDocs *ExternalDocumentation `json:"externalDocs,omitzero"`
+	ID           string                 `json:"operationId,omitempty"`
+	Deprecated   bool                   `json:"deprecated,omitempty,omitzero"`
```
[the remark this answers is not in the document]
dsnet: In "go-json-experiment", the semantics of `omitempty` was changed to mean "omit any empty JSON value" which is currently defined as a JSON null, empty string, object, or array; note that it excludes false. It also supports `omitzero` which means "omit any zero Go value" according to the zero value definition in the Go language; which includes false.

Have both `omitzero` and `omitempty` specified takes the OR of the two and provides a way for this to operate the same under both "encoding/json" and "go-json-experiment".

pkg/util/jsontesting/json_roundtrip.go
apelisse: Fixed

PR #328, opened 2022-11-15
cleanup code in builder3 test

Huang-Wei:
Seems the logic of calculating `defNames` is redundant.

/kind cleanup


apelisse:
/lgtm
/approve
thanks

PR #331, opened 2022-11-23
Fix minor bug not deserializing a full-value in Paths#UnmarshalNextJSON

apelisse:


pkg/validation/spec/paths_test.go
```diff
@@ -33,15 +33,19 @@ var paths = Paths{
 }
 
 const pathsJSON = `{"x-framework":"go-swagger","/":{"$ref":"cats"}}`
+const pathsJSONInvalidKey = `{"x-framework":"go-swagger","/":{"$ref":"cats"}, "not-path-nor-extension": "invalid"}`
```
liggitt, line 36: the issue before was that we didn't continue decoding on a syntactically valid but semantically invalid value, right? should the invalid value come *between* the `x-framework` and `/` keys to make sure we continued decoding?

apelisse: Yes, though continuing to decode immediately resulted in an error for the json parser since its state machine is messed-up.

WIthout the fix, the new test fails with the following error:
```
--- FAIL: TestIntegrationPaths (0.00s)
    paths_test.go:43: 
        	Error Trace:	paths_test.go:43
        	Error:      	Received unexpected error:
        	            	json: cannot unmarshal Go value of type spec.Paths: must read exactly one JSON value
        	Test:       	TestIntegrationPaths
    structs_test.go:53: 
        	Error Trace:	structs_test.go:53
        	            				paths_test.go:48
        	Error:      	Failed to unmarshal JSON
        	Test:       	TestIntegrationPaths
        	Messages:   	unable to unmarshal from json (json: cannot unmarshal Go value of type spec.Paths: must read exactly one JSON value): {"x-framework":"go-swagger","/":{"$ref":"cats"}, "not-path-nor-extension": "invalid"}
FAIL
FAIL	k8s.io/kube-openapi/pkg/validation/spec	2.188s
FAIL
```

But you're right, it doesn't hurt to just put it in the middle to be safe.

apelisse: PTAL

PR #332, opened 2022-11-30
Publish ETag for root OpenAPI V3 document

Jefftree:
Publish ETag for root OpenAPI V3 document

Partially addresses https://github.com/kubernetes/kubernetes/issues/113903

http.ServeContent automatically handles the `If-None-Match` header if it's passed in the request. This is verified in the tests.

/assign @apelisse


pkg/handler3/handler.go
```diff
@@ -191,6 +191,7 @@ func ToV3ProtoBinary(json []byte) ([]byte, error) {
 
 func (o *OpenAPIService) HandleDiscovery(w http.ResponseWriter, r *http.Request) {
 	data, _ := o.getGroupBytes()
+	w.Header().Set("Etag", strconv.Quote(computeETag(data)))
```
apelisse, line 194: Do we need to change/set some cache headers too? I think we only accept `json`, should we set the content-type? Should we reject requests with non-`application/json` accept headers? Since we only have one content-type, I guess we probably don't need `Vary`?

pkg/handler3/handler_test.go
```diff
@@ -104,15 +194,27 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
 		if err != nil {
 			t.Errorf("Accept: %v: Unexpected error in serving HTTP request: %v", tc.acceptHeader, err)
 		}
+		defer resp.Body.Close()
 
 		if resp.StatusCode != tc.respStatus {
 			t.Errorf("Accept: %v: Unexpected response status code, want: %v, got: %v", tc.acceptHeader, tc.respStatus, resp.StatusCode)
 		}
-		defer resp.Body.Close()
+		if tc.respStatus != 200 {
```
apelisse, line 209: Don't we want to test that the content is empty when we get a 304?

pkg/handler3/handler_test.go
```diff
@@ -82,19 +83,108 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 		respStatus   int
 		urlPath      string
 		respBody     []byte
+		expectedETag string
+		sendETag     bool
```
apelisse, line 87: instead of `sendETag`, you could have a `IfNoneMatch` that is configurable (possibly by setting the same value), so that you can test the behavior of the cache at the same time? Though I think you have another test somewhere that does that?

pkg/handler3/handler.go
Jefftree: Updated. Unsure about rejecting non-"application/json" requests since we have a few cases that might also need to be checked for `*/*`, `application/*`, */json`, empty.

PR #333, opened 2022-11-30
OpenAPI V2 add Content Type return

Jefftree:
Partially addresses https://github.com/kubernetes/kubernetes/issues/113903

OpenAPI V3 fix will be in a separate PR.

/assign @apelisse

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -203,6 +203,9 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 							return
 						}
 					}
+					contentType := accepts.Type + "/" + accepts.SubType
+					w.Header().Set("Content-Type", contentType)
+
 					// ETag must be enclosed in double quotes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
 					w.Header().Set("Etag", strconv.Quote(etag))
 					// ServeContent will take care of caching using eTag.
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -89,8 +89,8 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 			continue
 		}
 
-		if resp.Header.Get("Content-Type") != tc.expectedContentType {
-			t.Errorf("Content-Type: Unexpected content type, want: %v, got: %v", tc.expectedContentType, resp.Header.Get("Content-Type"))
+		if got := resp.Header.Get("Content-Type"); got != tc.expectedContentType {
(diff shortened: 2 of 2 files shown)


pkg/handler/handler_test.go
```diff
@@ -84,6 +85,13 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 		if resp.StatusCode != tc.respStatus {
 			t.Errorf("Accept: %v: Unexpected response status code, want: %v, got: %v", tc.acceptHeader, tc.respStatus, resp.StatusCode)
 		}
+		if resp.StatusCode != 200 {
+			continue
+		}
+
+		if resp.Header.Get("Content-Type") != tc.expectedContentType {
+			t.Errorf("Content-Type: Unexpected content type, want: %v, got: %v", tc.expectedContentType, resp.Header.Get("Content-Type"))
```
apelisse, line 93: Minor nit:
```suggestion
		if got := resp.Header.Get("Content-Type"); got != tc.expectedContentType {
			t.Errorf("Content-Type: Unexpected content type, want: %v, got: %v", tc.expectedContentType, got)
```

apelisse (commented):
/lgtm
/approve
/hold if you want to fix the nit

Jefftree:
Thanks, updated.

pkg/handler/handler_test.go
```diff
@@ -84,6 +85,13 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 		if resp.StatusCode != tc.respStatus {
 			t.Errorf("Accept: %v: Unexpected response status code, want: %v, got: %v", tc.acceptHeader, tc.respStatus, resp.StatusCode)
 		}
+		if resp.StatusCode != 200 {
```
apelisse, line 88: I'll just mention that while I'm not super excited about this, I think it's fine given the current state of the test.

pkg/handler/handler.go
```diff
@@ -203,6 +203,9 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 							return
 						}
 					}
+					contentType := accepts.Type + "/" + accepts.SubType
+					w.Header().Set("Content-Type", contentType)
```
liggitt, line 207: when this got pulled into kubernetes, we're seeing this error:

> Error from server (InternalError): Internal error occurred: mime: unexpected content after media subtype

liggitt: failed in https://github.com/kubernetes/kubernetes/pull/114869

liggitt: not sure if this PR or https://github.com/kubernetes/kube-openapi/pull/334/files was the cause... both were in the diff pulled in in that PR

PR #334, opened 2022-12-01
Remove mime AddExtensionsType

Jefftree:
I was going through test coverage and noticed this block being unused/untested. It looks to be unused based on the mime documentation (https://cs.opensource.google/go/go/+/go1.19.3:src/mime/type.go;l=167), the items added by `mime.AddExtensionsType` are only used if the items are fetched via `mime.ExtensionsByType` which is not seen in this repo or k/k. The other usage of AddExtensionsType is in k/k https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/discovery/discovery_client_test.go#L487 and seems like a no-op since the values added are never retrieved.

I think we can remove this but I'm not 100% sure and would like to get some feedback.

/cc @apelisse 
/cc @liggitt

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -22,7 +22,6 @@ import (
 	"crypto/sha512"
 	"encoding/json"
 	"fmt"
-	"mime"
 	"net/http"
 	"strconv"
 	"sync"
@@ -39,15 +38,6 @@ import (
 	"k8s.io/kube-openapi/pkg/common/restfuladapter"
 	"k8s.io/kube-openapi/pkg/internal/handler"
 	"k8s.io/kube-openapi/pkg/validation/spec"
-)
-
-const (
-	jsonExt = ".json"
-
-	mimeJson = "application/json"
-	// TODO(mehdy): change @68f4ded to a version tag when gnostic add version tags.
-	mimePb   = "application/com.github.googleapis.gnostic.OpenAPIv2@68f4ded+protobuf"
-	mimePbGz = "application/x-gzip"
 )
 
 func computeETag(data []byte) string {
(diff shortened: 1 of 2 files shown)


apelisse:
I'm looking into it too but I don't see how this is used?

apelisse:
I just can't parse the [mime](https://pkg.go.dev/mime) documentation, there seems to be side-effects but it's really unclear what they are.

apelisse:
https://codeandlife.com/2022/02/12/fixing-mime-type-golang-net-http-fileserver/

This seems to indicate such a side-effect, totally no idea how this works.

apelisse:
This *maybe* uses the mime information to do something: https://pkg.go.dev/net/http#ServeContent
I don't think we're using this function, that function is using [`DetectContentType`](https://pkg.go.dev/net/http#DetectContentType) based on the content, which may or may not detect the content, but `mime.AddExtensionType` probably uses the extension.
All-in-all, we have to fish, do you know if this mime information use to set the content-type properly for our specific use-case?

Jefftree:
Our Content-Type is returned incorrectly and all our requests return text/plain or text/octet-stream which you brought up in your bug https://github.com/kubernetes/kubernetes/issues/113903#issue-1448958851. 

```
DetectContentType implements the algorithm described at https://mimesniff.spec.whatwg.org/ to determine the Content-Type of the given data. It considers at most the first 512 bytes of data. DetectContentType always returns a valid MIME type: if it cannot determine a more specific one, it returns "application/octet-stream".
```

So we are using [http.ServeContent](https://github.com/kubernetes/kube-openapi/blob/master/pkg/handler/handler.go#L200) and we do have application/octet-stream returned...

Our OpenAPI return type is determined by the Accept header and not file extension, I wonder if that could be why the mime information isn't used.

apelisse:
From [`http.ServeContent`](https://pkg.go.dev/net/http#ServeContent):
> If the response's Content-Type header is not set, ServeContent first tries to deduce the type from name's file extension and, if that fails, falls back to reading the first block of the content and passing it to DetectContentType. The name is otherwise unused; in particular it can be empty and is never sent in the response.

It looks like if we want to set the content-type, then we should and the mime thing will not be used. Let's merge #333 first, but that should be good.

PR #335, opened 2022-12-01
Remove unused gzip function

Jefftree:
gzip functionality is provided by the gziphandler (https://github.com/kubernetes/kube-openapi/blob/master/pkg/handler/handler.go#L31) library. This code path is unused

/assign @apelisse

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -18,7 +18,6 @@ package handler
 
 import (
 	"bytes"
-	"compress/gzip"
 	"crypto/sha512"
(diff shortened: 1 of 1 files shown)


apelisse:
thanks, every time I see, I'm wondering why we have this lol
/lgtm
/approve

apelisse:
Do we have a test that shows that we support gzip btw?

Jefftree:
> Do we have a test that shows that we support gzip btw?

That is lacking coverage, I can look into it.

PR #337, opened 2022-12-05
Add gofmt to CI and upgrade to go 1.19

Jefftree:
Add gofmt to CI and upgrade to go 1.19

The diff code was partially copied from ./hack/verify-gofmt.sh from k/k.

This code also updated some of the files in `internal/third_party`, should those be ignored?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.18', '1.x' ]
+        go: [ '1.19', '1.18', '1.x' ]
         gostable: [true]
         include:
         - go: '1.x'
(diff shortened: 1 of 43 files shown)


apelisse:
Curious why you moved this to 1.19? I think we probably want to continue to handle 1.18 though possibly we would branch the repo if we wanted to cherry-pick something in an older version of kubernetes. What's your opinion @liggitt?

Jefftree:
I was trying to match with k/k version but that makes sense that we want to continue handling older versions for cherry picks. What should our strategy be in general around version upgrades?

liggitt:
I don't think this PR alone would prevent picking up HEAD in older kubernetes versions, but bumping the go version would enable using 1.19-only go features, etc. Unless there's a compelling reason we *need* to update, I would probably recommend keeping this on the trailing go version still in support (which is go1.18 until ~Feb 2023).

FWIW, I'm working on getting supported kubernetes release branches up to go1.19 in https://github.com/kubernetes/kubernetes/pull/113956 and https://github.com/kubernetes/kubernetes/pull/113983

Jefftree:
ack, removed the go version upgrade

apelisse:
I think it still makes sense to have `1.19` hard-coded in the list in CI.
/lgtm
/approve

Jefftree:
Added 1.19 to the CI list and maintained 1.18 for go mod version.

apelisse:
Still good to me, thanks
/lgtm

PR #342, opened 2022-12-16
New handling method to create spec lazily

apelisse:
And a minor refactor.

This will help maybe shave some time we lose generating the spec too eagerly. This should be totally backwards compatible.

cc @alexzielenski @lavalamp

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -49,10 +49,8 @@ func computeETag(data []byte) string {
 // OpenAPIService is the service responsible for serving OpenAPI spec. It has
 // the ability to safely change the spec while serving it.
 type OpenAPIService struct {
-	// rwMutex protects All members of this service.
-	rwMutex sync.RWMutex
-
-	lastModified time.Time
+	// mutex protects All members of this service.
+	mutex sync.Mutex
 
 	jsonCache  handler.HandlerCache
 	protoCache handler.HandlerCache
@@ -68,37 +66,9 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
 	return o, nil
 }
 
-func (o *OpenAPIService) getSwaggerBytes() ([]byte, string, time.Time, error) {
-	o.rwMutex.RLock()
-	defer o.rwMutex.RUnlock()
-	specBytes, err := o.jsonCache.Get()
-	if err != nil {
-		return nil, "", time.Time{}, err
-	}
-	etagBytes, err := o.etagCache.Get()
-	if err != nil {
-		return nil, "", time.Time{}, err
-	}
-	return specBytes, string(etagBytes), o.lastModified, nil
-}
-
-func (o *OpenAPIService) getSwaggerPbBytes() ([]byte, string, time.Time, error) {
-	o.rwMutex.RLock()
-	defer o.rwMutex.RUnlock()
-	specPb, err := o.protoCache.Get()
-	if err != nil {
-		return nil, "", time.Time{}, err
-	}
-	etagBytes, err := o.etagCache.Get()
-	if err != nil {
-		return nil, "", time.Time{}, err
-	}
-	return specPb, string(etagBytes), o.lastModified, nil
-}
-
 func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
-	o.rwMutex.Lock()
-	defer o.rwMutex.Unlock()
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
 	o.jsonCache = o.jsonCache.New(func() ([]byte, error) {
 		return json.Marshal(openapiSpec)
 	})
@@ -116,8 +86,6 @@ func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
 		}
 		return []byte(computeETag(json)), nil
 	})
-	o.lastModified = time.Now()
-
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/handler/handler.go
```diff
@@ -140,17 +144,36 @@ func RegisterOpenAPIVersionedService(spec *spec.Swagger, servePath string, handl
...
 // RegisterOpenAPIVersionedService registers a handler to provide access to provided swagger spec.
 func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handler common.PathHandler) error {
 	accepted := []struct {
 		Type           string
 		SubType        string
-		GetDataAndETag func() ([]byte, string, time.Time, error)
+		GetDataAndETag func() ([]byte, string, error)
 	}{
 		{"application", "json", o.getSwaggerBytes},
 		{"application", "com.github.proto-openapi.spec.v2@v1.0+protobuf", o.getSwaggerPbBytes},
```
lavalamp, line 172: Shouldn't these each have a separate etag cache?

pkg/handler/handler.go
```diff
@@ -116,11 +118,13 @@ func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
 		}
 		return []byte(computeETag(json)), nil
 	})
-	o.lastModified = time.Now()
-
 	return nil
 }
 
+func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
```
lavalamp, line 124: I guess I am confused about how this actually defers anything if this function is still getting called?

apelisse: Well, UpdateSpec continues to work like it did, but there is a new `UpdateSpecLazy` that allows you to give a method rather than a resolved spec, which isn't used yet in k/k.

pkg/handler/handler.go
apelisse: Because of the Vary header, I don't think that matters, clients can't re-use etags across content-type.

pkg/handler/handler.go
lavalamp: Ah the call comes from k/k? got it

pkg/handler/handler.go
```diff
@@ -140,17 +144,36 @@ func RegisterOpenAPIVersionedService(spec *spec.Swagger, servePath string, handl
 	return o, o.RegisterOpenAPIVersionedService(servePath, handler)
 }
 
+type ETagCache struct {
```
alexzielenski, line 147: does this need to be exported mr api cleanup guy

pkg/handler/handler.go
```diff
@@ -181,6 +204,8 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 					contentType := accepts.Type + "/" + accepts.SubType
 					w.Header().Set("Content-Type", contentType)
 
+					lastModified := etagCache.LastModified(etag)
```
lavalamp, line 207: I was expecting both content types to have distinct etags, but this code path (which can serve either one) clobbers the time and existing etag either way.

But I think I'm wrong and the etag is shared by both formats. Is that intentional?

pkg/handler/handler.go
lavalamp: I asked a more specific question below.

pkg/handler/handler.go
```diff
@@ -68,39 +66,43 @@ func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
...
 	if err != nil {
-		return nil, "", time.Time{}, err
+		return nil, "", err
 	}
-	return specPb, string(etagBytes), o.lastModified, nil
+	return specPb, string(etagBytes), nil
 }
 
-func (o *OpenAPIService) UpdateSpec(openapiSpec *spec.Swagger) (err error) {
+func (o *OpenAPIService) UpdateSpecLazy(specGetter func() (*spec.Swagger, error)) error {
```
lavalamp, line 97: Can you add a godoc comment explaining what is going on here, when the function will be evaluated, etc?

pkg/handler/handler.go
alexzielenski: The tag is shared by both formats, I didnt write the original code, but I think it is intentional. The e-tag does not necessarily depend upon the byte value of the response.

[From MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#directives):
>  The method by which ETag values are generated is not specified. Typically, the ETag value is a hash of the content, a hash of the last modification timestamp, or just a revision number. 

I do think it is confusing that the per-content-type `GetDataAndEtag()` returns an etag, but the etag doesn't vary according to content type.

 Given this is now an invariant now being required by the etagcache the new code might be more error prone: alternating requests of each content type would keep bumping `lastModified` if the invariant was violated.

alexzielenski:
nit piling on about the confusing etag-contenttype relationship, but lgtm once comments addressed

apelisse:
I spent some time looking at this from different perspective, right now this is the best I could see :-)

pkg/handler/handler.go
```diff
@@ -168,21 +150,38 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 						continue
 					}
 
+					contentType := accepts.Type + "/" + accepts.SubType
+					w.Header().Set("Content-Type", contentType)
+
+					o.mutex.Lock()
```
Jefftree, line 156: It feels a bit weird holding a write lock when serving the OpenAPI. Not blocking but maybe we could refactor this lock into the cache instead of locking/unlocking in the serve method.

apelisse: Me too, initially. But a few things to note:
1. There's nothing we can do if we have two requests coming-in for the data and the data isn't ready. We should let the first one that arrived prepare the content, and the others should wait passively.
2. We need the etag and the data to correspond to the same data, so we shouldn't release the lock in between the two
3. Once the data is ready, the lock will be held for a minimal amount of time, that's fine
4. We don't actually hold the lock while "serving" the data, just while the data is being prepared (if it's not), we can still send data to more than 1 client at a time.
I think the code is fine.

apelisse:
/close

Actually this is not good, because we won't have a good way to know when the spec is invalidated, so we can't cache it in that handler indefinitely.

PR #344, opened 2022-12-29
Bump github.com/go-openapi/jsonreference

### LINKED ISSUE #356: PuerkitoBio/urlesc is archived
https://github.com/kubernetes/kube-openapi/blob/10e5f0effbd23ac351f8da2138292759505a93a8/go.mod#L34

follow up of https://github.com/kubernetes/kubernetes/pull/114912

olivierlemasle:
This includes the removal of indirect dependency `github.com/PuerkitoBio/urlesc`, which is now unmaintained and archived.

Fixes #356


olivierlemasle:
Closing as tests fail.

apelisse:
I'd be happy to see a change that removes dependencies, I don't see what the test failures are, are they unfixable?

olivierlemasle:
@apelisse I've fixed the tests and re-opened the PR.

Tests failed because `github.com/go-openapi/jsonreference` changed the way it normalizes URLs internally (see the [diff between v0.19.3 and v0.20.0](https://github.com/go-openapi/jsonreference/compare/v0.19.3...v0.20.0)) and now **unexported** URL field in `github.com/go-openapi/jsonreference.Ref` may differ when compared with `reflect.DeepEqual`.

I've switched to using `github.com/google/go-cmp/cmp` to compare `Swagger` instances (it was already used to get the diff), with the `SwaggerDiffOptions` option which compare `Ref` by `String()`.

apelisse:
Have you tried to use this code in k8s? I'm curious if this is going to break something. Though even if it does, we might want to bite the bullet anyway ...

olivierlemasle:
@apelisse I tried `./hack/update-codegen.sh` on k/k with kube-openapi updated to this branch, and there was nothing wrong. Should I also test something else?

apelisse:
Yeah, make a PR out of it and see if it passes the tests

olivierlemasle:
Thanks @apelisse 
I've created https://github.com/kubernetes/kubernetes/pull/114879

apelisse:
I think that's good, thanks
/lgtm
/assign @liggitt

pkg/validation/spec/gnostic_test.go
```diff
@@ -55,7 +55,7 @@ func gnosticCommonTest(t testing.TB, fuzzer *fuzz.Fuzzer) {
 	ok, err := actual.FromGnostic(gnosticSpec)
 	require.NoError(t, err)
 	require.True(t, ok)
-	if !reflect.DeepEqual(expected, actual) {
+	if !cmp.Equal(expected, actual, SwaggerDiffOptions...) {
```
liggitt, line 58: curious about this change... did this test fail. without this? same question below

apelisse: Yes, because the new reference has unexported fields, so you need to compare refs as a string.

apelisse: That's why I made OP run the tests in k/k, which seemed to have passed. I can't think of a place where we directly compare them?

olivierlemasle:
Hmm, I thought an approver's `/lgtm` was simultaneously interpreted as an `/approve`...
@apelisse Could you please `/approve` this? Perhaps as you're not an approver in `/pkg/validation`, you need to explicitly `/approve` this one for the root directory?

apelisse:
Yeah, I typed lgtm by mistake, I meant
/approve

Sorry :-)

apelisse:
I'm surprised that this wasn't affected somehow:
https://github.com/kubernetes/kube-openapi/blob/0d48a347b3839510b3902c604b01a5c1de7aa36a/pkg/aggregator/aggregator.go#L199

Any clue?

liggitt:
hmm... is that code exercised in tests?

apelisse:
I need to look deeper, but I would say we definitely merge specs to generate the `api/openapi-spec/swagger.json`, so we should have noticed if this had failed/generated something wrong?

liggitt:
> I need to look deeper, but I would say we definitely merge specs to generate the `api/openapi-spec/swagger.json`, so we should have noticed if this had failed/generated something wrong?

is the failure mode a panic or a spurious false? a spurious false could trigger `..._2` entries in a merged openapi spec

apelisse:
I've just looked at the test in https://github.com/kubernetes/kube-openapi/blob/0d48a347b3839510b3902c604b01a5c1de7aa36a/pkg/aggregator/aggregator_test.go, they do merge things, they do seem to use some reference and to work, I'm trying to see why because there's definitely DeelEqual being called.

The failure mode should be a panic, but even then, the tests should fail or the `swagger.json` should have some of these invalid `.._2` entries?

apelisse:
Actually you're right, it seems that DeepEqual fails.

liggitt:
Should we revert https://github.com/kubernetes/kubernetes/pull/114879 until this gets straightened out?

apelisse:
All the merge tests are still working correctly, the only tests that fails are the ones that depend on the fuzz generated swagger. So I suspect it's a red-herring, but I'm still not completely sure yet.

liggitt:
Are the fuzz tests exercising a rarely set field (ref, maybe, based on the comment above) that would fail if encountered in real life in a merge?

apelisse:
OK, I figured it out. I think when we're coming from the fuzzer, we parse the crazy utf-8 string as it is, which generate a RawFragment in the url.URL:
```
#/definitions/ƻR.ăŨơ嵣ƨ劒ʀd -> &jsonreference.Ref{referenceURL:(*url.URL)(0xc00013ccf0), referencePointer:jsonpointer.Pointer{referenceTokens:[]string{"definitions", "ƻR.ăŨơ嵣ƨ劒ʀd"}}, HasFullURL:false, HasURLPathOnly:false, HasFragmentOnly:true, HasFileScheme:false, HasFullFilePath:false} &url.URL{Scheme:"", Opaque:"", User:(*url.Userinfo)(nil), Host:"", Path:"", RawPath:"", OmitHost:false, ForceQuery:false, RawQuery:"", Fragment:"/definitions/ƻR.ăŨơ嵣ƨ劒ʀd", RawFragment:"/definitions/ƻR.ăŨơ嵣ƨ劒ʀd"}
```
The interesting part is that a `String()` of that Ref gives the following, escaped form:
```
#/definitions/%C6%BBR.%C4%83%C5%A8%C6%A1%E5%B5%A3%C6%A8%E5%8A%92%CA%80d
```

When we're serialize that to Json, and then deserialize that back into a spec.Swagger, we unmarshal the escaped form, which doesn't set the `RawFragment`. The two objects end-up being different from one another.

Re-reading the original body of this pull-request:
> This includes the removal of indirect dependency `github.com/PuerkitoBio/urlesc`

Yeah, it definitely has to do with url escaping ;-)

I don't think this is actually causing any problem in practice, but it's definitely inconsistent and error prone.
We have multiple options:
1. Change the spec merging code to use Diff instead of reflect.DeepEqual, hoping that there are no other occurences, not great
2. Revert this change and never update to a newer version of jsonreference, not great
3. Ignore the bug since it's unlikely to actually cause any problem, until it does, not great
4. Update jsonreference to basically to avoid the problem, I'm going to submit a patch right now.

liggitt:
thanks for digging

PR #347, opened 2023-01-04
Fix wrong extension name in parse schema

qingwave:
Fix wrong extension name in [document_v3.go](https://github.com/qingwave/kube-openapi/blob/f3cff1453715b47553c6543752d634081b0e9a38/pkg/util/proto/document_v3.go#L123)

cc  @alexzielenski @apelisse

### DIFF
--- a/pkg/util/proto/document_v3.go
+++ b/pkg/util/proto/document_v3.go
@@ -120,7 +120,7 @@ func (d *Definitions) ParseSchemaV3(s *openapi_v3.Schema, path *Path) (Schema, e
 	switch s.GetType() {
 	case object:
 		for _, extension := range s.GetSpecificationExtension() {
-			if extension.Name == "x-kuberentes-group-version-kind" {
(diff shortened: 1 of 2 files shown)


apelisse:
Thanks for the change, good catch! Can you check why this wasn't caught by a test? Are we light on the code coverage, could you please add a test?

qingwave:
> Thanks for the change, good catch! Can you check why this wasn't caught by a test? Are we light on the code coverage, could you please add a test?

@apelisse Good point, test added in new commit.

apelisse:
/lgtm
/approve

I'm curious, how did you find that out? I don't think anyone should be using this code, we're trying to deprecate it?

qingwave:
Hi @apelisse, when I read the [openapi doc](https://fossies.org/linux/kubernetes/api/openapi-spec/README.md), I find some special extensions like `x-kubernetes-group-version-kind`, and just search the extension in kubernetes,  find the issue coincidentally.

PR #346, opened 2023-01-03
Fix integration tests

olivierlemasle:
- Integration tests build "cmd/openapi-gen" so it must require its dependencies.
- Since integration tests are in a separate module, they were not run by CI.

---

Currently, running `go test ./test/...` (as CI does) does not test anything:

```
go: warning: "./test/..." matched no packages
no packages to test
```

And running `cd test/integration && go test .` fails with:

```
Running Suite: Integration Test Suite - /path/to/kube-openapi/test/integration
===========================================================================================
Random Seed: 1672749418

Will run 4 of 4 specs
------------------------------
[BeforeSuite] [FAILED] [0.055 seconds]
[BeforeSuite] 
/path/to/kube-openapi/test/integration/integration_suite_test.go:71

  Begin Captured GinkgoWriter Output >>
    STEP: building openapi-gen 01/03/23 13:36:58.61
  << End Captured GinkgoWriter Output

  Unexpected error:
      <*errors.errorString | 0xc000069f50>: {
[...]

PR #348, opened 2023-01-04
Fix GO-2022-0493 and GO-2022-1059

olivierlemasle:
Bump the following dependencies:
- golang.org/x/mod
- golang.org/x/net
- golang.org/x/sys
- golang.org/x/text
- golang.org/x/tools
- golang.org/x/xerrors

to fix:
- [GO-2022-0493](https://pkg.go.dev/vuln/GO-2022-0493)
- [GO-2022-1059](https://pkg.go.dev/vuln/GO-2022-1059)


liggitt:
looks like this made CI start failing at HEAD, not sure why (fix in https://github.com/kubernetes/kube-openapi/pull/353)

olivierlemasle:
@liggitt Yes, this PR should have been rebased after #346 was merged.
Before #346, the integration tests were not executed, so the `go mod tidy` was not checked.

Thanks for #353

PR #349, opened 2023-01-05
Protect against divide-by-zero panic with multipleOf

liggitt:
xref https://github.com/go-openapi/validate/issues/150

MultipleOf already returned errors when evaluating floats (since `x / 0.0` is Inf, which failed the `IsFloat64AJSONInteger` check), so making the int/uint versions return errors as well (instead of panics) was consistent.


apelisse:
Thanks
/lgtm
How did you find that out?

PR #351, opened 2023-01-06
Improve format of generated enum lines

liggitt:
Tweaks to improve formatting of generated enums.

Noticed when rebuilding generated go files in https://github.com/kubernetes/kubernetes/pull/114869

cc @apelisse


pkg/generators/openapi.go
```diff
@@ -686,7 +686,7 @@ func (g openAPITypeWriter) generateProperty(m *types.Member, parent *types.Type)
 		g.generateSimpleProperty(typeString, format)
 		if enumType, isEnum := g.enumContext.EnumType(m.Type); isEnum {
 			// original type is an enum, add "Enum: " and the values
-			g.Do("Enum: []interface{}{$.$}", strings.Join(enumType.ValueStrings(), ", "))
+			g.Do("Enum: []interface{}{$.$},\n", strings.Join(enumType.ValueStrings(), ", "))
```
apelisse, line 689: That's a lot less error-prone, thanks

PR #352, opened 2023-01-06
Revert content type change

liggitt:
Reverts https://github.com/kubernetes/kube-openapi/pull/333 since it breaks serving of the openapi doc in kubernetes (xref https://github.com/kubernetes/kubernetes/pull/114869#discussion_r1063439550 and https://github.com/kubernetes/kubernetes/pull/114869#issuecomment-1373127716)

The proto content type has an invalid subtype (`@` is not a valid character in mime subtypes) and gets rejected by the server when writing: https://github.com/kubernetes/kube-openapi/issues/354

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -178,9 +178,6 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 							return
 						}
 					}
-					contentType := accepts.Type + "/" + accepts.SubType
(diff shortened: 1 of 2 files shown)


liggitt:
After this merges and we recover the ability to serve, https://github.com/kubernetes/kube-openapi/issues/354 tracks the issue with our invalid subtype and a possible path forward

apelisse:
Thanks!
/lgtm

PR #355, opened 2023-01-06
Cleanup CI config

olivierlemasle:
- Update GitHub actions
- Remove duplicated entries in Go versions matrix:
  - '1.x' currently resolves to 'go1.19.4', just like '1.19'
  - `stable=false` option (which was [removed](https://github.com/actions/setup-go/pull/195) in actions/setup-go@v3) enabled the installation of pre-release Go versions (now possible without that option), but it does not install "tip" version of Go
  - Jobs '1.x' with gostable=true and gostable=false were therefore duplicates of '1.19'
- Use default checkout path and default GOPATH

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,24 +13,16 @@ jobs:
     strategy:
       matrix:
         go: [ '1.19', '1.18', '1.x' ]
-        gostable: [true]
-        include:
-        - go: '1.x'
-          gostable: false
     steps:
-    - uses: actions/checkout@v2
-      with:
-        path: src/k8s.io/kube-openapi/
+    - uses: actions/checkout@v3
 
     - name: Set up Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: "${{ matrix.go }}"
-        stable: ${{ matrix.gostable }}"
 
     - name: Build
       run: |
-        cd ${GITHUB_WORKSPACE}/src/k8s.io/kube-openapi/
         go mod tidy && git diff --exit-code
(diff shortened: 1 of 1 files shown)


Jefftree:
Thank you for fixing this!

cc @apelisse, I don't have as much context around the 1.x test. When go 1.20 is released, I'm assuming 1.x will default to the latest version? Wouldn't it still be useful to have that test around?

olivierlemasle:
If we want to always test with the latest Go version, we can also use [`stable` and `oldstable` aliases](https://github.com/actions/setup-go#using-stableoldstable-aliases) (they currently resolve to the latest 1.19.x and the latest 1.18.x version, but they'll be updated automatically when Go 1.20 is released)

.github/workflows/ci.yml
```diff
@@ -12,25 +12,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.19', '1.18', '1.x' ]
-        gostable: [true]
-        include:
-        - go: '1.x'
-          gostable: false
+        go: [ '1.19', '1.18']
```
apelisse, line 15: nit, but spaces are inconsistent. I agree with Jeff that I like having 1.x automatically pick the newest version. I'm not a big fan of using `stable` or `oldstable` since they don't necessarily match the policy that we have. Can we revert that line specifically?

olivierlemasle: Ok, I reverted that line.

I also don't like `stable` and `oldstable` very much; I prefer having control on the Go versions used by CI.
However, by keeping both `1.19` and `1.x`, we're wasting CI compute resource :seedling:

PR #359, opened 2023-01-13
move protomodels-specific code into separate file

alexzielenski:
cleans up code in various spots

moves proto models-specific SMD stuff into its own file, to prepare for direct conversion


apelisse (commented):
That seems to be mostly a move, very minor changes. I can't really diff the json/yaml files, what's up with the number of diffed lines?

alexzielenski:
>  I can't really diff the json/yaml files, what's up with the number of diffed lines?

I dont really know what to say about it other than a lot of stuff changed between 1.13 and 1.27

PR #339, opened 2022-12-13
Add direct conversion from spec.Schema to SMD schema.Schema

alexzielenski:
Intended to remove very expensive conversion via json & gnostic from apiserver.

PR #360, opened 2023-01-18
add doc to new openapi schemaconv api

alexzielenski:
followup to #339 adds discussed comment to new public api

also fixes removes normalization code added to proto_models.go and moves it to tests which care about it. It was causing unwanted changes from prior to the pr when using protomodels backend

/cc @apelisse


pkg/schemaconv/openapi.go
```diff
@@ -25,6 +25,12 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v4/schema"
 )
 
+// Expects a map from definition name to OpenAPI V3 structural schema for each definition.
```
apelisse, line 28: Shouldn't the comments start with `// ToSchemaFromOpenAPI ...`

apelisse: Does this accept an openapi v2? How are we supposed to use it for built-ins? Do we generate a structural schemas for these? That sounds like a waste.

apelisse:
> but I think we want to keep the diffs from before completely the same when using protomodels.

I don't know if I have a strong opinion either way

pkg/schemaconv/openapi.go
alexzielenski: Builtins are all OpenAPIV3. Usually they are same schema for v2 and v3. The builtins which need a separate v2 and v3 schema provide one. (There is a hack which stuffs the v2 schema inside the v3schema as an extension)

alexzielenski: added // ToSchemaFromOpenAPI ...

PR #362, opened 2023-01-19
Fix incorrect types OpenAPI V3

Jefftree:
Discovered as part of the OpenAPI V3 fuzzer tests, these fields are unused by k/k.

Explode should be a boolean (https://swagger.io/specification/#encoding-object).

SecurityRequirements object is not vendor extensible or refable.

Tests will be bundled with the fuzzer tests


pkg/spec3/operation.go
```diff
@@ -73,7 +73,7 @@ type OperationProps struct {
 	// Deprecated declares this operation to be deprecated
 	Deprecated bool `json:"deprecated,omitempty"`
 	// SecurityRequirement holds a declaration of which security mechanisms can be used for this operation
-	SecurityRequirement []*SecurityRequirement `json:"security,omitempty"`
+	SecurityRequirement []map[string][]string `json:"security,omitempty"`
```
apelisse, line 76: What's the reason behind that change? I guess you don't see any value in splitting vendor extensions from actual security requirements? I think it's unfortunate that one has to filter specifically the `x-...` when they are going through the list?

Jefftree: Security Requirements do not support vendor extensions so having a separate struct seems unnecessary.

PR #363, opened 2023-01-23
Move sanitize and fromMap to internal package

Jefftree:
These functions are used in unmarshaling from both OpenAPI v2 and OpenAPI v3 and need to be accessed from the `pkg/spec3` package in order to use the go-json-experiment v2 marshaler for OpenAPI V3.

/cc @apelisse 
/cc @liggitt
/cc @alexzielenski

### DIFF
--- a/pkg/internal/serialization.go
+++ b/pkg/internal/serialization.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+import (
+	"github.com/go-openapi/jsonreference"
+)
+
+// JSONRefFromMap populates a json reference object if the map v contains a $ref key.
+func JSONRefFromMap(jsonRef *jsonreference.Ref, v map[string]interface{}) error {
+	if v == nil {
+		return nil
+	}
+	if vv, ok := v["$ref"]; ok {
+		if str, ok := vv.(string); ok {
+			ref, err := jsonreference.New(str)
+			if err != nil {
+				return err
+			}
+			*jsonRef = ref
+		}
+	}
+	return nil
+}
+
+// SanitizeExtensions sanitizes the input map such that non extension
+// keys (non x-*, X-*) keys are dropped from the map. Returns the new
+// modified map, or nil if the map is now empty.
+func SanitizeExtensions(e map[string]interface{}) map[string]interface{} {
+	for k := range e {
+		if !IsExtensionKey(k) {
+			delete(e, k)
+		}
+	}
+	if len(e) == 0 {
+		e = nil
+	}
+	return e
+}
+
+// IsExtensionKey returns true if the input string is of format x-* or X-*
+func IsExtensionKey(k string) bool {
+	return len(k) > 1 && (k[0] == 'x' || k[0] == 'X') && k[1] == '-'
+}
--- a/pkg/validation/spec/header.go
+++ b/pkg/validation/spec/header.go
@@ -94,12 +94,8 @@ func (h *Header) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Dec
 
 	h.CommonValidations = x.CommonValidations
 	h.SimpleSchema = x.SimpleSchema
-	h.Extensions = x.Extensions
+	h.Extensions = internal.SanitizeExtensions(x.Extensions)
 	h.HeaderProps = x.HeaderProps
 
-	h.Extensions.sanitize()
-	if len(h.Extensions) == 0 {
-		h.Extensions = nil
-	}
 	return nil
 }
--- a/pkg/validation/spec/info.go
+++ b/pkg/validation/spec/info.go
@@ -89,17 +89,9 @@ func (e Extensions) GetObject(key string, out interface{}) error {
 	return nil
 }
 
-func (e Extensions) sanitize() {
-	for k := range e {
-		if !isExtensionKey(k) {
-			delete(e, k)
-		}
-	}
-}
-
 func (e Extensions) sanitizeWithExtra() (extra map[string]any) {
 	for k, v := range e {
-		if !isExtensionKey(k) {
+		if !internal.IsExtensionKey(k) {
 			if extra == nil {
 				extra = make(map[string]any)
 			}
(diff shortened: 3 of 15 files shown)


apelisse:
I'm really not a big fan, that's unfortunate. I don't know if we have another option :-/
@liggitt thoughts?

pkg/validation/spec/info.go
```diff
@@ -89,15 +89,20 @@ func (e Extensions) GetObject(key string, out interface{}) error {
 	return nil
 }
 
-func (e Extensions) sanitize() {
+// Sanitize sanitizes the Extensions map such that only keys with the
+// "x-" prefix are preserved
+func (e Extensions) Sanitize() {
```
liggitt, line 94: I'm not sure how hard we should work to avoid exporting these beyond this module, but a possible alternative to exporting these might be to put them in the internal package as functions and invoke them from here and the v3 package.

Jefftree: Yeah that was the alternative I was thinking of as well. We don't necessarily need the functions as a member of the corresponding types since they're not really accessing any private fields.

alexzielenski: If `sanitize` is the only motivator behind this reshuffling, I think -implementing it in both packages isn't terrible. It deletes keys prefixed with `x-`....

Jefftree: One drawback I see to that is it may break backwards compatibility since I  imagine `spec.Schema` and `spec3.OpenAPI` are imported in quite a few places in client libraries/etc.

pkg/internal/serialization.go
```diff
@@ -0,0 +1,46 @@
...
+limitations under the License.
+*/
+
+package internal
+
+import (
+	"github.com/go-openapi/jsonreference"
+)
+
+func JSONRefFromMap(v map[string]interface{}) (*jsonreference.Ref, error) {
```
apelisse, line 23: At first I was wondering why this returns a pointer to ref, but it's because you want to differentiate empty ref from missing ref. I don't think we need to differentiate? See my comment below.

pkg/validation/spec/ref.go
```diff
@@ -149,19 +151,12 @@ func (r *Ref) UnmarshalJSON(d []byte) error {
...
-	if vv, ok := v["$ref"]; ok {
-		if str, ok := vv.(string); ok {
-			ref, err := jsonreference.New(str)
-			if err != nil {
-				return err
-			}
-			*r = Ref{Ref: ref}
-		}
+	if jsonRef != nil {
+		*r = Ref{Ref: *jsonRef}
```
apelisse, line 159: Can't we just set it, even if the ref is empty? That's also more consistent with what we were doing?
```suggestion
	r.Ref = jsonRef
```

Jefftree: We're only setting the ref `if vv, ok := v["$ref"]; ok {` previously, so this behavior is consistent (although I also dislike that we're returning a pointer). We could make it semi-consistent by returning an empty jsonRef and dropping the pointer, resulting in an extra memory alloc (probably negligible?) every time the ref should be nil.

apelisse: Yeah, and the merge logic wants us to not unset it if it's unset. What about passing a pointer to the ref, like this:
```golang
err := internal.JSONRefFromMap(&r.Ref, v)
```

Jefftree: We would have a circular import reference between `validation/spec` and `internal` so we can't expose `spec.Ref` to the internal package.

Jefftree: Thanks for the suggestion, this looks much cleaner!

pkg/validation/spec/header.go
```diff
@@ -92,14 +92,14 @@ func (h *Header) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Dec
 		return err
 	}
 
+	internal.SanitizeExtensions(x.Extensions)
+	if len(x.Extensions) == 0 {
+		x.Extensions = nil
+	}
```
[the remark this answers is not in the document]
Jefftree: Extensions is aliased to `map[string]interface{}` which can be passed into sanitize because the types match. It seems like a pointer `*Extensions` is different from `*map[string]interface{}` and cannot be used directly. Are we okay with casting before the sanitize? I'm not exactly sure how much this would improve readability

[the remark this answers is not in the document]
apelisse: Yeah, that's not great, let's keep it as it is, thanks!

pkg/validation/spec/info.go
```diff
@@ -89,17 +89,9 @@ func (e Extensions) GetObject(key string, out interface{}) error {
...
-		if !isExtensionKey(k) {
-			delete(e, k)
-		}
-	}
-}
-
 func (e Extensions) sanitizeWithExtra() (extra map[string]any) {
 	for k, v := range e {
-		if !isExtensionKey(k) {
+		if !internal.IsExtensionKey(k) {
```
apelisse, line 94: The reason you keep `sanitizeWithExtra` is that it's not going to be needed in v3?

Jefftree: sanitizeWithExtra is only used with spec.Schema's ExtraProps.

pkg/internal/serialization.go
```diff
@@ -0,0 +1,53 @@
...
+
+// SanitizeExtensions sanitizes the input map such that non extension
+// keys (non x-*, X-*) keys are dropped from the map
+func SanitizeExtensions(e map[string]interface{}) {
+	for k := range e {
+		if !IsExtensionKey(k) {
+			delete(e, k)
+		}
+	}
+}
```
apelisse, line 48: OK What about this then ;-)
```suggestion
// SanitizeExtensions sanitizes the input map such that non extension
// keys (non x-*, X-*) keys are dropped from the map. Returns the new 
// modified map, or nil if the map is now empty.
func SanitizeExtensions(e map[string]interface{}) map[string]interface{} {
	for k := range e {
		if !IsExtensionKey(k) {
			delete(e, k)
		}
	}
	if len(e) == 0 {
	    e = nil
	}
	return e
}
```

And then everywhere you can do something like ...
```
p.Extensions = internal.SanitizeExtensions(x.Extensions)
```
it's a little misleading because sanitize doesn't make a new map, it modifies in place AND returns itself. I don't mind keeping it this way otherwise.

WDYT?

Jefftree: Agree on the slightly misleading part so we should be explicit in the documentation but I do like how concise it makes the unmarshal code. Updated

apelisse (commented):
Thanks!
/lgtm
/approve
/assign @liggitt

pkg/validation/spec/info.go
```diff
@@ -212,11 +200,7 @@ func (i *Info) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Decod
 	if err := opts.UnmarshalNext(dec, &x); err != nil {
 		return err
 	}
-	x.Extensions.sanitize()
-	if len(x.Extensions) == 0 {
-		x.Extensions = nil
-	}
-	i.VendorExtensible.Extensions = x.Extensions
+	i.Extensions = internal.SanitizeExtensions(x.Extensions)
```
liggitt, line 203: did this intentionally change from assigning to `i.VendorExtensible.Extensions`?

liggitt: answering my own question, it resolves the same and is not ambiguous

liggitt (commented):
/approve

PR #367, opened 2023-01-23
remove use of embed so CRDs aren't pulled into vendor

alexzielenski:
Does a few things from feedback when pulling into k8s:

- Remove todos in favor of issues
- Remove use of embed in test
- Make test blackbox instead of whitebox
- Ensure `RawExtension` and `Quantity` are the same for both protomodels and openapi implementations

/assign @apelisse

### DIFF
--- a/pkg/schemaconv/openapi_test.go
+++ b/pkg/schemaconv/openapi_test.go
@@ -17,8 +17,8 @@ limitations under the License.
 package schemaconv_test
 
 import (
-	"embed"
 	"encoding/json"
+	"os"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/schemaconv/openapi_test.go
```diff
@@ -53,8 +62,24 @@ func toPtrMap[T comparable, V any](m map[T]V) map[T]*V {
 }
 
 func normalizeTypeRef(tr *schema.TypeRef) {
-	if tr.Inlined == deducedDef.Atom {
-		// Deduplicate deducedDef (mostly for testing)
+	var untypedScalar schema.Scalar = "untyped"
```
apelisse, line 62: Why did you defined a couple at the top and this one here?

pkg/schemaconv/openapi_test.go
```diff
@@ -17,8 +17,8 @@ limitations under the License.
 package schemaconv_test
 
 import (
-	"embed"
```
apelisse, line 20: That's sad but yeah ...

pkg/schemaconv/openapi_test.go
alexzielenski: eh cuz its not used anywhere else. the others are

PR #368, opened 2023-01-24
Experimental unmarshal for OpenAPI V3

Jefftree:
Implement the experimental unmarshaler for OpenAPI V3.

Testing: `TestOpenAPIV3Deserialize` unmarshal the appsv1 and authorizationv1 OpenAPI specs via jsonv1 and jsonv2 and compares for equality.

Benchmarks:

Run `go test ./pkg/spec3/... -bench=BenchmarkOpenAPIV3Deserialize`

```
goos: linux
goarch: amd64
pkg: k8s.io/kube-openapi/pkg/spec3
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkOpenAPIV3Deserialize
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv1
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv1-24                        9         118930200 ns/op        17571548 B/op     208938 allocs/op
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2_via_jsonv1_v2_partial_optimized
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2_via_jsonv1_v2_partial_optimized-24                       10         100551076 ns/op        14432451 B/op     154751 allocs/op
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2_via_jsonv1_full_optimized
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2_via_jsonv1_full_optimized-24                             56          21943007 ns/op         5527198 B/op      27295 allocs/op
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2
BenchmarkOpenAPIV3Deserialize/appsv1spec.json_jsonv2-24                                                       75          14364996 ns/op         5516345 B/op      26968 allocs/op
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv1
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv1-24                                             136           8863784 ns/op         1484916 B/op      16969 allocs/op
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2_via_jsonv1_v2_partial_optimized
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2_via_jsonv1_v2_partial_optimized-24                     181           6520713 ns/op         1100177 B/op      10570 allocs/op
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2_via_jsonv1_full_optimized
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2_via_jsonv1_full_optimized-24                           648           1881193 ns/op          539557 B/op       2571 allocs/op
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2
BenchmarkOpenAPIV3Deserialize/authorizationv1spec.json_jsonv2-24                                                     969           1234902 ns/op          538057 B/op       2522 allocs/op
PASS
```

/cc @apelisse @alexzielenski

### DIFF
--- a/pkg/spec3/benchmark_serialization_test.go
+++ b/pkg/spec3/benchmark_serialization_test.go
@@ -111,7 +111,7 @@ func BenchmarkOpenAPIV3Deserialize(b *testing.B) {
 			internal.UseOptimizedJSONUnmarshalingV3 = true
 			for i := 0; i < b2.N; i++ {
 				var result *OpenAPI
-				if err := result.CustomUnmarshalJSON(originalJSON); err != nil {
+				if err := result.UnmarshalJSON(originalJSON); err != nil {
 					b2.Fatal(err)
 				}
 			}
--- a/pkg/spec3/benchmark_serialization_test.go
+++ b/pkg/spec3/benchmark_serialization_test.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"k8s.io/kube-openapi/pkg/internal"
-	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
@@ -112,7 +111,7 @@ func BenchmarkOpenAPIV3Deserialize(b *testing.B) {
 			internal.UseOptimizedJSONUnmarshalingV3 = true
 			for i := 0; i < b2.N; i++ {
 				var result *OpenAPI
-				if err := jsonv2.Unmarshal(originalJSON, &result); err != nil {
+				if err := result.CustomUnmarshalJSON(originalJSON); err != nil {
 					b2.Fatal(err)
 				}
 			}
--- a/pkg/spec3/benchmark_serialization_test.go
+++ b/pkg/spec3/benchmark_serialization_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"k8s.io/kube-openapi/pkg/internal"
+	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
@@ -105,16 +106,16 @@ func BenchmarkOpenAPIV3Deserialize(b *testing.B) {
 			}
 		})
 
-		// TODO: Enable this benchmark when jsonv2 is functional for OpenAPI V3.
-		// b.Run("jsonv2", func(b2 *testing.B) {
-		// b2.ReportAllocs()
-		// 	internal.UseOptimizedJSONUnmarshaling = true
-		// 	for i := 0; i < b2.N; i++ {
-		// 		var result OpenAPI
-		// 		if err := result.UnmarshalJSON(originalJSON); err != nil {
-		// 			b2.Fatal(err)
-		// 		}
-		// 	}
-		// })
+		b.Run("jsonv2", func(b2 *testing.B) {
+			b2.ReportAllocs()
+			internal.UseOptimizedJSONUnmarshaling = true
+			internal.UseOptimizedJSONUnmarshalingV3 = true
+			for i := 0; i < b2.N; i++ {
+				var result *OpenAPI
+				if err := jsonv2.Unmarshal(originalJSON, &result); err != nil {
+					b2.Fatal(err)
+				}
+			}
+		})
 	}
 }
--- a/pkg/spec3/example.go
+++ b/pkg/spec3/example.go
@@ -20,6 +20,9 @@ import (
 	"encoding/json"
 
 	"github.com/go-openapi/swag"
+	"k8s.io/kube-openapi/pkg/internal"
+	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
+
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
@@ -49,6 +52,9 @@ func (e *Example) MarshalJSON() ([]byte, error) {
 }
 
 func (e *Example) UnmarshalJSON(data []byte) error {
+	if internal.UseOptimizedJSONUnmarshalingV3 {
+		return jsonv2.Unmarshal(data, e)
+	}
 	if err := json.Unmarshal(data, &e.Refable); err != nil {
 		return err
 	}
@@ -58,6 +64,23 @@ func (e *Example) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &e.VendorExtensible); err != nil {
 		return err
 	}
+	return nil
+}
+
+func (e *Example) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Decoder) error {
+	var x struct {
+		spec.Extensions
+		ExampleProps
+	}
+	if err := opts.UnmarshalNext(dec, &x); err != nil {
+		return err
+	}
+	if err := internal.JSONRefFromMap(&e.Ref.Ref, x.Extensions); err != nil {
+		return err
+	}
+	e.Extensions = internal.SanitizeExtensions(x.Extensions)
+	e.ExampleProps = x.ExampleProps
+
 	return nil
 }
 
--- a/pkg/spec3/path.go
+++ b/pkg/spec3/path.go
@@ -87,7 +87,8 @@ func (p *Paths) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Deco
 	}
 	switch k := tok.Kind(); k {
 	case 'n':
-		return nil // noop
+		*p = Paths{}
+		return nil
 	case '{':
 		for {
 			tok, err := dec.ReadToken()
--- a/pkg/spec3/path.go
+++ b/pkg/spec3/path.go
@@ -18,9 +18,12 @@ package spec3
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/go-openapi/swag"
+	"k8s.io/kube-openapi/pkg/internal"
+	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
@@ -45,6 +48,9 @@ func (p *Paths) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON hydrates this items instance with the data from JSON
 func (p *Paths) UnmarshalJSON(data []byte) error {
+	if internal.UseOptimizedJSONUnmarshalingV3 {
+		return jsonv2.Unmarshal(data, p)
+	}
 	var res map[string]json.RawMessage
 	if err := json.Unmarshal(data, &res); err != nil {
 		return err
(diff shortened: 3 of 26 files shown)


pkg/spec3/serialization.go
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+package spec3
+
+import (
+	"github.com/go-openapi/jsonreference"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+func refFromMap(r *spec.Ref, v map[string]interface{}) error {
```
Jefftree, line 24: I duplicated these functions for now. Still thinking about the best way to handle these

alexzielenski: could some of them go into internal as Jordan suggested?

Jefftree: There's an import cycle with the internal package depending on the types defined in `spec/validation` and the types in the package depending on the function defined in the internal package. We've usually handled these cycles via interfaces but that probably wouldn't work in our scenario.

alexzielenski: Hmm I did not think of that. Maybe refFromMap could return jsonReference. Also `sanitizeExtensions` nor `postProcessExtensions` need to operate on `spec.Extensions`: I think that type is aliased to a `map[string]any`

Jefftree: Those are great suggestions, I'll give it a shot thanks!

pkg/spec3/path.go
```diff
@@ -74,6 +80,59 @@ func (p *Paths) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (p *Paths) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Decoder) error {
+	tok, err := dec.ReadToken()
+	if err != nil {
+		return err
+	}
+	var ext any
```
apelisse, line 88: why is ext defined here rather than in a more nested way?

pkg/spec3/path.go
```diff
@@ -74,6 +80,59 @@ func (p *Paths) UnmarshalJSON(data []byte) error {
...
+	switch k := tok.Kind(); k {
+	case 'n':
+		return nil // noop
```
apelisse, line 91: We probably had the same thing for v2, but are we supposed to leave `p` unaffected when we get a null value, or should we reset it? That's relevant when we re-deserialize into an existing non-empty struct. jsonv2 has stronger semantics than v1 that this should behave like a json-merge-patch.

From https://pkg.go.dev/github.com/go-json-experiment/json#UnmarshalOptions.UnmarshalNext:
> In general, unmarshaling follows merge semantics (similar to [RFC 7396](https://rfc-editor.org/rfc/rfc7396.html)) where the decoded Go value replaces the destination value for any JSON kind other than an object. For JSON objects, the input object is merged into the destination value where matching object members recursively apply merge semantics.

pkg/spec3/response.go
```diff
@@ -97,6 +105,61 @@ func (r *ResponsesProps) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (r *Responses) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Decoder) (err error) {
+	tok, err := dec.ReadToken()
+	if err != nil {
+		return err
+	}
+	var ext any
```
apelisse, line 113: Same comment as above for location of this var (if I'm not missing something)

apelisse (commented):
Looks good to me. Minor nits

pkg/spec3/benchmark_serialization_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"reflect"
+	"testing"
+
+	"k8s.io/kube-openapi/pkg/internal"
+	jsonv2 "k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
+)
+
+func TestOpenAPIV3Deserialize(t *testing.T) {
+	swagFile, err := os.Open("./testdata/appsv1spec.json")
```
apelisse, line 16: Have you written a fuzzer to complement the comparison here and make sure that we generate the same thing?

pkg/spec3/path.go
Jefftree: cc @alexzielenski, were you able to exercise this particular case in OpenAPI V2? I couldn't get a JSON example where this code path is exercised, setting "paths" to null in the JSON seems to bypass calling the UnmarshalNextJSON in this struct.

pkg/spec3/serialization.go
Jefftree: https://github.com/kubernetes/kube-openapi/pull/363 is updated based on the suggestions here

pkg/spec3/benchmark_serialization_test.go
Jefftree: Yes, have a separate PR for it

Jefftree: https://github.com/kubernetes/kube-openapi/pull/370 here's the fuzzer. I've tested this Unmarshal with the fuzzer and can make it a permanent test once some of these PRs are merged since there is quite a bit of a dependency tree with the PRs at the moment.

apelisse (commented):
Beside this one question about ignoring a couple of`null` values, looks good to me.
/assign @liggitt

pkg/spec3/path.go
alexzielenski: If you unmarshal directly into a `spec.Paths` struct with JSON `null` it will hit this line.

Thats a good point Antoine. I think we do want to reset `p`

Jefftree: Added resets for spec3.Paths and spec3.Responses. Should we make this change for the v2 unmarshaller as well?

apelisse: In a separate PR yes, it's relatively minor-ish (we never merge openapi specs that way ...) but we should do it for correctness.

apelisse:
Null and empty are not the same but I don't think you can avoid that so ... looks good to me.

liggitt (commented):
The round-trip unit tests in the `spec` package added in https://github.com/kubernetes/kube-openapi/pull/319#discussion_r986387086 gave a lot of confidence that json fields routed into the correct embedded structs. I think we need the same level of testing for these types.

pkg/spec3/example.go
```diff
@@ -61,6 +67,23 @@ func (e *Example) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (e *Example) UnmarshalNextJSON(opts jsonv2.UnmarshalOptions, dec *jsonv2.Decoder) error {
+	var x struct {
+		spec.Extensions
+		ExampleProps
```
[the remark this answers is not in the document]
liggitt: I cloned this PR, and if I comment out lines like `e.ExampleProps = x.ExampleProps` or `e.EncodingProps = x.EncodingProps` completely, all tests still pass, which doesn't give me a lot of confidence that we're round-tripping correctly (we *may* be, but we wouldn't know if we weren't)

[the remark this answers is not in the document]
apelisse: Order shouldn't matter, this works as a "catch what's left".
Agreed, as mentioned in https://github.com/kubernetes/kube-openapi/pull/368#discussion_r1089422141, #370 adds a fuzzer, we might want to merge it first.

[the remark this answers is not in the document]
liggitt: fuzzing is good... explicitly exercising fields from each of the embedded structs like the `spec` package unit tests do is also good

pkg/spec3/benchmark_serialization_test.go
```diff
@@ -105,16 +106,16 @@ func BenchmarkOpenAPIV3Deserialize(b *testing.B) {
...
-		// 		}
-		// 	}
-		// })
+		b.Run("jsonv2", func(b2 *testing.B) {
+			b2.ReportAllocs()
+			internal.UseOptimizedJSONUnmarshaling = true
+			internal.UseOptimizedJSONUnmarshalingV3 = true
+			for i := 0; i < b2.N; i++ {
+				var result *OpenAPI
+				if err := jsonv2.Unmarshal(originalJSON, &result); err != nil {
```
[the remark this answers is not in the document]
Jefftree: Thanks for bringing this up, yeah external callers shouldn't be able to access the internal json Unmarshal and a custom Unmarshal function should be implemented. I was having some trouble implementing the `UnmarshalJSON` function because we want to switch to jsonv2 if a flag is enabled, but continue with jsonv1 otherwise. Implementing a custom unmarshaler bypasses the default jsonv1 unmarshalling and I didn't find an easy way to call it (Ideally something like a `MarshalNoCustom()` kind of function similar to the `FuzzNoCustom()` in gofuzz). The temporary workaround is to name the function something other than `UnmarshalJSON` (currently set as `CustomUnmarshalJSON`) though I'm not really a fan of doing it this way. I'll keep looking into this

```
func (o *OpenAPI) CustomUnmarshalJSON(data []byte) error {
    if internal.UseOptimizedJSONUnmarshalingV3 {
        return jsonv2.Unmarshal(data, &o)
    }
    return json.Unmarshal(data, &o)
}
```

pkg/spec3/example.go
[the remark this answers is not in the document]
Jefftree: Added round trip tests for embedded structs, +1 on getting the fuzzer merged first and integrating a test with the fuzzer + different unmarshal flag combinations.

pkg/spec3/benchmark_serialization_test.go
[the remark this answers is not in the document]
alexzielenski: https://github.com/kubernetes/kube-openapi/blob/39b3636cbaa32f381a5e7090cda15c71f38dd993/pkg/validation/spec/tag.go#L56-L65

If the flag is on, then we switch to jsonv2.

Otherwise, we call the jsonv1 version on all the embedded structs. Embedded structs were things like `TagProps` which do not have custom marshaling functions.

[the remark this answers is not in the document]
Jefftree: The prerequisite is the struct must be an embedded struct though right? For a non embedded struct, this unmarshal process would have to be repeated on all fields? https://github.com/kubernetes/kube-openapi/blob/master/pkg/spec3/spec.go#L24-L37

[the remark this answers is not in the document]
alexzielenski: Here is a trick that can work for the case you linked:

```go
func (o *OpenAPI) UnmarshalJSON(data []byte) error {
        if internal.UseOptimizedJSONUnmarshaling { 
 		return jsonv2.Unmarshal(data, t) 
 	} 
	type OpenAPIWithNoFunctions OpenAPI
	var me *OpenAPIWithNoFunctions = (*OpenAPIWithNoFunctions)(o)
	return json.Unmarshal(data, me)
}
```

I think this might have a penalty performing syntax validation on the input twice.

Passed this test:

```go
func TestUnmarshalOpenAPI(t *testing.T) {
	actual := &spec3.OpenAPI{}
	text := `{"openapi":"3.0.0", "info": {"title":"Hello, world!"}}`
	expect := &spec3.OpenAPI{
		Version: "3.0.0",
		Info: &spec.Info{
			InfoProps: spec.InfoProps{
				Title: "Hello, world!",
			},
		},
	}

	require.NoError(t, json.Unmarshal([]byte(text), actual))
	require.Equal(t, expect, actual)
}
```

apelisse:
I think it's lgtm for me,
/approve

I'd appreciate if you give another look @liggitt

apelisse:
Reformatting the results up-top for clarity:
```
name                                              old time/op    new time/op    delta
OpenAPIV3Deserialize/appsv1spec.json-24              119ms ± 0%      14ms ± 0%  -87.92%  (p=0.008 n=5+5)
OpenAPIV3Deserialize/authorizationv1spec.json-24    8.86ms ± 0%    1.23ms ± 0%  -86.07%  (p=0.008 n=5+5)

name                                              old alloc/op   new alloc/op   delta
OpenAPIV3Deserialize/appsv1spec.json-24             17.6MB ± 0%     5.5MB ± 0%  -68.61%  (p=0.008 n=5+5)
OpenAPIV3Deserialize/authorizationv1spec.json-24    1.48MB ± 0%    0.54MB ± 0%  -63.77%  (p=0.008 n=5+5)

name                                              old allocs/op  new allocs/op  delta
OpenAPIV3Deserialize/appsv1spec.json-24               209k ± 0%       27k ± 0%  -87.09%  (p=0.008 n=5+5)
OpenAPIV3Deserialize/authorizationv1spec.json-24     17.0k ± 0%      2.5k ± 0%  -85.14%  (p=0.008 n=5+5)
```

PR #370, opened 2023-01-27
Add fuzzer for OpenAPI V3

Jefftree:
Marking as WIP because it may still be incomplete, most of the important things should be there though.

/cc @apelisse @alexzielenski

### DIFF
--- a/pkg/spec3/benchmark_serialization_test.go
+++ b/pkg/spec3/benchmark_serialization_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	fuzz "github.com/google/gofuzz"
 	"k8s.io/kube-openapi/pkg/internal"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
@@ -18,6 +19,29 @@ import (
 var swaggerDiffOptions = []cmp.Option{cmp.Comparer(func(a spec.Ref, b spec.Ref) bool {
 	return a.String() == b.String()
 })}
+
+func TestOpenAPIV3RoundTrip(t *testing.T) {
+	var fuzzer *fuzz.Fuzzer
+	fuzzer = fuzz.NewWithSeed(1646791953)
+	// Make sure we have enough depth such that maps do not yield nil elements
+	fuzzer.MaxDepth(22).NilChance(0.5).NumElements(1, 7)
+	fuzzer.Funcs(OpenAPIV3FuzzFuncs...)
+	expected := &OpenAPI{}
+	fuzzer.Fuzz(expected)
+
+	j, err := json.Marshal(expected)
+	if err != nil {
+		t.Fatal(err)
+	}
+	var actual *OpenAPI
+	err = json.Unmarshal(j, &actual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !reflect.DeepEqual(expected, actual) {
+		t.Fatal(cmp.Diff(expected, actual, swaggerDiffOptions...))
+	}
+}
 
 func TestOpenAPIV3Deserialize(t *testing.T) {
 	swagFile, err := os.Open("./testdata/appsv1spec.json")
--- a/pkg/spec3/fuzz.go
+++ b/pkg/spec3/fuzz.go
@@ -1,0 +1,250 @@
+package spec3
+
+import (
+	fuzz "github.com/google/gofuzz"
+	"math/rand"
+	"strings"
+
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// refChance is the chance that a particular component will use a $ref
+// instead of fuzzed. Expressed as a fraction 1/n, currently there is
+// a 1/3 chance that a ref will be used.
+const refChance = 3
+
+const alphaNumChars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+func randAlphanumString() string {
+	arr := make([]string, rand.Intn(10)+5)
+	for i := 0; i < len(arr); i++ {
+		arr[i] = string(alphaNumChars[rand.Intn(len(alphaNumChars))])
+	}
+	return strings.Join(arr, "")
+}
+
+var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
+	func(s *string, c fuzz.Continue) {
+		// All OpenAPI V3 map keys must follow the corresponding
+		// regex. Note that this restricts the range for all other
+		// string values as well.
+		str := randAlphanumString()
+		*s = str
+	},
+	func(o *OpenAPI, c fuzz.Continue) {
+		c.FuzzNoCustom(o)
+		o.Version = "3.0.0"
+	},
+	func(r *interface{}, c fuzz.Continue) {
+		switch c.Intn(3) {
+		case 0:
+			*r = nil
+		case 1:
+			n := c.RandString() + "x"
+			*r = n
+		case 2:
+			n := c.Float64()
+			*r = n
+		}
+	},
+	func(v **spec.Info, c fuzz.Continue) {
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -6,13 +6,15 @@ require (
 	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/emicklei/go-restful/v3 v3.8.0
+	github.com/getkin/kin-openapi v0.76.0
```
Jefftree, line 9: I recall that we did a fair bit of work to ensure that kin-openapi isn't getting imported in k/k. Should the fuzzer testing be moved into integration instead?

pkg/spec3/fuzz_test.go
```diff
@@ -0,0 +1,28 @@
+package spec3
+
+import (
+	"encoding/json"
+
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
```
apelisse, line 8: I could drop this test and get rid of the dependency. Why do you work so hard trying to make sure that the fuzzer generates entirely valid openapi?

apelisse:
I'd love @alexzielenski to make a first pass. He's a lot more familiar with this than I am.

pkg/spec3/fuzz_test.go
```diff
@@ -0,0 +1,28 @@
+package spec3
```
alexzielenski, line 1: can this be blackbox test?

alexzielenski (changes requested):

pkg/spec3/response.go
```diff
@@ -78,20 +78,29 @@ func (r ResponsesProps) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON unmarshals responses from JSON
 func (r *ResponsesProps) UnmarshalJSON(data []byte) error {
-	var res map[string]*Response
+	var res map[string]json.RawMessage
```
Jefftree, line 81: This change will be removed here and included in a separate PR: https://github.com/kubernetes/kube-openapi/pull/372

pkg/spec3/benchmark_serialization_test.go
```diff
@@ -19,6 +20,29 @@ var swaggerDiffOptions = []cmp.Option{cmp.Comparer(func(a spec.Ref, b spec.Ref)
...
+	}
+	var actual *OpenAPI
+	err = json.Unmarshal(j, &actual)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !reflect.DeepEqual(expected, actual) {
+		t.Fatal(cmp.Diff(expected, actual, swaggerDiffOptions...))
+	}
+}
```
apelisse, line 44: Should we also check that marshaling is stable too?

pkg/spec3/fuzz.go
```diff
@@ -0,0 +1,250 @@
+package spec3
+
+import (
+	fuzz "github.com/google/gofuzz"
+	"github.com/lucasjones/reggen"
```
apelisse, line 5: Can we add this dependency? That seems unfortunate. Any other way to generate valid arbitrary strings?

Jefftree: It's a single go file https://github.com/lucasjones/reggen/blob/master/reggen.go, if it's not a desired dependency, I can write a custom function to generate valid arbitrary strings.

pkg/spec3/benchmark_serialization_test.go
Jefftree: Isn't that something that's given based on the conditions here?

We don't have a OpenAPI JSON fuzzer (and I don't think we should build one) so the base case starts from a fuzzed OpenAPI that's marshalled. The unmarshal test above proves x == Unmarshal(Marshal(x)) where x is a fuzzed OpenAPI. Writing a new test to for Unmarshal is trying to determine the relationship between Marshal(x) and Marshal(Unmarshal(Marshal(x))). Since we already showed x == Unmarshal(Marshal(x)), the latter simplifies to Marshal(x). The two would only not equal if Marshal is not stable and wouldn't exactly be a roundtrip test? 

When we implement the experimental marshaler, then it makes sense to add serialization tests to ensure that the old and new marshaler have the same behavior. WDYT?

pkg/spec3/benchmark_serialization_test.go
```diff
@@ -19,6 +20,29 @@ var swaggerDiffOptions = []cmp.Option{cmp.Comparer(func(a spec.Ref, b spec.Ref)
...
+	var fuzzer *fuzz.Fuzzer
+	fuzzer = fuzz.NewWithSeed(1646791953)
+	// Make sure we have enough depth such that maps do not yield nil elements
+	fuzzer.MaxDepth(22).NilChance(0.5).NumElements(1, 7)
+	fuzzer.Funcs(OpenAPIV3FuzzFuncs...)
+	expected := &OpenAPI{}
+	fuzzer.Fuzz(expected)
+
+	j, err := json.Marshal(expected)
+	if err != nil {
```
alexzielenski, line 33: I turned down the nil chance, and ran this test in a loop with a few random seeds, and found something is up with `Paths` that doe snot round trip properly

alexzielenski (changes requested):

pkg/spec3/benchmark_serialization_test.go
```diff
@@ -19,6 +20,29 @@ var swaggerDiffOptions = []cmp.Option{cmp.Comparer(func(a spec.Ref, b spec.Ref)
 	return a.String() == b.String()
 })}
 
+func TestOpenAPIV3RoundTrip(t *testing.T) {
```
alexzielenski, line 23: https://github.com/kubernetes/kube-openapi/blob/f73e7dbcfe2954706d8ee5db7f1adaed3950e8de/pkg/util/jsontesting/json_roundtrip.go#L28-L44

might be suitable to use here

Jefftree: To conform to the interface, Unmarshal is implemented in https://github.com/kubernetes/kube-openapi/pull/368, can I fix this in a follow up?

pkg/spec3/benchmark_serialization_test.go
Jefftree: Fixed. Accidentally removed the requirement that paths needed to start with a slash in a refactor.

alexzielenski (changes requested):
only saw small thing, then lgtm

pkg/spec3/fuzz.go
```diff
@@ -0,0 +1,253 @@
...
+			return
+		}
+		c.Fuzz(&v.VendorExtensible)
+		c.Fuzz(&v.Description)
+		c.Fuzz(&v.Nullable)
+		c.Fuzz(&v.Title)
+		c.Fuzz(&v.Required)
+		c.Fuzz(&v.ExternalDocs)
+		n := c.Intn(8)
+		switch n {
```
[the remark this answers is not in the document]
alexzielenski: That makes sense, I did not realize.

pkg/spec3/fuzz.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		for i := 0; i < numChildren; i++ {
+			v.Add("x-"+randAlphanumString(), c.RandString()+"x")
+		}
+	},
+	func(v *spec.ExternalDocumentation, c fuzz.Continue) {
+		c.Fuzz(&v.Description)
+		v.URL = "https://" + randAlphanumString()
+	},
+	func(v *spec.Schema, c fuzz.Continue) {
```
apelisse, line 185: I have only checked quickly but this is missing a lot of fields, which means that if these fields are not round-tripped properly, we're very likely to miss that in the tests.

Jefftree: Let me see if we can reuse the fuzzer that Alex has already wrote for this.

PR #371, opened 2023-02-02
Use smarter cache

apelisse:
Use the previously introduced smarter cache in the OpenAPI handlers. Since it's fairly easy to use, it allows us to now cache the cache busting root document for OpenAPI v3 which we couldn't cache before.

cc @alexzielenski @Jefftree @lavalamp


pkg/handler/handler.go
```diff
@@ -50,77 +51,79 @@ func computeETag(data []byte) string {
...
 }
 
 // NewOpenAPIService builds an OpenAPIService starting with the given spec.
-func NewOpenAPIService(spec *spec.Swagger) (*OpenAPIService, error) {
+func NewOpenAPIService(swagger *spec.Swagger) (*OpenAPIService, error) {
+	return NewOpenAPIServiceLazy(cached.NewResultOK(swagger, uuid.New().String()))
+}
+
+// NewOpenAPIServiceLazy builds an OpenAPIService from lazy spec.
+func NewOpenAPIServiceLazy(swagger cached.Data[*spec.Swagger]) (*OpenAPIService, error) {
```
Jefftree, line 69: Can we ever return an error here?

apelisse: Good catch, we can't, thanks.

apelisse: It means the one above can't return an error either.

apelisse: Fixed, thanks

pkg/handler3/handler.go
```diff
@@ -92,94 +134,91 @@ func constructServerRelativeURL(gvString, etag string) string {
...
-	if err != nil {
-		return nil, err
-	}
-	return proto.Marshal(document)
+func (o *OpenAPIService) DeleteGroupVersion(group string) {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+	delete(o.v3Schema, group)
+	// Rebuild the merge cache map since the items have changed.
+	o.discoveryCache.Replace(o.buildDiscoveryCacheLocked())
```
alexzielenski, line 206: I feel like there is a missed opportunity here to prepend the `discoveryCache` pipeline with the v3Schema somehow to avoid having to remember to update discoveryCache (like etag for v3Schema changes as you add/remove/update), but I haven't thought it though completely.

pkg/handler3/handler.go
```diff
@@ -92,94 +134,91 @@ func constructServerRelativeURL(gvString, etag string) string {
...
-	for k := range o.v3Schema {
-		keys[i] = k
-		i++
+func (o *OpenAPIService) buildDiscoveryCacheLocked() cached.Data[[]byte] {
+	caches := make(map[string]cached.Data[[]byte], len(o.v3Schema))
+	for gvName, group := range o.v3Schema {
+		caches[gvName] = group.jsonCache
 	}
-
-	sort.Strings(keys)
```
alexzielenski, line 111: This `keys` array was unused. nice cleanup

alexzielenski (commented):
like how it simplifies control a bit by removing unused error case from NewOpenAPIService/UpdateSpec

have some Q's about things I noticed

pkg/handler/handler.go
```diff
@@ -50,77 +51,79 @@ func computeETag(data []byte) string {
...
-		json, err := o.jsonCache.Get()
-		if err != nil {
-			return nil, err
-		}
-		return []byte(computeETag(json)), nil
-	})
-	o.lastModified = time.Now()
-
+func (o *OpenAPIService) UpdateSpec(swagger *spec.Swagger) error {
+	o.UpdateSpecLazy(cached.NewResultOK(swagger, uuid.New().String()))
```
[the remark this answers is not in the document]
apelisse: That UUID here basically means "anytime someone calls this function, re-compute the json to see if it changed", i.e. invalidate the json-cache since something might have changed.

pkg/handler3/handler.go
apelisse: I think I see what you mean, let me try to consider it

pkg/handler/handler.go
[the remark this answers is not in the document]
alexzielenski: Gotcha. I see comptueEtag used in jsonCache now:

https://github.com/kubernetes/kube-openapi/pull/371/files#diff-47683fcfa1b9a283cbc534ece58037debb4ceefd6584606550485d55c10e4dd9R88

resolved

pkg/handler/handler.go
```diff
@@ -50,77 +51,79 @@ func computeETag(data []byte) string {
...
-		return nil, "", time.Time{}, err
-	}
-	etagBytes, err := o.etagCache.Get()
-	if err != nil {
-		return nil, "", time.Time{}, err
-	}
-	return specPb, string(etagBytes), o.lastModified, nil
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+	result := o.protoCache.Get()
```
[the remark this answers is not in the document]
apelisse: Done, great suggestion!

pkg/handler/handler.go
```diff
@@ -50,77 +51,79 @@ func computeETag(data []byte) string {
...
-		return nil, err
-	}
-	return o, nil
+	o.UpdateSpecLazy(swagger)
+
+	// Use a pass-through transformer that does nothing but record
+	// the time of last modification, since this is only called when
+	// the jsonCache is updated.
+	timeUpdate := cached.NewTransformer[*spec.Swagger](func(result cached.Result[*spec.Swagger]) cached.Result[*spec.Swagger] {
+		o.lastModified = time.Now()
```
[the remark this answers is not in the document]
apelisse: > I think this is semantically different from the original implementation, but still faithful to the requirements of Last-Modified HTTP Header for caching,

I agree with all of that.

Rest of the comment addressed, thanks!

pkg/handler3/handler.go
apelisse: Yeah, I think we need the lock on `v3Schema` unfortunately (and I can't think of a way around it), and we need to invalidate some cache anytime the dict changes anyway. Though to be fair, I think this is still absolutely great:
1. If you change the dict itself, it'll regenerate the top-level document, but will not recompute anything for sub-documents that haven't changed
2. If any of the sub-document changes, this will invalidate the top-level document, and all other sub-documents are not recomputed.
Basically, only the things that need to be re-computed will ever be, it's as good as it gets.

alexzielenski (commented):
minor comment but lgtm otherwise

pkg/handler3/handler.go
```diff
@@ -57,23 +57,63 @@ type OpenAPIV3DiscoveryGroupVersion struct {
...
-	rwMutex      sync.RWMutex
+func ToV3ProtoBinary(json []byte) ([]byte, error) {
+	document, err := openapi_v3.ParseDocument(json)
+	if err != nil {
+		return nil, err
+	}
+	return proto.Marshal(document)
+}
+
+type TimedSpec struct {
```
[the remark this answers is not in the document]
apelisse: No, good call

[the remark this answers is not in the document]
apelisse: ran `s/TimedSpec/timedSpec/`, thanks

pkg/handler3/handler_test.go
```diff
@@ -64,7 +64,7 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 	}
 
 	mux := http.NewServeMux()
-	o, err := NewOpenAPIService(nil)
```
alexzielenski, line 67: I checked the test suite for this and it does not seem to flex the mechanism for changing a groupversions content at all in the tests. Seeing as this PR pertains most to that functionality, do you think you can add some?

pkg/handler/handler_test.go
```diff
@@ -40,11 +40,8 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
 	}
 
 	mux := http.NewServeMux()
-	o, err := NewOpenAPIService(&s)
```
alexzielenski, line 43: This also doesn't test `UpdateSpec`/`UpdateSpecLazy` codepath

alexzielenski (changes requested):
I think it needs better test coverage, actually

apelisse:
@alexzielenski @seans3 Testing the OpenAPI v3 handler would be a lot simpler with a client, what do you guys think we move some of the client logic in here?

apelisse:
Added even more test, can I haz a review?

pkg/handler3/handler_test.go
```diff
@@ -432,3 +434,118 @@ func TestCacheBusting(t *testing.T) {
 		}
 	}
 }
+
+func MarshalOrDie(spec *spec3.OpenAPI) []byte {
```
Jefftree, line 438: is this used anywhere?

pkg/handler3/handler_test.go
```diff
@@ -432,3 +434,118 @@ func TestCacheBusting(t *testing.T) {
...
+	j, err := json.Marshal(spec)
+	if err != nil {
+		panic(err)
+	}
+	return j
+
+}
+
+func OpenAPIOrDie(name string) *spec3.OpenAPI {
```
Jefftree, line 447: Private function here should be fine?

pkg/handler3/handler_test.go
```diff
@@ -432,3 +434,118 @@ func TestCacheBusting(t *testing.T) {
...
+	}
+	return discovery, resp.Header.Get("etag"), nil
+}
+
+func TestUpdateSpecLazy(t *testing.T) {
+	mux := http.NewServeMux()
+	o := NewOpenAPIService()
+
+	mux.Handle("/openapi/v3", http.HandlerFunc(o.HandleDiscovery))
+	mux.Handle("/openapi/v3/apis/apps/v1", http.HandlerFunc(o.HandleGroupVersion))
```
Jefftree, line 493: I don't see anywhere this path is being requested. I'd assume either we'd want to check the contents of the path (as opposed to just discovery), or this line can probably be removed since the contents of the path is not of interest in the current test.

pkg/handler3/handler_test.go
```diff
@@ -432,3 +434,118 @@ func TestCacheBusting(t *testing.T) {
...
+	}
+
+	discovery := &OpenAPIV3Discovery{}
+	if err := json.Unmarshal(body, &discovery); err != nil {
+		return nil, "", fmt.Errorf("failed to unmarshal discovery: %v", err)
+	}
+	return discovery, resp.Header.Get("etag"), nil
+}
+
+func TestUpdateSpecLazy(t *testing.T) {
```
Jefftree, line 488: Contrary to the v2 test, only `UpdateGroupVersion` is called. If this is intended, can we update the name?

pkg/handler/handler_test.go
```diff
@@ -116,6 +115,84 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
+func TestUpdateSpecLazy(t *testing.T) {
+	returnedJSON := normalizeSwaggerOrDie(returnedSwagger)
+	var s spec.Swagger
+	err := s.UnmarshalJSON(returnedJSON)
+	if err != nil {
+		t.Errorf("Unexpected error in unmarshalling SwaggerJSON: %v", err)
+	}
+
+	mux := http.NewServeMux()
+	o := NewOpenAPIService(&s)
```
Jefftree, line 167: Do we need a test for NewOpenAPIServiceLazy?

pkg/handler/handler_test.go
```diff
@@ -116,6 +115,84 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
+		panic(fmt.Errorf("Unexpected error in unmarshalling SwaggerJSON: %v", err))
+	}
+	rj, err := json.Marshal(s)
+	if err != nil {
+		panic(fmt.Errorf("Unexpected error in preparing returnedJSON: %v", err))
+	}
+	return rj
+}
+
+func TestUpdateSpecLazy(t *testing.T) {
```
Jefftree, line 158: This is great, thanks!

pkg/handler/handler_test.go
```diff
@@ -116,6 +115,84 @@ func TestRegisterOpenAPIVersionedService(t *testing.T) {
...
+
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		panic(fmt.Errorf("Unexpected error in reading response body: %v", err))
+	}
+	return string(body)
+}
+
+func normalizeSwaggerOrDie(j []byte) []byte {
```
Jefftree, line 145: Thanks for refactoring this out. Can this be reused in the test above too? L26-35 I think

Jefftree (commented):
some minor non-blocking comments. LGTM from me

pkg/handler3/handler_test.go
apelisse: Yeah, I mostly wanted to see how the discovery is behaving. I'm assuming this endpoint is already tested by the test above.

pkg/handler/handler_test.go
apelisse: We could but it's not really doing much :-) I'll see if I can add one.

pkg/handler3/handler_test.go
apelisse: I've taken the line out.

pkg/handler/handler_test.go
apelisse: NewOpenAPIService actually is basically the same thing as NewOpenAPIServiceLazy so, I don't see a lot of value.

apelisse:
Updated, PTAL!

pkg/cached/cache_test.go
```diff
@@ -622,6 +622,14 @@ func TestReplaceable(t *testing.T) {
 	}
 }
 
+func TestReplaceableDifferentType(t *testing.T) {
+	replaceable := cached.Replaceable[bool]{}
+	replaceable.Replace(cached.NewSource(func() cached.Result[bool] {
+		return cached.NewResultOK(false, "hash")
+	}))
+	replaceable.Replace(cached.NewResultOK(true, "hash2"))
```
lavalamp, line 630: maybe test that results are as expected and not just that it doesn't crash?

apelisse: fixed

lavalamp:
I don't see anything wrong but it does make me nervous that this much of the handler is not in the main repo and we don't get to see e2e test results until we vendor this :/

apelisse:
I'm just vendoring this into kubernetes right now to make a test so let's see how this goes before we merge

lavalamp:
awesome, thanks!

apelisse:
https://github.com/kubernetes/kubernetes/pull/116349

This is just updating to use that PR. It doesn't have the change to actually make it useful. The change to do that is mostly ready, I need to rebase it, I'll push later, either on the same PR or on a different PR.

apelisse:
Passed! (except the dependency check, which is expected since I ran against my own copy of kube-openapi).

alexzielenski:
/lgtm

/hold
in case you want more eyes

apelisse:
/hold cancel

Thanks!

PR #372, opened 2023-02-07
Fix json unmarshal bug with extensions and Responses

Jefftree:
Almost a duplicate of https://github.com/kubernetes/kube-openapi/pull/320, for OpenAPI V3 instead of V2.

/assign @apelisse @alexzielenski


apelisse:
/lgtm
/approve
Thanks!

PR #374, opened 2023-02-16
Update go-json-experiment to latest version

Jefftree:
Update go-json-experiment to latest version. Also create a `hack/update-json-library.sh` script for updating this in the future.

This also drops support for go 1.18

/cc @apelisse @alexzielenski @liggitt


hack/update-json-library.sh
```diff
@@ -0,0 +1,25 @@
+# This script can be called via ./hack/update-json-library.sh` to update the
+# go-json-experiment fork in the repo. 
+
+GO_JSON_EXPERIMENT_DIR="pkg/internal/third_party/go-json-experiment/json"
+rm -rf $GO_JSON_EXPERIMENT_DIR
+git clone https://github.com/go-json-experiment/json $GO_JSON_EXPERIMENT_DIR
```
alexzielenski, line 6: can use `-depth 1` for shallow clone

pkg/internal/third_party/go-json-experiment/json/example_orderedobject_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"log"
 	"reflect"
 
-	"k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json"
+	"github.com/go-json-experiment/json"
```
liggitt, line 12: all of these need to be rewritten to our imported package or we'll take a dependency on the external repo (as showed up in the CI go.mod verification failure)

alexzielenski (commented):
do we also need a presubmit check that this directory remains equal to the output of this script?

apelisse:
That seems expensive, and the CI will break anytime they update something, so we would depend on something outside our control?

alexzielenski:
At the very least I mean something to ensure this dir doesn't accidentally get changed. I dont mean we should track HEAD.

apelisse:
Sounds like we could use the hash saved in the file to check that the dir maps to the given hash.

Also, we could have a daily/weekly script run and try the most recent of repo to see if it still works.

Jefftree:
it looks like go-json-experiment now has a dependency on go 1.19 (https://github.com/go-json-experiment/json/commit/ef074b63cd94c0c0f523f4312bc4140a749f0e85) so the go 1.18 CI tests are failing.

liggitt:
> This PR also drops support for go 1.18 since Kubernetes v1.23+ are all on go 1.19.

they are at HEAD of release branches, but still pass unit and integration tests on their original go versions (go1.17 for v1.23, go1.18 for v1.24), and we wouldn't want to backport a change to k8s release branches that break people using kubernetes libraries with their original go version on a patch release. 

That said, if we had to backport a kube-openapi change to 1.23 (unlikely, since 1.23 hits EOL in a couple weeks) or 1.24, we could make a kube-openapi branch prior to this commit

hack/update-json-library.sh
```diff
@@ -0,0 +1,27 @@
+# This script can be called via ./hack/update-json-library.sh` to update the
```
liggitt, line 1: add `#!/usr/bin/env bash`?

hack/update-json-library.sh
```diff
@@ -0,0 +1,27 @@
...
+# go-json-experiment fork in the repo. 
+
+GO_JSON_EXPERIMENT_DIR="pkg/internal/third_party/go-json-experiment/json"
+rm -rf $GO_JSON_EXPERIMENT_DIR
+git clone --depth 1 https://github.com/go-json-experiment/json $GO_JSON_EXPERIMENT_DIR
+cd $GO_JSON_EXPERIMENT_DIR
+HASH=$(git rev-parse HEAD)
```
liggitt, line 8: would it make more sense to set the hash we consume in this script, rather than always taking HEAD?

that would also mean that a presubmit could just run the script and check that there's no diff to verify

liggitt: I don't have a sense for whether the go-json-experiment repo always treats HEAD as stable / releasable, or if we might be grabbing a series of commits at an unfortunate point in time.

Specifying the hash we grab seems good for that reason, and I'd recommend getting an ack from dsnet when we're thinking of bumping.

Jefftree: Added the hash as a variable in the script. Definitely agree we should get an ack from dsnet on version bumps, I'm also imagining this script will be temporary until `go-json-experiment` becomes stable.

liggitt: Nothing is so permanent as a temporary solution :-D

hack/update-json-library.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+# This script can be called via ./hack/update-json-library.sh` to update the
+# go-json-experiment fork in the repo. 
+
+# HASH should be updated to the desired commit ID in https://github.com/go-json-experiment/json.
+HASH="540f014424240312547dcccddf11a8229ca0f463"
+
+GO_JSON_EXPERIMENT_DIR="pkg/internal/third_party/go-json-experiment/json"
+rm -rf $GO_JSON_EXPERIMENT_DIR
+git clone --depth 1 https://github.com/go-json-experiment/json $GO_JSON_EXPERIMENT_DIR
```
liggitt, line 10: I think `--depth 1` likely won't work with a hash other than HEAD

Jefftree: Thanks for the catch, updated

apelisse:
> That said, if we had to backport a kube-openapi change to 1.23 (unlikely, since 1.23 hits EOL in a couple weeks) or 1.24, we could make a kube-openapi branch prior to this commit, so go1.19+ support here seems fine

I don't think we would be able to backport kube-openapi that far since it would carry a lot of unwanted changes? We've branched in the past, notably because of smd version changing in backwards incompatible ways, I don't think that would be a problem.

hack/update-json-library.sh
apelisse: Should we ... make the hash a parameter of the script so that we can say, use "HEAD" in a CI script to test the latest version and see if it still works or experiment?

liggitt: overrideable default could work ... `HASH="${HASH:-540f014424240312547dcccddf11a8229ca0f463}"` or something

apelisse: Yeah, agreed, though I don't know if I want to change the script everytime we change the hash, whatever.
Also could be ...
```bash
HASH="${HASH:-$(cat JSON-EXPERIMENTAL-HASH)}"
```

Jefftree: Hmm, I would prefer that the hash is hard coded since we won't ever need to figure out which revision the repo is on. It's a bit of an extra step, but we can easily get the latest version by finding the latest commit hash and replacing it with that.

Jefftree: Adopted `HASH="${HASH:-$(cat ./hack/JSON-EXPERIMENTAL-HASH)}"`. A bit unsure if this file is the best way to express this variable, but it's been updated.

hack/update-json-library.sh
```diff
@@ -0,0 +1,33 @@
...
+rm -rf ".git"
+rm -rf ".github"
+# Remove images
+rm *.png
+# Remove go.{mod|sum}
+# NOTE: go-json-experiment has no go mod dependencies at the moment.
+#       If this changes, the code will need to be updated.
+rm go.mod go.sum
+# Update references to point to the fork
+find . -type f -name "*.go" | xargs sed -i 's/github.com\/go-json-experiment\/json/k8s.io\/kube-openapi\/pkg\/internal\/third_party\/go-json-experiment\/json/g'
```
apelisse, line 33: it's usually a little safer to do find -print0 | xargs -0 to avoid problems with spaces in files, though I don't think that'll be a problem here.

hack/update-json-library.sh
```diff
@@ -0,0 +1,33 @@
...
+rm -rf ".git"
+rm -rf ".github"
+# Remove images
+rm *.png
+# Remove go.{mod|sum}
+# NOTE: go-json-experiment has no go mod dependencies at the moment.
+#       If this changes, the code will need to be updated.
+rm go.mod go.sum
+# Update references to point to the fork
+find . -type f -name "*.go" -print0 | xargs -0 sed -i 's/github.com\/go-json-experiment\/json/k8s.io\/kube-openapi\/pkg\/internal\/third_party\/go-json-experiment\/json/g'
```
apelisse, line 33: So that at least it continues using the variable defined above.
```suggestion
find . -type f -name "*.go" -print0 | xargs -0 sed -i '' "s#github.com/go-json-experiment/json#k8s.io/kube-openapi/${GO_JSON_EXPERIMENT_DIR}#g"
```
```

**Comment by apelisse** [comment]
OK I'm fine with it
/approve

@liggitt Want to LGTM?

**Comment by liggitt** [comment]
mobile, last round looked good to me if dsnet indicated HEAD is good to grab now

File: hack/update-json-library.sh [hunk 8]
```diff
@@ -0,0 +1,33 @@
...
+rm -rf ".git"
+rm -rf ".github"
+# Remove images
+rm *.png
+# Remove go.{mod|sum}
+# NOTE: go-json-experiment has no go mod dependencies at the moment.
+#       If this changes, the code will need to be updated.
+rm go.mod go.sum
+# Update references to point to the fork
+find . -type f -name "*.go" -print0 | xargs -0 sed -i'' "s#github.com/go-json-experiment/json#k8s.io/kube-openapi/${GO_JSON_EXPERIMENT_DIR}#g"
```
apelisse, line 33:
```suggestion
find . -type f -name "*.go" -print0 | xargs -0 perl -pi -e "s#github.com/go-json-experiment/json#k8s.io/kube-openapi/${GO_JSON_EXPERIMENT_DIR}#g"
```
This works, though I think we turned into monsters.

apelisse:
@dsnet, would you recommend using the current HEAD or a specific hash for this?

apelisse:
/lgtm

We can make a new PR if the hash is no bueno.

PR #377, opened 2023-03-01
spec: Fix missing field in v2 serialization

apelisse:
cc @Jefftree @alexzielenski @liggitt 

We should disallow certain fields if we want to in the schema fuzzer, not hard-code all the allowed fields, otherwise we might miss fields, or forget to update with new fields, but for now, this code and test properly show the failure and the fix. I've also tested this against k8s, and it fixes the problem properly.

### DIFF
--- a/pkg/spec3/fuzz.go
+++ b/pkg/spec3/fuzz.go
@@ -1,9 +1,10 @@
 package spec3
 
 import (
-	fuzz "github.com/google/gofuzz"
 	"math/rand"
 	"strings"
+
+	fuzz "github.com/google/gofuzz"
 
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
@@ -182,11 +183,15 @@ var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
 		c.Fuzz(&v.Description)
 		v.URL = "https://" + randAlphanumString()
 	},
+	func(v *spec.SchemaURL, c fuzz.Continue) {
+		*v = spec.SchemaURL("https://" + randAlphanumString())
+	},
 	func(v *spec.Schema, c fuzz.Continue) {
 		if c.Intn(refChance) == 0 {
 			c.Fuzz(&v.Ref)
 			return
 		}
+		c.Fuzz(&v.Schema)
 		c.Fuzz(&v.VendorExtensible)
 		c.Fuzz(&v.Description)
 		c.Fuzz(&v.Nullable)
--- a/pkg/spec3/fuzz.go
+++ b/pkg/spec3/fuzz.go
@@ -170,13 +170,12 @@ var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
 		c.Fuzz(&v.VendorExtensible)
 	},
 	func(v *spec.Extensions, c fuzz.Continue) {
-		*v = spec.Extensions{}
 		numChildren := c.Intn(5)
-		if numChildren == 0 {
-			*v = nil
-		}
 		for i := 0; i < numChildren; i++ {
-			v.Add("x-"+randAlphanumString(), c.RandString()+"x")
+			if *v == nil {
+				*v = spec.Extensions{}
+			}
+			(*v)["x-"+c.RandString()] = c.RandString()
 		}
 	},
 	func(v *spec.ExternalDocumentation, c fuzz.Continue) {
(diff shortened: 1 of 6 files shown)


pkg/spec3/benchmark_serialization_test.go
```diff
@@ -22,7 +22,7 @@ var swaggerDiffOptions = []cmp.Option{cmp.Comparer(func(a spec.Ref, b spec.Ref)
 
 func TestOpenAPIV3RoundTrip(t *testing.T) {
 	var fuzzer *fuzz.Fuzzer
-	fuzzer = fuzz.NewWithSeed(1646791953)
+	fuzzer = fuzz.New()
```
apelisse, line 25: That's somewhat unrelated, but I think having the test be random is more likely to catch something.

apelisse:
/hold

I'm still looking into a few things right now, I'd rather clear these before we move forward :-(

apelisse:
Found another bug, I'm working on a few other things.

pkg/spec3/benchmark_serialization_test.go
apelisse: I'll revert that, I don't want the test to be flaky, but I'll run many instances of it before I do so.

apelisse:
removed the seed and ran some of these tests a few thousands times, always passed. I'll try this PR against k8s (at least for the fuzzed test that used to fail)

apelisse:
Ran again in k8s, passed a thousand times.

pkg/util/jsontesting/json_roundtrip.go
```diff
@@ -135,9 +130,5 @@ func (t RoundTripTestCase) RoundTripTest(example MarshalerUnmarshaler) error {
 		return fmt.Errorf("expected equal values: %v", cmp.Diff(expected, actual))
 	}
 
-	if err := JsonCompare(reEncoded, reEncodedV2); err != nil {
```
liggitt, line 138: was this dropped on purpose?

pkg/validation/spec/swagger_test.go
```diff
@@ -277,12 +289,6 @@ func TestSwaggerSpec_ExperimentalUnmarshal(t *testing.T) {
 	if !reflect.DeepEqual(control, actual) {
 		t.Fatal(cmp.Diff(control, actual, SwaggerDiffOptions...))
 	}
-
-	newJsonBytes, err := json.Marshal(actual)
-	require.NoError(t, err)
-	if err := jsontesting.JsonCompare(jsonBytes, newJsonBytes); err != nil {
```
liggitt, line 283: was this dropped on purpose?

pkg/validation/spec/swagger.go
```diff
@@ -385,15 +385,15 @@ func (s SchemaOrArray) MarshalJSON() ([]byte, error) {
 	if internal.UseOptimizedJSONMarshaling {
 		return internal.DeterministicMarshal(s)
 	}
-	if len(s.Schemas) > 0 {
+	if s.Schemas != nil {
```
liggitt, line 388: this is changing the pre-existing MarshalJSON function?

apelisse: Yes, the function was NOT round-tripping with the former code either ...

pkg/validation/spec/swagger.go
```diff
@@ -164,14 +164,14 @@ func (s *SchemaOrBool) UnmarshalJSON(data []byte) error {
...
 
 	var nw SchemaOrBool
-	if len(data) >= 4 {
-		if data[0] == '{' {
-			var sch Schema
-			if err := json.Unmarshal(data, &sch); err != nil {
-				return err
-			}
-			nw.Schema = &sch
+	if data[0] == '{' {
```
liggitt, line 167: can data be length 0 and this panic?

apelisse: Yeah, good point

pkg/validation/spec/swagger.go
```diff
@@ -164,14 +164,14 @@ func (s *SchemaOrBool) UnmarshalJSON(data []byte) error {
...
+		var sch Schema
+		if err := json.Unmarshal(data, &sch); err != nil {
+			return err
 		}
+		nw.Schema = &sch
+		nw.Allows = true
+	} else {
 		nw.Allows = !(data[0] == 'f' && data[1] == 'a' && data[2] == 'l' && data[3] == 's' && data[4] == 'e')
```
liggitt, line 175: huh, wondering how this guarded against a value of `fals` passing the `len(data) >= 4` check and panicking on `data[4] == 'e' `

pkg/validation/spec/swagger.go
liggitt: huh... surprised no fuzzing anywhere caught that

apelisse: Yeah, but if you have an empty list of schemas, this will end-up serializing `null`, which doesn't get deserialized to empty list of schemas.

pkg/validation/spec/swagger.go
apelisse: Yeah, I think this whole thing was just very wrong

apelisse: But it's very unlikely that you will have a json that finishes in `fals`

pkg/spec3/fuzz.go
```diff
@@ -169,82 +170,85 @@ var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
 		c.Fuzz(&v.VendorExtensible)
 	},
 	func(v *spec.Extensions, c fuzz.Continue) {
-		*v = spec.Extensions{}
```
Jefftree, line 172: I remember refactoring to this new version after a [comment](https://github.com/kubernetes/kube-openapi/pull/370#discussion_r1119160338) around the *v == nil case. Is this causing round tripping problems?

apelisse: No this one was actually just because it made me feel weird to set *v to a value to unset it later if unneeded. It felt more natural to set it when needed.

pkg/validation/spec/swagger_test.go
apelisse: These two were dropped on purpose, they basically test nothing :-(

pkg/validation/spec/swagger.go
apelisse: fixed

pkg/validation/spec/swagger.go
```diff
@@ -164,15 +164,15 @@ func (s *SchemaOrBool) UnmarshalJSON(data []byte) error {
...
+	if len(data) > 0 && data[0] == '{' {
+		var sch Schema
+		if err := json.Unmarshal(data, &sch); err != nil {
+			return err
 		}
-		nw.Allows = !(data[0] == 'f' && data[1] == 'a' && data[2] == 'l' && data[3] == 's' && data[4] == 'e')
+		nw.Schema = &sch
+		nw.Allows = true
+	} else {
+		json.Unmarshal(data, &nw.Allows)
```
apelisse, line 175: I think this would have tolerated a string or a number before and turned it into "true", which it won't now.

apelisse:
/hold cancel
I've discussed it with Alex and Jeff already

PR #381, opened 2023-03-30
Fix x-kubernetes-validations and add OpenAPI value validations support

jpbetz:
TODO:

- [ ] Add 'reason' field to x-kuberenetes-validations so we can support 'required', 'forbidden', ...


jpbetz:
/close
We will open a separate PR when we get back to this.

PR #384, opened 2023-04-12
Add k8s name related formats

jpbetz:
This adds the most heavily used name formats in k8s:

- k8s.io/short-name (represents the DNS 1123 label format)
- k8s.io/long-name (represents the DNS 1123 subdomain format)

Note: These formats implemented in k8s in ways that do not exactly match the RFCs (https://github.com/kubernetes/kubernetes/issues/71140, 
https://github.com/kubernetes/kubernetes/issues/79351). But, intend to use these formats to communicate the existing Kubernetes validation rules via OpenAPI, so we want to keep them as-is.

Benefits:

- Easier for CRDs to use the same name format already used in k8s.
- Useful for publication of OpenAPI from Declarative Validation

xref: https://json-schema.org/draft/2020-12/json-schema-validation#name-custom-format-attributes, https://spec.openapis.org/registry/format/

### DIFF
--- a/pkg/validation/strfmt/format.go
+++ b/pkg/validation/strfmt/format.go
@@ -16,6 +16,7 @@ package strfmt
 
 import (
 	"encoding"
+	"encoding/json"
 	"reflect"
 	"strings"
 	"sync"
@@ -230,4 +231,27 @@ func (f *defaultFormats) Parse(name, data string) (interface{}, error) {
 		}
 	}
 	return nil, errors.InvalidTypeName(name)
+}
+
+// unmarshalJSON provides a generic implementation of json.Unmarshaler interface's UnmarshalJSON function for basic string formats.
+func unmarshalJSON[T ~string](r *T, data []byte) error {
+	if string(data) == jsonNull {
+		return nil
+	}
+	var ustr string
+	if err := json.Unmarshal(data, &ustr); err != nil {
+		return err
+	}
+	*r = T(ustr)
+	return nil
+}
+
+// deepCopy provides a generic implementation of DeepCopy for basic string formats.
+func deepCopy[T ~string](r *T) *T {
+	if r == nil {
+		return nil
+	}
+	out := new(T)
+	*out = *r
+	return out
 }
--- a/pkg/validation/strfmt/kubernetes-extensions.go
+++ b/pkg/validation/strfmt/kubernetes-extensions.go
@@ -1,0 +1,143 @@
+// Copyright 2024 go-swagger maintainers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package strfmt
+
+import (
+	"encoding/json"
+	"regexp"
+)
+
+const k8sPrefix = "k8s-"
+
+func init() {
+	// register formats in the KubernetesExtensions registry:
+	//   - k8s-short-name
+	//   - k8s-long-name
+	shortName := ShortName("")
+	Default.Add(k8sPrefix+"short-name", &shortName, IsShortName)
+
+	longName := LongName("")
+	Default.Add(k8sPrefix+"long-name", &longName, IsLongName)
+}
+
+// ShortName is a name, up to 63 characters long, composed of alphanumeric
+// characters and dashes, which cannot begin or end with a dash.
+//
+// ShortName almost conforms to the definition of a label in DNS (RFC 1123),
+// except that uppercase letters are not allowed.
+//
+// xref: https://github.com/kubernetes/kubernetes/issues/71140
+//
+// swagger:strfmt k8s-short-name
+type ShortName string
+
+func (r ShortName) MarshalText() ([]byte, error) {
+	return []byte(string(r)), nil
+}
+
+func (r *ShortName) UnmarshalText(data []byte) error { // validation is performed later on
+	*r = ShortName(data)
+	return nil
+}
+
+func (r ShortName) String() string {
+	return string(r)
+}
+
+func (r ShortName) MarshalJSON() ([]byte, error) {
+	return json.Marshal(string(r))
+}
+
+func (r *ShortName) UnmarshalJSON(data []byte) error {
+	return unmarshalJSON(r, data)
+}
+
+func (r *ShortName) DeepCopyInto(out *ShortName) {
+	*out = *r
+}
+
+func (r *ShortName) DeepCopy() *ShortName {
+	return deepCopy(r)
+}
+
+const shortNameFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+
+// ShortNameMaxLength is a label's max length in DNS (RFC 1123)
+const ShortNameMaxLength int = 63
+
+var shortNameRegexp = regexp.MustCompile("^" + shortNameFmt + "$")
+
+// IsShortName checks if a string is a valid ShortName.
+func IsShortName(value string) bool {
+	return len(value) <= ShortNameMaxLength &&
+		shortNameRegexp.MatchString(value)
+}
+
+// LongName is a name, up to 253 characters long, composed of dot-separated
+// segments; each segment uses only alphanumerics and dashes (no
+// leading/trailing).
+//
+// LongName almost conforms to the definition of a subdomain in DNS (RFC 1123),
+// except that uppercase letters are not allowed, and there is no max length
+// limit of 63 for each of the dot-separated DNS Labels that make up the
+// subdomain.
+//
+// xref: https://github.com/kubernetes/kubernetes/issues/71140
+// xref: https://github.com/kubernetes/kubernetes/issues/79351
+//
+// swagger:strfmt k8s-long-name
+type LongName string
+
+func (r LongName) MarshalText() ([]byte, error) {
+	return []byte(string(r)), nil
+}
+
+func (r *LongName) UnmarshalText(data []byte) error { // validation is performed later on
+	*r = LongName(data)
+	return nil
+}
+
+func (r LongName) String() string {
+	return string(r)
+}
+
+func (r LongName) MarshalJSON() ([]byte, error) {
+	return json.Marshal(string(r))
+}
+
+func (r *LongName) UnmarshalJSON(data []byte) error {
+	return unmarshalJSON(r, data)
+}
+
+func (r *LongName) DeepCopyInto(out *LongName) {
+	*out = *r
+}
+
+func (r *LongName) DeepCopy() *LongName {
+	return deepCopy(r)
+}
+
+const longNameFmt string = shortNameFmt + "(\\." + shortNameFmt + ")*"
+
+// LongNameMaxLength is a subdomain's max length in DNS (RFC 1123)
+const LongNameMaxLength int = 253
+
+var longNameRegexp = regexp.MustCompile("^" + longNameFmt + "$")
+
+// IsLongName checks if a string is a valid LongName.
+func IsLongName(value string) bool {
+	return len(value) <= LongNameMaxLength &&
+		longNameRegexp.MatchString(value)
+}
--- a/pkg/validation/strfmt/default_test.go
+++ b/pkg/validation/strfmt/default_test.go
@@ -317,6 +317,10 @@ type testableFormat interface {
 }
 
 func testStringFormat(t *testing.T, what testableFormat, format, with string, validSamples, invalidSamples []string) {
+	testStringFormatWithRegistry(t, Default, what, format, with, validSamples, invalidSamples)
+}
+
+func testStringFormatWithRegistry(t *testing.T, registry Registry, what testableFormat, format, with string, validSamples, invalidSamples []string) {
 	// text encoding interface
 	b := []byte(with)
 	err := what.UnmarshalText(b)
@@ -347,24 +351,39 @@ func testStringFormat(t *testing.T, what testableFormat, format, with string, va
 	assert.Equalf(t, bj, b, "[%s]MarshalJSON: expected %v and %v to be value equal as []byte", format, string(b), with)
 
 	// validation with Registry
-	for _, valid := range append(validSamples, with) {
-		testValid(t, format, valid)
-	}
-
-	for _, invalid := range invalidSamples {
-		testInvalid(t, format, invalid)
-	}
+	t.Run("valid", func(t *testing.T) {
+		for _, valid := range append(validSamples, with) {
+			t.Run(valid, func(t *testing.T) {
+				testValidWithRegistry(t, registry, format, valid)
+			})
+		}
+	})
+	t.Run("invalid", func(t *testing.T) {
+		for _, invalid := range invalidSamples {
+			t.Run(invalid, func(t *testing.T) {
+				testInvalidWithRegistry(t, registry, format, invalid)
+			})
+		}
+	})
 }
 
 func testValid(t *testing.T, name, value string) {
-	ok := Default.Validates(name, value)
+	testValidWithRegistry(t, Default, name, value)
+}
+
+func testValidWithRegistry(t *testing.T, registry Registry, name, value string) {
+	ok := registry.Validates(name, value)
 	if !ok {
 		t.Errorf("expected %q of type %s to be valid", value, name)
 	}
 }
 
 func testInvalid(t *testing.T, name, value string) {
-	ok := Default.Validates(name, value)
+	testInvalidWithRegistry(t, Default, name, value)
+}
+
+func testInvalidWithRegistry(t *testing.T, registry Registry, name, value string) {
+	ok := registry.Validates(name, value)
 	if ok {
 		t.Errorf("expected %q of type %s to be invalid", value, name)
 	}
--- a/pkg/validation/strfmt/kubernetes-extensions_test.go
+++ b/pkg/validation/strfmt/kubernetes-extensions_test.go
@@ -1,0 +1,77 @@
+// Copyright 2024 go-swagger maintainers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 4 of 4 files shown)


liggitt:
are non-standard format strings supposed to be namespaced or not?

what are the implications for CRDs created using these format strings prior to this, allowing in data that ignores those format requirements, then auto-tightening validation when upgrading to a control plane with this support added?

jpbetz:
> are non-standard format strings supposed to be namespaced or not?

I have not found any mention of name-spacing in the specs. xref https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#name-custom-format-attributes

I am open to naming suggestions for these formats though. I had briefly considered something like "kubernetes-name" but couldn't think of anything that aided rather than harmed clarity on what the format actually is.

> what are the implications for CRDs created using these format strings prior to this, allowing in data that ignores those format requirements, then auto-tightening validation when upgrading to a control plane with this support added?

Yes, exactly. 

Today, [supported formats](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/formats.go) is used strip out unsupported
formats when used in older version of the kube-apiserver. We could extend this to have
two lists:  `supported` and `storage`.  All new formats will be added to `storage` for
at least one minor release of Kubernetes before the format is added to `supported` so
that any usages of the new formats written to CRDs downgrade safely.

jpbetz:
/assign @apelisse 

Would you be willing to review?

apelisse (commented):
looks good thanks

pkg/validation/strfmt/default.go
```diff
@@ -124,6 +124,69 @@ func IsEmail(str string) bool {
...
+const dns1123LabelFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+
+// DNS1123LabelMaxLength is a label's max length in DNS (RFC 1123)
+const DNS1123LabelMaxLength int = 63
+
+var dns1123LabelRegexp = regexp.MustCompile("^" + dns1123LabelFmt + "$")
+
+// IsDNS1123Label tests for a string that conforms to the definition of a label in
+// DNS (RFC 1123).
+func IsDNS1123Label(value string) bool {
```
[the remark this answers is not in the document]
jpbetz: Big oops!  I've simplified it all.  For reference, the code I copy-pasted from is at: https://github.com/kubernetes/kubernetes/blob/928adb27abf755b402be7594844e3c248b0b7670/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L188

apelisse:
Thanks, looks better to me :-)

pkg/validation/strfmt/default.go
```diff
@@ -124,6 +124,48 @@ func IsEmail(str string) bool {
 	return e == nil && addr.Address != ""
 }
 
+const dns1123LabelFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
```
liggitt, line 127: if we're copying/pasting from apimachinery validation, are we inheriting these issues:

https://github.com/kubernetes/kubernetes/issues/71140
* This requires things to be lowercase, which is more strict than the RFC

https://github.com/kubernetes/kubernetes/issues/79351#issuecomment-510493021
* This tolerates individual labels exceeding 63 chars, which is not as strict as the RFC

Should we fix those issues before adopting those formats here?

pkg/validation/strfmt/default.go
```diff
@@ -205,6 +247,15 @@ func init() {
...
 	Default.Add("password", &pw, func(_ string) bool { return true })
+
+	dns1123subdomain := DNS1123Subdomain("")
+	Default.Add("dns1123subdomain", &dns1123subdomain, IsDNS1123Subdomain)
+
+	dns1123label := DNS1123Label("")
+	Default.Add("dns1123label", &dns1123label, IsDNS1123Label)
+
+	dns1035label := DNS1035Label("")
+	Default.Add("dns1035label", &dns1035label, IsDNS1035Label)
```
liggitt, line 258: if these get added to the default set of recognized formats here, how do we roll them out in a controlled way in kubernetes?

liggitt:
> > what are the implications for CRDs created using these format strings prior to this, allowing in data that ignores those format requirements, then auto-tightening validation when upgrading to a control plane with this support added?
> 
> Yes, exactly.
> 
> Today, [supported formats](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/formats.go) is used strip out unsupported formats when used in older version of the kube-apiserver. We could extend this to have two lists: `supported` and `storage`. All new formats will be added to `storage` for at least one minor release of Kubernetes before the format is added to `supported` so that any usages of the new formats written to CRDs downgrade safely.

I'm not sure I understand how that makes existing persisted CR data safe. However long we take to roll out support for the format, at *some* point, would we start treating existing persisted data that didn't match the specified format as invalid?

jpbetz:
> I'm not sure I understand how that makes existing persisted CR data safe. However long we take to roll out support for the format, at _some_ point, would we start treating existing persisted data that didn't match the specified format as invalid?

We could solve this together with the more general problem of CRD validation ratcheting.  

E.g. a CRD author adds a format to a CRD, and all persisted CR data becomes invalid. This problem we're discussing is similar except that instead of the CRD author adding a format, the CRD author already had a non-validating format in the CRD and then Kubernetes changed the format to be validating.

Ratcheting rule would be: if any CR data fails validation but is unchanged from the persisted value, we ignore the validation failure and allow the update.

jpbetz:
> > We could solve this together with the more general problem of CRD validation ratcheting.
> 
> I agree a solution to that would resolve the issue with this. Shouldn't we get that solution in place before adding this? This would be the first example I can think of where _we_ made a change that would reinterpret existing persisted CRDs in ways that would invalidate existing data

SGTM. I'll poke around and see who might be able to implement.

jpbetz:
/close
When we circle back on this we'll either reopen or create a new PR.

alexzielenski:
> We could solve this together with the more general problem of CRD validation ratcheting.
> I agree a solution to that would resolve the issue with this. Shouldn't we get that solution in place before adding this?

Just updating progress on this front was made with CRD Ratcheting Alpha added in 1.28: https://kep.k8s.io/4008

Will need to wait for it to mature before using it for name formats, but in the interim we can use inline OpenAPI schemas to describe the regex/cel expressions used for most formats.

jpbetz:
/reopen

CRD Ratcheting was promoted to beta in 1.30, freeing us up to resume progress on this.

pkg/validation/strfmt/default.go
jpbetz: We intend to publish OpenAPI the in-tree types using these formats.  So I think we're stuck inheriting the it all.

That said, we shouldn't using the RFC identifiers for formats that are not exactly compliant with the RFC.  I'll propose some alternative names.

pkg/validation/strfmt/default.go
jpbetz: This was held until CRD Ratcheting was promoted to beta in 1.30, we'll leverage that capability.

jpbetz: I'll update this code so we can manage the newly added formats separate from the pre-existing ones.

liggitt:
Now that we have CRD ratcheting, it would be good to give guidance around when expanded openapi formats are the preferred way to model a validation, and when a CEL library is.

xref https://github.com/kubernetes/kubernetes/issues/124490#issuecomment-2741495825

jpbetz:
> Now that we have CRD ratcheting, it would be good to give guidance around when expanded openapi formats are the preferred way to model a validation, and when a CEL library is.
> 
> xref [kubernetes/kubernetes#124490 (comment)](https://github.com/kubernetes/kubernetes/issues/124490#issuecomment-2741495825)

We are missing well documented guidance.  Let's get that going.  I like the idea of CEL "supplementing" OpenAPI formats only were clearly needed.

Related- We've fallen behind on supporting formats.  https://github.com/kubernetes/kubernetes/issues/130639 describes where we CEL support for formats that SHOULD have an OpenAPI format.

To help catch up, we've added emulation version support to formats: https://github.com/kubernetes/kubernetes/pull/130783.  Combined with ratcheting, this should give us all the machinery we need to add the formats we agree should be added.

jpbetz:
cc @thockin for the Declarative Validation aspect.  We anticipate needing OpenAPI formats for the built-in formats that we validate with hand written code so that we can publish the declarative validation rules through OpenAPI.

thockin:
Is there an extension-mechanism for format names?  Like `x-something` or `example.com/something`?

The ultimate goal, I think is unification.  CRDs can be written saying "this field is validated by format=x-k8s-dns-label" and we will do that server-side; Builtin types can express the same thing in a different way; apiserver publishes this in a way that an "enlightened" client can handle if they wanted to to do shift-left validation (e.g. as a git pre-commit).

jpbetz:
> Is there an extension-mechanism for format names?  Like `x-something` or `example.com/something`?

Unfortunately no.  The RFC drafts that describe "format" don't provide an extension mechanism.  https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-7.2 proposes a core set of formats but the only wording for extensions is:

```
Implementations MAY add custom format attributes.  Save for agreement
   between parties, schema authors SHALL NOT expect a peer
   implementation to support this keyword and/or custom format
   attributes.
```

thockin:
Should we make one up?

jpbetz:
> Should we make one up?

Yes, that's probably best.

I poked around the set of formats listed in the draft RFC is being kept to a very small list and doesn't look to be accepting almost any new formats (semver was proposed and rejected).

The naming convention of formats is not specified, but the core set use `lower-case` with heavy uses of acronyms (`uuid`, `uri`, ...)

I'd prefer something not terribly long  like `x-k8s-`  or just `k8s-`.  But I can see an argument for `x-kubernetes-` because we use that for JSON Schema customizations already.  The context is different here. We're defining string values not field names, but still..

thockin:
Another option might be label-style?  "k8s.io/dns-label"?  As long as we pick something unique, we can always adapt later. The biggest concern I have is violating some documented or undocumented assumptions, like "must be alphanumeric".

I defer to your wisdom here

jpbetz:
This PR has been refreshed and is ready for review.

I've gone with "k8s.io/" as the prefix.

thockin (commented):
A couple  thoughts on names.  Over time we have used 3 different DNS formats.  But what we use is not really DNS compatible, it's DNS-ish.  

1) Should we continue with the "DNS" nomenclature?  This looks like a significant move towards publishing this terminology so this may be our last chance to name the formats better.  Or it might be too late already.

2) Should we nominate one of the 3 as the "normal" format (e.g. `k8s.io/dns-label`) and let the others be wierd?

jpbetz:
> A couple thoughts on names. Over time we have used 3 different DNS formats. But what we use is not really DNS compatible, it's DNS-ish.
> 
> 1. Should we continue with the "DNS" nomenclature?  This looks like a significant move towards publishing this terminology so this may be our last chance to name the formats better.  Or it might be too late already.
> 2. Should we nominate one of the 3 as the "normal" format (e.g. `k8s.io/dns-label`) and let the others be wierd?

Since DNS 1123 subdomain is the type we use most often for resource name validation, how about a format name that emphasizes that?  "k8s.io/basic-name" or something?

thockin:
I am terrible at naming things, but a few quick thoughts:

1) DNS 1123 Subdomain and Label are most common, 1035 less so, and it looks like 952 has been weeded out.  So 1123 is "obviously" the default.

2) We already use the word "qualified" in "qualified name".  I hate it there, too.

3) We should avoid "simple" in names.

4) AFAICT Service is the last and only resource which demands 1035.  Maybe we can ratchet it to 1123?   Can we do the same for other users of 1035?

Then the name could just be "dns-label" or "dns-style-label" or something?  Would love alternatives to "qualified name" too :)

jpbetz:
> 2. We already use the word "qualified" in "qualified name".  I hate it there, too.

Maybe we just stick with subdomain then?  We could consider "dns-full-name" or something. But if we're avoiding "qualified", is "full name" is any better? 

> 3. We should avoid "simple" in names.

+1

> 4. AFAICT Service is the last and only resource which demands 1035.  Maybe we can ratchet it to 1123?   Can we do the same for other users of 1035?

Dropping 1035 from the format list SGTM.

> Then the name could just be "dns-label" or "dns-style-label" or something? Would love alternatives to "qualified name" too :)

How about:

- `k8s.io/dns-subdomain` (dns 1123 subdomain)
- `k8s.io/dns-subdomain-prefix`
- `k8s.io/dns-label` (dns 1123 label)
- `k8s.io/dns-label-prefix`

?

jpbetz:
My current concerns:

- Not sure what value the term "DNS" really adds to the users of these formats (they don't conform to DNS specs.. even though they ALMOST conform)
- The user of the term "label" is pulling in a term from DNS feels like it could confuse anyone that is thinking in k8s terms and immediately starts thinking of k8s labels when they see it.

jpbetz:
I've updated this to use `k8s.io/short-name` and `k8s.io/long-name`

danwinship:
> If we don't need to handle Service

I think either way, we should probably ignore Service for purposes of this PR; either we manage to fix Service and it becomes just a `k8s.io/short-name`, or we don't, and we can add `k8s.io/service-name` later.

thockin:
I asked https://github.com/OAI/learn.openapis.org/issues/128 to see if there is a real norm here.  I *like* the `k8s.io/something` extension mechanism, but from reading a lot of the issues around this, I wonder if just `k8s-something` is more "regular"?

k8s-short-name + k8s-long-name ?

k8s-shortname + k8s-longname ?

k8s-name + k8s-long-name ?

k8s-shortname + k8s-name

jpbetz:
> I wonder if just `k8s-something` is more "regular"?

We've gone full circle.  This is what I'd proposed originally.

thockin:
We have two open questions in this thread:

## 1) Overall notation

Almost full circle - after looking at all the openapi and JSON schema got repos and specs I figured:

1) Nothing in the spec *prevents* us from using `k8s.io/something`
2) Nobody else seems to use any notation other than `lower-case-with-dashes` for format names
3) It seems conventional to use a "sufficiently unique" prefix for locally defined format names

As such, I think the lowest risk and most non-contentious path is to declare a prefix (eg. `k8s-`) and use it on all of our custom format names.  We should use `k8s-something-name` style, even though it is not the same syntax as label names.

I think this should extend into declarative validation, and all our custom formats should be "k8s-something"

Agree or discuss more?

## 2) Rename "DNS label" and "DNS subdomain"

I can't come up with anything more semantically meaningful that "short" and "long".  Maybe just more descriptive.  We want to eschew the words "DNS" and "label" in the name.  "object name" isn't right because we use them other places.  Even "name" is questionable.  The short form is similar to a hostname, but we should avoid that, too since it is not exact.

When it comes to object names we use subdomain more than label (sadly) but for other fields we seem to use label more (totally haphazard estimate by line-counting calls to each func).

So I am fine with "k8s-short-name" and "k8s-long-name" or maybe with dropping either "long" or "short" (making that one more default-ish).

Or I am fine with "k8s-name" (or "k8s-short-name") and "k8s-dotted-name" or something similarly descriptive.

You are closest to the consequences of this decision, Joe, so I defer.

jpbetz:
Apologies for the delay, I lost track of this one and am just circling back now.

> Agree or discuss more?

Agree. 

> So I am fine with "k8s-short-name" and "k8s-long-name" or maybe with dropping either "long" or "short" (making that one more default-ish).
> 
> Or I am fine with "k8s-name" (or "k8s-short-name") and "k8s-dotted-name" or something similarly descriptive.

I also could live with any of these variations.

I'm going to go with `k8s-short-name` and `k8s-long-name`.

While I'm tempted to have just `k8s-name` instead of `k8s-short-name`,  we use `k8s-long-name` to identify resources, so I don't think crowing `k8s-short-name` as "the k8s name format" is quite right.

jpbetz:
@thockin @liggitt this is ready

pkg/validation/strfmt/kubernetes-extensions.go
```diff
@@ -0,0 +1,155 @@
...
+package strfmt
+
+import (
+	"encoding/json"
+	"regexp"
+)
+
+// KubernetesExtensions is the formats registry for JSON Schema formats
+// extensions defined by the Kubernetes project.
+var KubernetesExtensions = NewSeededFormats(nil, nil)
```
liggitt, line 24: This is peer to the existing `Default` var? looks like we have a lot of things using that other var... will they need to switch to this one?

pkg/validation/strfmt/format.go
```diff
@@ -231,3 +233,40 @@ func (f *defaultFormats) Parse(name, data string) (interface{}, error) {
...
+
+// scan provides a generic implementation of sql.Scanner interface's Scan function for basic string formats.
+func scan[T ~string](r *T, raw interface{}) error {
+	switch v := raw.(type) {
+	case []byte:
+		*r = T(v)
+	case string:
+		*r = T(v)
+	default:
+		return fmt.Errorf("cannot sql.Scan() strfmt.StringFormat from: %#v", v)
```
liggitt, line 245: 🤨  are the sql.Scan errors load-bearing or can we stop outputting that?

pkg/validation/strfmt/kubernetes-extensions.go
```diff
@@ -0,0 +1,155 @@
...
+	return len(value) <= ShortNameMaxLength &&
+		shortNameRegexp.MatchString(value)
+}
+
+// LongName is a name, up to 253 characters long, composed of dot-separated
+// segments; each segment uses only alphanumerics and dashes (no
+// leading/trailing).
+//
+// LongName almost conforms to the definition of a subdomain in DNS (RFC 1123),
+// except that uppercase letters are not allowed. and there is no max length
```
liggitt, line 101:
```suggestion
// except that uppercase letters are not allowed, and there is no max length
```

pkg/validation/strfmt/format.go
jpbetz: For some reason most of the string formats in kube-openapi support `sql.Scanner` and output errors like this.  But a few don't.  I'll just not support it and open an issue to drop it from the other string formats.

jpbetz: https://github.com/kubernetes/kube-openapi/issues/536

thockin:
If the API validation functions were exposed in a low-deps library, would we take a dep on it here?

/lgtm

thockin:
anything holding up this naming decision?

jpbetz:
> anything holding up this naming decision?

I'm ready for this to merge.

jpbetz:
/approve
(This PR must pre-date my approver status)

PR #385, opened 2023-04-19
Make sure updating spec is not racy

apelisse:
cc @soltysh @liggitt

### DIFF
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -227,23 +227,31 @@ func TestConcurrentReadStaleCache(t *testing.T) {
 		t.Errorf("Unexpected error in preparing returnedPb: %v", err)
 	}
 
-	jsonResults := make(chan []byte)
-	protoResults := make(chan []byte)
+	jsonResultsChan := make(chan []byte)
+	protoResultsChan := make(chan []byte)
+	updateSpecChan := make(chan struct{})
 	for i := 0; i < concurrency; i++ {
-		go func() { jsonResults <- getJSONBodyOrDie(server) }()
-		go func() { protoResults <- getProtoBodyOrDie(server) }()
+		go func() {
+			sc := s
+			o.UpdateSpec(&sc)
+			updateSpecChan <- struct{}{}
+		}()
+		go func() { jsonResultsChan <- getJSONBodyOrDie(server) }()
+		go func() { protoResultsChan <- getProtoBodyOrDie(server) }()
 	}
 	for i := 0; i < concurrency; i++ {
-		r := <-jsonResults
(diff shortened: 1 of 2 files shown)


apelisse:
I've tried giving a nil swagger, and it fails when trying to serialize the json (as expected), not with a nil json ...

Next question is: if the swagger object changes causing a race, would it return a nil/empty json? I guess it's possible since that's probably undefined behavior.

Jefftree:
@apelisse I don't think we need this anymore now that the cache is thread safe?

apelisse:
I've removed the lock and improved the race test a little bit, should be good to go now, PTAL.

pkg/handler/handler_test.go
```diff
@@ -229,7 +229,13 @@ func TestConcurrentReadStaleCache(t *testing.T) {
 
 	jsonResults := make(chan []byte)
 	protoResults := make(chan []byte)
+	updateSpec := make(chan struct{})
```
Jefftree, line 232: nit: rename to something that indicates this is a channel?

pkg/handler/handler_test.go
```diff
@@ -229,7 +229,13 @@ func TestConcurrentReadStaleCache(t *testing.T) {
...
 	for i := 0; i < concurrency; i++ {
+		go func() {
+			sc := s
+			o.UpdateSpec(&sc)
+			updateSpec <- struct{}{}
```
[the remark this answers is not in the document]
apelisse: Because you've used channels throughout so I was just reusing the same technique for consistency.

pkg/handler/handler_test.go
apelisse: `specUpdatedChan`? `specUpdated`? I'm trying to be consistent with the two above

pkg/handler/handler_test.go
```diff
@@ -229,7 +229,13 @@ func TestConcurrentReadStaleCache(t *testing.T) {
 
 	jsonResults := make(chan []byte)
 	protoResults := make(chan []byte)
+	updateSpec := make(chan struct{})
 	for i := 0; i < concurrency; i++ {
+		go func() {
+			sc := s
```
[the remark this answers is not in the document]
apelisse: I want to be sure that `Replace` actually replaces with something new. Agreed the spec is not going to trigger a full update of anything.

pkg/handler/handler_test.go
Jefftree: I like `specUpdatedChan`, yeah I should have added that suffix to those as well

apelisse:
Appended `Chan` to all the channels in that test.

PR #387, opened 2023-04-26
Fix cache data race

Jefftree:
Fixes https://github.com/kubernetes/kubernetes/issues/117363

Add lock to all cache get operations for OpenAPI V2. OpenAPI V3 is already handled with a [lock](https://github.com/kubernetes/kube-openapi/blob/master/pkg/handler3/handler.go#L168)

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/NYTimes/gziphandler"
@@ -135,6 +136,9 @@ func RegisterOpenAPIVersionedService(spec *spec.Swagger, servePath string, handl
 
 // RegisterOpenAPIVersionedService registers a handler to provide access to provided swagger spec.
 func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handler common.PathHandler) error {
+	// Mutex protects the cache chain
+	var mutex sync.Mutex
+
 	accepted := []struct {
 		Type                string
 		SubType             string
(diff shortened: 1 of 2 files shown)


pkg/handler/handler.go
```diff
@@ -163,7 +175,7 @@ func (o *OpenAPIService) RegisterOpenAPIVersionedService(servePath string, handl
 						continue
 					}
 					// serve the first matching media type in the sorted clause list
-					result := accepts.GetDataAndEtag.Get()
+					result := o.getDataAndEtagFromCache(accepts.GetDataAndEtag)
```
apelisse, line 178: Ah, I think I know now how I committed this atrocity. I think I wanted to have the mutex NOT in the strcut, but directly in the RegisterOpenAPIVersionedService function. You basically only use the mutex in one place here so I think it's fine to just do have a local mutex?

apelisse:
Also, have you properly tested v3, are you sure it doesn't suffer from the same thing?

pkg/handler/handler_test.go
```diff
@@ -200,3 +200,32 @@ func TestToProtoBinary(t *testing.T) {
...
+		t.Errorf("Unexpected error in register OpenAPI versioned service: %v", err)
+	}
+	server := httptest.NewServer(mux)
+	defer server.Close()
+
+	returnedJSON := string(normalizeSwaggerOrDie(returnedSwagger))
+
+	results := make(chan string)
+	for i := 0; i < 2; i++ {
+		go func() { results <- getBodyOrDie(server) }()
```
apelisse, line 223: I know the code doesn't let you do that easily, but should we try to get different types of content at the same time, i.e. proto and json for example?

Jefftree:
> Also, have you properly tested v3, are you sure it doesn't suffer from the same thing?

Here's the V3 test: https://github.com/kubernetes/kube-openapi/pull/389

apelisse:
Thanks!
/lgtm
/approve

PR #389, opened 2023-04-28
[WIP] OpenAPI V3 race test

Jefftree:
Ensure that OpenAPI V3 fetching is thread safe

/cc @apelisse


pkg/handler3/handler_test.go
```diff
@@ -539,3 +553,80 @@ func TestUpdateGroupVersion(t *testing.T) {
 		t.Fatalf("Invalid number of Paths, expected 2: %v", discovery.Paths)
 	}
 }
+
+func TestConcurrentReadStaleCache(t *testing.T) {
```
apelisse, line 557: Thanks, looks generally good, can we also get a few of the `root` document, and maybe some other document at the same time, just to be super safe?

PR #390, opened 2023-05-15
Drop unused mapstructure functionality

liggitt:
Cleans up unused dependency


apelisse:
Looks good to me, but curious if @alexzielenski uses that?

liggitt:
I *think* this was residual from the old go-openapi repos

alexzielenski:
> Looks good to me, but curious if @alexzielenski uses that?

nope, we are doing some things with strfmt in kubectl-validate but nothing regarding this mapstructure dep

liggitt:
dropped WIP, ready whenever you are

PR #392, opened 2023-05-23
Remove empty error returned from registering OpenAPI

Jefftree:
We never return an error from registering the openapi so remove it from the function signature


Jefftree:
/assign @apelisse 

Breaks our interface and will require updating, but cleans up the code and clients won't need an `if err != nil` check anymore.

apelisse:
Thanks!
/lgtm
/approve

PR #396, opened 2023-05-24
cached: Parallelizes ListMerge caches refresh

apelisse:
I was curious if we needed to have both a parallel and a serial, but I do not see a purpose for the serial right now so I've decided to only keep parallel for now. Also I couldn't quite see how to add a new test for that.


apelisse:
cc @Jefftree @justinsb

apelisse:
I hate that I have to used an unnamed struct here for the pair, but if that's what golang wants me to do, then I'm just going to do that :-)

pkg/cached/cache.go
```diff
@@ -183,10 +183,27 @@ func NewListMerger[T, V any](mergeFn func(results []Result[T]) Result[V], caches
...
+				Result[T]
+			}{
+				index,
+				c.caches[index].Get(),
+			}
+		}(i)
+	}
+	for i := 0; i < len(c.caches); i++ {
+		res := <-ch
+		cacheResults[res.int] = res.Result
```
Jefftree, line 206: Out of curiosity, is this single thread safe? Is it a good time to revisit our atomicity guarantees for the library?

apelisse: I don't think this changes the guarantees that are already documented. Do you suggest that we need more tests to verify these guarantees?

Jefftree: Not being thread safe usually means that we still have a guarantee for correctness when the function is not called in parallel. Spawning threads to call a Get() that is not thread safe seems like it may break that guarantee?

apelisse:
OK this is replaced by #403

PR #398, opened 2023-05-25
cached: Do not sort ListMerger test outputs

apelisse:
This is actually making the test a lot less useful because we don't see anymore if the output is stable which is the main purpose of this.


Jefftree:
/lgtm

Oh wow, I definitely agree that this should be removed. But also, etags and the merge list should definitely not be sorted separately since the etags list should correspond to the order of sorted merge list instead of the order produced by `sort.Strings()`

PR #399, opened 2023-05-25
cached: Add test to show that diamond is thread-safe

apelisse:


Jefftree:
Question: Does this fail with parallelized listmerge cache refreshes?

apelisse:
Yes

Issue #397, opened 2023-05-25
Request to remove reference to deprecated module 'github.com/golang/protobuf


abhithmv:
Hi,
kube-openapi has a reference to deprecated module 'github.com/golang/protobuf'. Because of which our consuming module depends indirectly on deprecated module.
Is it possible to remove the deprecated module and release a new version of kube-openapi?
Thanks.

Jefftree:
/cc @apelisse @liggitt

Definitely see some use cases of this library in handler.go that could be updated: https://github.com/kubernetes/kube-openapi/blob/master/pkg/handler/handler.go#L111-L117

I haven't looked too deeply, but it looks to also be an indirect dependency based on a library that we use: https://github.com/google/gnostic which also has some usage but also as part of its indirect dependencies from other imports.

apelisse:
I'm not sure what you're suggesting? We definitely use protobuf to serialize the openapi into protobuf.

Jefftree:
github.com/golang/protobuf -> google.golang.org/protobuf

apelisse:
> kube-openapi has a reference to deprecated module 'github.com/golang/protobuf'. Because of which our consuming module depends indirectly on deprecated module.

What problem does this cause?

PR #401, opened 2023-05-30
Use new repo for protobuf library

Jefftree:
Fixes https://github.com/kubernetes/kube-openapi/issues/397

Per https://github.com/golang/protobuf#go-support-for-protocol-buffers, github.com/golang/protobuf is deprecated and superceded by google.golang.org/protobuf.

/assign @apelisse 
/cc @abhithmv


apelisse:
Do we still need to get gnostic to migrate?

Jefftree:
> Do we still need to get gnostic to migrate?

To remove the indirect dependency, yes.

PR #402, opened 2023-05-30
Use gnostic-models over gnostic

Jefftree:
gnostic is pinned on [v0.5.7-v3refs](https://github.com/kubernetes/kube-openapi/blob/master/go.mod#L12) because we did not want to capture the additional dependencies introduced with later versions of gnostic. The OpenAPI component has been moved to gnostic-models so update the library path to reflect that.

/assign @apelisse


apelisse:
That removes a lot of dependencies, nice.
/lgtm
/approve

laszlocph:
Can this be related?

Just updated with `go get -u` and it is complaining about gnostic structs.

```
../../go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.14.2/openapi/openapi.go:683:33: cannot use doc (variable of type *"github.com/google/gnostic/openapiv2".Document) as *"github.com/google/gnost
ic-models/openapiv2".Document value in argument to swagger.FromGnostic                                                                                                                        
# k8s.io/kubectl/pkg/util/openapi
../../go/pkg/mod/k8s.io/kubectl@v0.27.2/pkg/util/openapi/openapi.go:52:38: cannot use doc (variable of type *"github.com/google/gnostic/openapiv2".Document) as *"github.com/google/gnostic-models/openapiv2".Document value in argument to proto.NewOpenAPIData
# k8s.io/client-go/applyconfigurations/meta/v1
../../go/pkg/mod/k8s.io/client-go@v0.27.2/applyconfigurations/meta/v1/unstructured.go:64:38: cannot use doc (variable of type *"github.com/google/gnostic/openapiv2".Document) as *"github.com/google/gnostic-models/openapiv2".Document value in argument to proto.NewOpenAPIData
```

apelisse:
Yeah, this change will break all the things that used to pull the former gnostic model library. That's a fair amount of things.
How do you want to proceed Jeff? We can either give up this change, that's sad, or we can at least update kubectl to use the new models, or maybe we should tag/cut a branch for 1.27?

Jefftree:
It's unfortunate that we can't do this update atomically. I've opened https://github.com/kubernetes-sigs/kustomize/pull/5186 for kustomize and https://github.com/kubernetes/kubernetes/issues/118340 to track this migration across k/k.

Out of curiosity, what repo is the `go get -u` done from? We've generally discouraged `go get -u` in k/k because of issues like these and pin our dependencies appropriately to avoid breaking things.

stefanprodan:
The fact that gnostic doesn't import gnostic-models breaks all tools in the Kubernetes ecosystem. For us in [Flux](https://github.com/fluxcd) means that we need to wait for all our Kubernetes dependencies (apimachinery, apiextensions-apiserver, client-go, cli-utils) and for controller-runtime, Kustomize, Helm, Cosign, etc to switch to gnostic-models, before we can update Flux. We are looking at months of being behind upstream waiting on everyone to make the switch, if any the these tools have a major CVE, this change makes it impossible to patch Flux.

sttts:
@stefanprodan why is that? Can you elaborate? How is go.mod complaining?

PR #403, opened 2023-05-31
Thread safe cached

apelisse:
Improve cached framework to be thread-safe. I was convinced by the following arguments:
1. We can have multiple different entry points into the "get-chain", so it's possibly hard to properly secure all of them
2. If the dependency tree ends-up having a diamond shape, we can't make multithreaded Get resolution without causing a race-condition (that feature is also added in the pull-request)
3. It's unintuitive that "Get" needs to be protected, since it sounds like a read method rather than a write.

cc @Jefftree

### DIFF
--- a/pkg/cached/cache.go
+++ b/pkg/cached/cache.go
@@ -19,6 +19,8 @@ limitations under the License.
 // operations are not repeated unnecessarily. The operations can be
 // created as a tree, and replaced dynamically as needed.
 //
+// All the operations in this module are thread-safe.
+//
 // # Dependencies and types of caches
 //
 // This package uses a source/transform/sink model of caches to build
@@ -34,15 +36,6 @@ limitations under the License.
 //     replaced with a new one, and saves the previous results in case an
 //     error pops-up.
 //
-// # Atomicity
-//
-// Most of the operations are not atomic/thread-safe, except for
-// [Replaceable.Replace] which can be performed while the objects are
-// being read. Specifically, `Get` methods are NOT thread-safe. Never
-// call `Get()` without a lock on a multi-threaded environment, since
-// it's usually performing updates to caches that will require write
-// operations.
-//
 // # Etags
 //
 // Etags in this library is a cache version identifier. It doesn't
@@ -57,6 +50,7 @@ package cached
 
 import (
 	"fmt"
+	"sync"
 	"sync/atomic"
 )
 
@@ -100,13 +94,6 @@ func (r Result[T]) Get() Result[T] {
 type Data[T any] interface {
 	// Returns the cached data, as well as an "etag" to identify the
 	// version of the cache, or an error if something happened.
-	//
-	// # Important note
-	//
-	// This method is NEVER thread-safe, never assume it is OK to
-	// call `Get()` without holding a proper mutex in a
-	// multi-threaded environment, especially since `Get()` will
-	// usually update the cache and perform write operations.
 	Get() Result[T]
 }
 
@@ -155,6 +142,7 @@ func NewMerger[K comparable, T, V any](mergeFn func(results map[K]Result[T]) Res
 }
 
 type listMerger[T, V any] struct {
+	lock         sync.Mutex
 	mergeFn      func([]Result[T]) Result[V]
 	caches       []Data[T]
 	cacheResults []Result[T]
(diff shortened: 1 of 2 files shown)


pkg/cached/cache.go
```diff
@@ -286,23 +292,21 @@ type Replaceable[T any] struct {
...
-// As all implementations of Get, this implementation is NOT
-// thread-safe. Please properly lock a mutex before calling this method
-// if you are in a multi-threaded environment, since this method will
-// update the cache and perform write operations.
 func (c *Replaceable[T]) Get() Result[T] {
 	result := (*c.cache.Load()).Get()
-	if result.Err != nil && c.result != nil && c.result.Err == nil {
-		return *c.result
+
+	for {
```
Jefftree, line 298: It's cool that you did this without a mutex, the logic feels a bit complex and I'm not sure if we get much performance benefit. Logically it makes sense that we'll terminate when the CompareAndSwap succeeds but having an infinite for loop could cause problems if we're not careful. wdyt?

pkg/cached/cache.go
```diff
@@ -19,6 +19,8 @@ limitations under the License.
 // operations are not repeated unnecessarily. The operations can be
 // created as a tree, and replaced dynamically as needed.
```
Jefftree, line 20: Technically we support DAGs now :)

pkg/cached/cache.go
apelisse: Yeah, I don't think the infinite loop is so bad but I might be wrong, feel free to think of a test that could trigger that?
Complexity is a little higher than mutexes, my main goal here is that I don't want to lock both in Get and in Replace. I could still use a mutex in Get and keep the atomic pointer for replace.

pkg/cached/cache_test.go
```diff
@@ -1054,3 +1055,61 @@ func TestListDAG(t *testing.T) {
...
+		"two": NewRandomSource(),
+	})
+	wg.Add(1)
+	go repeatedGet(merger, end, &wg)
+	listmerger := cached.NewListMerger(func(results []cached.Result[int64]) cached.Result[int64] {
+		return results[0]
+	}, []cached.Data[int64]{static, result, &replaceable, merger})
+	wg.Add(1)
+	go repeatedGet(listmerger, end, &wg)
+
```
Jefftree, line 1114: Do we need a `wg.Wait()`?

pkg/cached/cache.go
Jefftree: >  I could still use a mutex in Get and keep the atomic pointer for replace.
I would probably slightly prefer that approach but will defer the decision to you. 

This might already be a pattern we use in aggregated discovery so I don't feel too strongly but did want to bring it up. Only case I can think of is a thread being starved and constantly losing the compareandswap operation which would increase latency on the get. However, this case probably never happens and we have bigger problems if we have such a high frequency of reads/writes to the cache 😄

pkg/cached/cache_test.go
apelisse: we do thanks

apelisse:
Updated, PTAL

pkg/cached/cache_test.go
```diff
@@ -1054,3 +1055,61 @@ func TestListDAG(t *testing.T) {
...
+	go func(r *cached.Replaceable[int64], end time.Time, wg *sync.WaitGroup) {
+		for time.Now().Before(end) {
+			r.Replace(NewRandomSource())
+		}
+		wg.Done()
+	}(&replaceable, end, &wg)
+	merger := cached.NewMerger(func(results map[string]cached.Result[int64]) cached.Result[int64] {
+		return results["one"]
+	}, map[string]cached.Data[int64]{
+		"one": NewRandomSource(),
```
Jefftree, line 1104: Why not append the two strings or something? A single map element is basically a transformer so we're not really testing the merger with multiple sources?

pkg/cached/cache_test.go
```diff
@@ -1054,3 +1055,61 @@ func TestListDAG(t *testing.T) {
...
+	})
+	wg.Add(1)
+	go repeatedGet(merger, end, &wg)
+	listmerger := cached.NewListMerger(func(results []cached.Result[int64]) cached.Result[int64] {
+		return results[0]
```
Jefftree, line 1109: Same comment, would prefer more than one element so it's not a transformer

apelisse:
OK I've made the test do something though the goal is not to test that they work, but merely that they are thread-safe internally.

Jefftree:
> OK I've made the test do something though the goal is not to test that they work, but merely that they are thread-safe internally.

+1 agree with this approach.

/lgtm
/hold

Feel free to unhold whenever, but just in case you want another pair of eyes

PR #405, opened 2023-06-06
cache: making it more Go idiomatic

sttts:
An attempt to make it look less clunky.

### DIFF
--- a/pkg/cached/cache.go
+++ b/pkg/cached/cache.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package cache provides a cache mechanism based on etags to lazily
+// Package cached provides a cache mechanism based on etags to lazily
 // build, and/or cache results from expensive operation such that those
 // operations are not repeated unnecessarily. The operations can be
 // created as a tree, and replaced dynamically as needed.
@@ -25,16 +25,18 @@ limitations under the License.
 //
 // This package uses a source/transform/sink model of caches to build
 // the dependency tree, and can be used as follows:
-//   - [NewSource]: A source cache that recomputes the content every time.
-//   - [NewStaticSource]: A source cache that always produces the
+//   - [Func]: A source cache that recomputes the content every time.
+//   - [Once]: A source cache that always produces the
 //     same content, it is only called once.
-//   - [NewTransformer]: A cache that transforms data from one format to
+//   - [Transform]: A cache that transforms data from one format to
 //     another. It's only refreshed when the source changes.
-//   - [NewMerger]: A cache that aggregates multiple caches into one.
+//   - [Merge]: A cache that aggregates multiple caches in a map into one.
 //     It's only refreshed when the source changes.
-//   - [Replaceable]: A cache adapter that can be atomically
-//     replaced with a new one, and saves the previous results in case an
-//     error pops-up.
+//   - [MergeList]: A cache that aggregates multiple caches in a list into one.
+//     It's only refreshed when the source changes.
+//   - [Atomic]: A cache adapter that atomically replaces the source with a new one.
+//   - [LastSuccess]: A cache adapter that caches the last successful and returns
+//     it if the next call fails. It extends [Atomic].
 //
 // # Etags
 //
@@ -54,61 +56,45 @@ import (
 	"sync/atomic"
 )
 
-// Result is the content returned from a call to a cache. It can either
-// be created with [NewResultOK] if the call was a success, or
-// [NewResultErr] if the call resulted in an error.
+// Value is wrapping a value behind a getter for lazy evaluation.
+type Value[T any] interface {
+	Get() (value T, etag string, err error)
+}
+
+// Result is wrapping T and error into a struct for cases where a tuple is more
+// convenient or necessary in Golang.
 type Result[T any] struct {
-	Data T
-	Etag string
-	Err  error
+	Value T
+	Etag  string
+	Err   error
 }
 
-// NewResultOK creates a new [Result] for a successful operation.
-func NewResultOK[T any](data T, etag string) Result[T] {
-	return Result[T]{
-		Data: data,
-		Etag: etag,
-	}
+func (r Result[T]) Get() (T, string, error) {
+	return r.Value, r.Etag, r.Err
 }
 
-// NewResultErr creates a new [Result] when an error has happened.
-func NewResultErr[T any](err error) Result[T] {
-	return Result[T]{
-		Err: err,
-	}
+// Func wraps a (thread-safe) function as a Value[T].
+func Func[T any](fn func() (T, string, error)) Value[T] {
+	return valueFunc[T](fn)
 }
 
-// Result can be treated as a [Data] if necessary.
-func (r Result[T]) Get() Result[T] {
-	return r
+type valueFunc[T any] func() (T, string, error)
+
+func (c valueFunc[T]) Get() (T, string, error) {
+	return c()
 }
 
-// Data is a cache that performs an action whose result data will be
-// cached. It also returns an "etag" identifier to version the cache, so
-// that the caller can know if they have the most recent version of the
-// cache (and can decide to cache some operation based on that).
-//
-// The [NewMerger] and [NewTransformer] automatically handle
-// that for you by checking if the etag is updated before calling the
-// merging or transforming function.
-type Data[T any] interface {
-	// Returns the cached data, as well as an "etag" to identify the
-	// version of the cache, or an error if something happened.
-	Get() Result[T]
+// Static returns constant values.
+func Static[T any](value T, etag string) Value[T] {
+	return Result[T]{Value: value, Etag: etag}
 }
 
-// NewMerger creates a new merge cache, a cache that merges the result
-// of other caches. The function only gets called if any of the
-// dependency has changed.
+// Merge merges a of cached values. The merge function only gets called if any of
+// the dependency has changed.
 //
 // If any of the dependency returned an error before, or any of the
 // dependency returned an error this time, or if the mergeFn failed
-// before, then the function is reran.
-//
-// The caches and results are mapped by K so that associated data can be
-// retrieved. The map of dependencies can not be modified after
-// creation, and a new merger should be created (and probably replaced
-// using a [Replaceable]).
+// before, then the function is run again.
 //
 // Note that this assumes there is no "partial" merge, the merge
 // function will remerge all the dependencies together everytime. Since
@@ -118,18 +104,19 @@ type Data[T any] interface {
 // Also note that Golang map iteration is not stable. If the mergeFn
 // depends on the order iteration to be stable, it will need to
 // implement its own sorting or iteration order.
-func NewMerger[K comparable, T, V any](mergeFn func(results map[K]Result[T]) Result[V], caches map[K]Data[T]) Data[V] {
-	listCaches := make([]Data[T], 0, len(caches))
-	// maps from index to key
+func Merge[K comparable, T, V any](mergeFn func(results map[K]Result[T]) (V, string, error), caches map[K]Value[T]) Value[V] {
+	list := make([]Value[T], 0, len(caches))
+
+	// map from index to key
 	indexes := make(map[int]K, len(caches))
 	i := 0
 	for k := range caches {
-		listCaches = append(listCaches, caches[k])
+		list = append(list, caches[k])
 		indexes[i] = k
 		i++
 	}
 
-	return NewListMerger(func(results []Result[T]) Result[V] {
+	return MergeList(func(results []Result[T]) (V, string, error) {
 		if len(results) != len(indexes) {
 			panic(fmt.Errorf("invalid result length %d, expected %d", len(results), len(indexes)))
 		}
@@ -138,20 +125,11 @@ func NewMerger[K comparable, T, V any](mergeFn func(results map[K]Result[T]) Res
 			m[indexes[i]] = results[i]
 		}
 		return mergeFn(m)
-	}, listCaches)
+	}, list)
 }
 
-type listMerger[T, V any] struct {
-	lock         sync.Mutex
-	mergeFn      func([]Result[T]) Result[V]
-	caches       []Data[T]
-	cacheResults []Result[T]
-	result       Result[V]
-}
-
-// NewListMerger creates a new merge cache that merges the results of
-// other caches in list form. The function only gets called if any of
-// the dependency has changed.
+// MergeList merges a list of cached values. The function only gets called if
+// any of the dependency has changed.
 //
 // The benefit of ListMerger over the basic Merger is that caches are
 // stored in an ordered list so the order of the cache will be
@@ -165,31 +143,37 @@ type listMerger[T, V any] struct {
 // function will remerge all the dependencies together everytime. Since
 // the list of dependencies is constant, there is no way to save some
 // partial merge information either.
-func NewListMerger[T, V any](mergeFn func(results []Result[T]) Result[V], caches []Data[T]) Data[V] {
+func MergeList[T, V any](mergeFn func(results []Result[T]) (V, string, error), delegates []Value[T]) Value[V] {
 	return &listMerger[T, V]{
-		mergeFn: mergeFn,
-		caches:  caches,
+		mergeFn:   mergeFn,
+		delegates: delegates,
 	}
 }
 
+type listMerger[T, V any] struct {
+	lock      sync.Mutex
+	mergeFn   func([]Result[T]) (V, string, error)
+	delegates []Value[T]
+	cache     []Result[T]
+	result    Result[V]
+}
+
 func (c *listMerger[T, V]) prepareResultsLocked() []Result[T] {
-	cacheResults := make([]Result[T], len(c.caches))
+	cacheResults := make([]Result[T], len(c.delegates))
 	ch := make(chan struct {
 		int
 		Result[T]
-	}, len(c.caches))
-	for i := range c.caches {
+	}, len(c.delegates))
+	for i := range c.delegates {
 		go func(index int) {
+			value, etag, err := c.delegates[index].Get()
 			ch <- struct {
 				int
 				Result[T]
-			}{
-				index,
-				c.caches[index].Get(),
-			}
+			}{index, Result[T]{Value: value, Etag: etag, Err: err}}
 		}(i)
 	}
-	for i := 0; i < len(c.caches); i++ {
+	for i := 0; i < len(c.delegates); i++ {
 		res := <-ch
 		cacheResults[res.int] = res.Result
 	}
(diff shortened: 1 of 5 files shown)


pkg/cached/cache.go
```diff
@@ -209,23 +226,21 @@ func NewTransformer[T, V any](transformerFn func(Result[T]) Result[V], source Da
...
 }
 
-// NewSource creates a new cache that generates some data. This
-// will always be called since we don't know the origin of the data and
-// if it needs to be updated or not.
-func NewSource[T any](sourceFn func() Result[T]) Data[T] {
-	c := source[T](sourceFn)
-	return &c
+// DataFunc creates a new cache that always reruns the function.
+func DataFunc[T any](fn func() Result[T]) Data[T] {
```
sttts, line 230: the constructor is not really needed. Compare 
```golang
// The HandlerFunc type is an adapter to allow the use of
// ordinary functions as HTTP handlers. If f is a function
// with the appropriate signature, HandlerFunc(f) is a
// Handler that calls f.
type HandlerFunc func(ResponseWriter, *Request)

// ServeHTTP calls f(w, r).
func (f HandlerFunc) ServeHTTP(w ResponseWriter, r *Request) {
	f(w, r)
}
```

But it looks like with generics, Golang will derive `T` automatically when calling a func, not for a type-cast.

Jefftree:
Assuming you're already aware cc: @apelisse

pkg/cached/cache.go
```diff
@@ -23,12 +23,12 @@ limitations under the License.
 //
 // This package uses a source/transform/sink model of caches to build
 // the dependency tree, and can be used as follows:
-//   - [NewSource]: A source cache that recomputes the content every time.
-//   - [NewStaticSource]: A source cache that always produces the
+//   - [DataFunc]: A source cache that recomputes the content every time.
```
apelisse, line 26: What about just `Func`?

apelisse (commented):
Looks generally good. @Jefftree how do you feel about breaking all the compatibility her? :D

pkg/cached/cache.go
sttts: I like `Func`.

Jefftree:
What's the plan with this PR? afaik there were talks to move this caching code to a different package, are we keeping it in kube-openapi?

sttts:
Would merge it here. Moving it elsewhere is a bigger topic. Let's not be perfect the enemy of good.

apelisse (commented):
Thanks for updating the doc :-)
Generally looks good, none of the tests have been renamed, which I think they should. Also pretty surprised that the Replaceable (whose behavior has changed) tests haven't changed.

pkg/cached/cache.go
```diff
@@ -278,35 +289,41 @@ func (c *static[T]) Get() Result[T] {
 	return c.result
 }
 
-// Replaceable is a cache that carries the result even when the cache is
-// replaced. This is the type that should typically be stored in
-// structs.
-type Replaceable[T any] struct {
-	cache  atomic.Pointer[Data[T]]
-	result atomic.Pointer[Result[T]]
+// Atomic is a cache adapter that can be atomically replaced with a new one.
```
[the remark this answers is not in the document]
Jefftree: +1, this was the only cache that could be replaced without needing to be reinstantiated.

pkg/cached/cache.go
```diff
@@ -55,51 +57,65 @@ import (
...
-	// Returns the cached data, as well as an "etag" to identify the
+	// Get returns the cached data, as well as an "etag" to identify the
 	// version of the cache, or an error if something happened.
 	Get() Result[T]
 }
 
-// NewMerger creates a new merge cache, a cache that merges the result
-// of other caches. The function only gets called if any of the
-// dependency has changed.
+// Merge merges the result of other caches. The function only gets
```
Jefftree, line 117: Question: Is it convention to remove the `New` for constructors? I feel like it's easy to get `Merge` confused with a function that merges vs returning a cache that merges on a `Get()`

pkg/cached/cache.go
```diff
@@ -278,35 +289,41 @@ func (c *static[T]) Get() Result[T] {
...
-		}
-		if c.result.CompareAndSwap(cResult, &result) {
-			return result
-		}
-	}
+func (x *Atomic[T]) Store(val Data[T]) *Atomic[T] { x.value.Store(&val); return x }
+func (x *Atomic[T]) Get() Result[T]               { return (*x.value.Load()).Get() }
+
+// NewLastSuccess creates a new LastSuccess cache.
+func NewLastSuccess[T any]() *LastSuccess[T] {
```
Jefftree, line 301: This seems to follow a different conventions than the previous caches with using `New` as part of the constructor?

pkg/cached/cache.go
```diff
@@ -226,40 +241,36 @@ func (c *listMerger[T, V]) Get() Result[V] {
...
+type dataFunc[T any] func() Result[T]
 
-func (c *source[T]) Get() Result[T] {
-	return (*c)()
+func (c dataFunc[T]) Get() Result[T] {
+	return c()
 }
 
-// NewStaticSource creates a new cache that always generates the
+// Cache creates a new cache that always generates the
```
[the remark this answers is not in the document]
Jefftree: Using this name `Cache` in a package called `cached` makes it a bit ambiguous. It's hard to tell what this function does without looking at the source code. This might also be misleading for the default way of using this package. I would probably prefer something like `StaticFunc` or `Once`.

pkg/cached/cache.go
```diff
@@ -55,51 +57,65 @@ import (
...
+// NewError creates a new [Result] when an error has happened.
+func NewError[T any](err error) Result[T] {
 	return Result[T]{
 		Err: err,
 	}
 }
 
-// Result can be treated as a [Data] if necessary.
+// Errorf creates a new [Result] when an error has happened.
+func Errorf[T any](s string, params ...any) Result[T] {
```
Jefftree, line 84: `NewErrorf`? `Errorf` feels inconsistent with `NewError`

Jefftree:
> Thanks for updating the doc :-) Generally looks good, none of the tests have been renamed, which I think they should. Also pretty surprised that the Replaceable (whose behavior has changed) tests haven't changed.

+1 would be great if the tests are also renamed. They'll probably be a good reference for those looking to use this library. `Replaceable` and `LastSuccess` behavior is identical so it makes sense that the test hasn't changed, but if we're introducing a new `Atomic` cache then we should probably have a basic test for it.

pkg/cached/cache.go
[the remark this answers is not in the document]
sttts: LazilyOnce ?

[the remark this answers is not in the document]
sttts: Once is also fine. I like that.

pkg/cached/cache.go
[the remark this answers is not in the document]
sttts: it's the atomic value interface after all.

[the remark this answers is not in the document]
sttts: I found Replacable confusing. Why should a cache be replaced?

pkg/cached/cache.go
```diff
@@ -278,35 +289,41 @@ func (c *static[T]) Get() Result[T] {
...
-	for {
-		cResult := c.result.Load()
-		if result.Err != nil && cResult != nil && cResult.Err == nil {
-			return *cResult
-		}
-		if c.result.CompareAndSwap(cResult, &result) {
-			return result
-		}
-	}
+func (x *Atomic[T]) Store(val Data[T]) *Atomic[T] { x.value.Store(&val); return x }
```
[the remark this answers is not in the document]
sttts: I thought we do in the proof PR. But we don't. And the atomic package types don't either. Removing.

pkg/cached/cache_test.go
```diff
@@ -698,16 +695,15 @@ func TestReplaceableAlternateError(t *testing.T) {
 }
 
 func TestReplaceableWithTransformer(t *testing.T) {
-	replaceable := cached.Replaceable[[]byte]{}
-	replaceable.Replace(cached.NewResultOK([]byte("source"), "source"))
+	replaceable := cached.NewLastSuccess[[]byte]().Store(cached.NewResult([]byte("source"), "source"))
```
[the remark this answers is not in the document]
sttts: no, that's just the constructor chained with `Store`, which is still possible after https://github.com/kubernetes/kube-openapi/pull/405#discussion_r1309920400.

pkg/cached/cache.go
sttts: These are logically not constructors, but verbs. That's the whole point of the PR. Promise-like code that mocks a non-promise-like interface. It's a detail that behind the scenes constructors create a datastructure.

pkg/cached/cache.go
sttts: updated.

pkg/cached/cache.go
```diff
@@ -226,87 +241,118 @@ func (c *listMerger[T, V]) Get() Result[V] {
...
+	return c()
 }
 
-// NewStaticSource creates a new cache that always generates the
-// same data. This will only be called once (lazily).
-func NewStaticSource[T any](staticFn func() Result[T]) Data[T] {
-	return &static[T]{
-		fn: staticFn,
+// Once calls Data[T].Get() lazily and only once, even in case of an error result.
+func Once[T any](d Data[T]) Data[T] {
```
Jefftree, line 272: Is lazily implied? Every promise here calls the underlying func lazily?

pkg/cached/cache_test.go
```diff
@@ -765,60 +765,61 @@ func TestReplaceableWithTransformer(t *testing.T) {
...
 	// This cache is computed once, and is not lazy at all.
-	three := cached.NewResultOK([]byte("three"), "three")
+	three := cached.NewResult([]byte("three"), "three")
 
 	// This cache will allow us to replace a branch of the tree
 	// efficiently.
-	replaceable := cached.Replaceable[[]byte]{}
-	replaceable.Replace(cached.NewResultOK([]byte("four"), "four"))
+
+	replaceable := cached.R(cached.LastSuccess[[]byte])(&cached.Atomic[[]byte]{})
```
Jefftree, line 784: Personally this feels like a very complicated way to accomplish what we want with replaceable, but will defer to @apelisse.

pkg/cached/cache.go
```diff
@@ -55,51 +57,65 @@ import (
...
 	return r
 }
 
+// Unpack return the individual elements of the result, for more
+// convenient usage:
+//
+//	if value, _, err := c.Get().Unpack(); err != nil {
+//	  ...
+//	}
+func (r Result[T]) Unpack() (T, string, error) {
```
Jefftree, line 99: though it's pretty intuitive, would be nice to see an example of usage in the tests

pkg/cached/cache.go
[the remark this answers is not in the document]
Jefftree: I do agree that this is basically the atomic value interface, but it's different than the "atomic" concept". I don't really have a better suggestion though.

apelisse (commented):
I have a few nits, or discomfort, but I also don't care very much about the names.

pkg/cached/cache.go
```diff
@@ -25,16 +25,18 @@ limitations under the License.
...
 //     another. It's only refreshed when the source changes.
-//   - [NewMerger]: A cache that aggregates multiple caches into one.
+//   - [Merge]: A cache that aggregates multiple caches in a map into one.
 //     It's only refreshed when the source changes.
-//   - [Replaceable]: A cache adapter that can be atomically
-//     replaced with a new one, and saves the previous results in case an
-//     error pops-up.
+//   - [MergeList]: A cache that aggregates multiple caches in a list into one.
+//     It's only refreshed when the source changes.
+//   - [Atomic]: A cache adapter that atomically replaces the source with a new one.
```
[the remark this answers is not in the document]
Jefftree: @sttts If we're using this name because of the atomic value interface, would something like `Value` be a better fit here?

pkg/cached/cache_test.go
Jefftree: Logically this usage makes sense, I'm just not sure about use case of having this kind of function since almost all our current use cases involve `cached.R` with LastSuccess and Atomic. Are there already use cases in mind for other invocations of `cached.R`? Perhaps it would be beneficial to have shorter alias to create this "Replaceable" cache since it's used so often.

sttts: Two ways forward:
1. ```golang
   replaceable := cached.Atomic{}
   foo.WireInto(LastSuccess(replaceable))
   replaceable.Store(something)
   ```
2. introduce a custom LastSuccess variant `Replaceable` implementing data type.

pkg/cached/cache.go
sttts: Tuples are idiomatic for return values, not structs that include errors. This lets us stay with tubles, but have a nice way to call it in a Go like way:
```golang
if value, etag, err := something.Get().Unpack(); err != nil {
  ...
}
```
Adding it to the tests.

pkg/cached/cache.go
```diff
@@ -25,16 +25,18 @@ limitations under the License.
...
-//   - [NewTransformer]: A cache that transforms data from one format to
+//   - [Transform]: A cache that transforms data from one format to
 //     another. It's only refreshed when the source changes.
-//   - [NewMerger]: A cache that aggregates multiple caches into one.
+//   - [Merge]: A cache that aggregates multiple caches in a map into one.
 //     It's only refreshed when the source changes.
-//   - [Replaceable]: A cache adapter that can be atomically
-//     replaced with a new one, and saves the previous results in case an
-//     error pops-up.
+//   - [MergeList]: A cache that aggregates multiple caches in a list into one.
```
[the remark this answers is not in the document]
sttts: I prefer the focus on the verbs. `MergeMap` would be an alternative. No strong opinion. I think I prefer `Merge`.

pkg/cached/cache.go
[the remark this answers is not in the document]
sttts: `Mutable` or `State` or `Variable` would be the right term. A value can be immutable.

@apelisse I tend to disagree: Golang has a meaning for an atomic value. It's not about how it works (no idea how it actually does under the hood), but about semantics of state that can be replaced without locking:
```
// A Bool is an atomic boolean value.
// The zero value is false.
type Bool struct {
	_ noCopy
	v uint32
}
```

pkg/cached/cache.go
sttts: Alternative: `Get() (value T, etag string, err error)`. Which one do you prefer?

pkg/cached/cache.go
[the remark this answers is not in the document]
Jefftree: Also +1 on keeping the focus on the verbs

pkg/cached/cache_test.go
Jefftree: No strong preferences, 2 would be slightly more convenient from the perspective of a user who would call this function a bunch.

sttts: have done 2 now. I like it.

pkg/cached/cache.go
sttts: have done the variant with `Get` now. IMO this is much more natural.

pkg/cached/cache.go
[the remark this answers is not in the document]
apelisse: I don't know if we disagree, we're just not at the same layer of abstraction. How it works and what it does changes based on the layer of abstraction that you're referring to :-). IOW, I agree that the Atomic object is what it is, not how it does it. But Replaceable is something that can be replaced, no matter how it's done (atomic or not).

pkg/cached/cache.go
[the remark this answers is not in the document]
apelisse: Sounds good to me.

pkg/cached/cache.go
```diff
@@ -54,61 +56,45 @@ import (
 	"sync/atomic"
 )
 
-// Result is the content returned from a call to a cache. It can either
-// be created with [NewResultOK] if the call was a success, or
-// [NewResultErr] if the call resulted in an error.
+// Value is wrapping a value behind a getter for lazy evaluation.
+type Value[T any] interface {
+	Get() (value T, etag string, err error)
+}
```
apelisse, line 62: I think I had this initially lol

pkg/cached/cache.go
```diff
@@ -54,61 +56,45 @@ import (
...
-// The [NewMerger] and [NewTransformer] automatically handle
-// that for you by checking if the etag is updated before calling the
-// merging or transforming function.
-type Data[T any] interface {
-	// Returns the cached data, as well as an "etag" to identify the
-	// version of the cache, or an error if something happened.
-	Get() Result[T]
+// Static returns constant values.
+func Static[T any](value T, etag string) Value[T] {
+	return Result[T]{Value: value, Etag: etag}
```
apelisse, line 89: That's not really useful? I don't have a strong opinion though, the same thing was more or less here before.

apelisse (commented):
I think that's fine, happy to lgtm, but I'll let Jeff do it.

pkg/cached/cache.go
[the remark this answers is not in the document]
sttts: Atomic cache 😁

I still disagree. Will you be in Chicago? We need a 🍻 to settle that discussion.

pkg/cached/cache.go
sttts: It‘s convenient as function calls in Go don‘t need the explicit type parameter. But I agree, one could live without

pkg/cached/cache_test.go
```diff
@@ -31,139 +31,138 @@ import (
...
+	if _, _, err := source.Get(); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
 }
 
-func TestResultError(t *testing.T) {
-	source := cached.NewResultErr[[]byte](errors.New("source error"))
-	if err := source.Get().Err; err == nil {
+func TestNewError(t *testing.T) {
+	source := cached.Result[[]byte]{Err: errors.New("source error")}
```
Jefftree, line 139: Is this fine as a way for instantiating errors? I noticed that the `NewError` and `Errorf` functions were removed.

pkg/handler/handler.go
```diff
@@ -59,52 +60,52 @@ type timedSpec struct {
 // OpenAPIService is the service responsible for serving OpenAPI spec. It has
 // the ability to safely change the spec while serving it.
 type OpenAPIService struct {
-	specCache  cached.Replaceable[*spec.Swagger]
-	jsonCache  cached.Data[timedSpec]
-	protoCache cached.Data[timedSpec]
+	specCache  cached.LastSuccess[*spec.Swagger]
```
Jefftree, line 63: What's our goal with the Replaceable interface? I noticed Atomic implements it, should LastSuccess also follow it?

Jefftree (commented):
Two questions otherwise lgtm

pkg/handler/handler.go
sttts: added a type assertion for `LastSuccess`. Done.

pkg/cached/cache_test.go
sttts: you usually won't have to construct these results as we have switches to tripples as return values.

Have fixed these two tests to reference what they do: test `Get` of `Result`.

Done.

sttts:
@Jefftree addressed your two comments. And squashed. Ready to go 🙏

Jefftree:
/lgtm
/approve

Thanks for all the rounds of iteration 🎉

PR #407, opened 2023-06-12
generators: Default is a valid enum value

### LINKED ISSUE #395: Generated enum fails validation when used for CRD creation
https://github.com/kubernetes/kube-openapi/blob/78281498afbbfb6fdabe43b8b6a7898f75892c44/pkg/generators/openapi_test.go#L1618

This shows an example of generated code:

https://github.com/kubernetes/kube-openapi/blob/78281498afbbfb6fdabe43b8b6a7898f75892c44/pkg/generators/openapi_test.go#L1655-L1662

However if we use this code to generate CRD it fails validation. The error message is like:

```
CustomResourceDefinition.apiextensions.k8s.io "..." is invalid: spec.validation.openAPIV3Schema...default: Unsupported value: "": supported values: "a", "b"
```

apelisse:
Fixes #395

This makes sure that the default value is part of the enum values, which makes sense. We shouldn't have a default that doesn't validate. There are probably a lot of other thing that don't exist right now but will exist in the future that we might want to validate for (maximum/minimum comes to mind). I'm also a little confused how we can have a default for a required value, we might want to disallow that.

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -585,26 +585,27 @@ func mustEnforceDefault(t *types.Type, omitEmpty bool) (interface{}, error) {
 	}
 }
 
-func (g openAPITypeWriter) generateDefault(comments []string, t *types.Type, omitEmpty bool) error {
+func (g openAPITypeWriter) generateDefault(comments []string, t *types.Type, omitEmpty bool) (*string, error) {
 	t = resolveAliasAndEmbeddedType(t)
 	def, err := defaultFromComments(comments)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	if enforced, err := mustEnforceDefault(t, omitEmpty); err != nil {
-		return err
+		return nil, err
 	} else if enforced != nil {
 		if def == nil {
 			def = enforced
 		} else if !reflect.DeepEqual(def, enforced) {
 			enforcedJson, _ := json.Marshal(enforced)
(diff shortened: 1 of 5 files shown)


apelisse:
cc @alexzielenski

pkg/generators/openapi.go
```diff
@@ -676,9 +677,16 @@ func (g openAPITypeWriter) generateProperty(m *types.Member, parent *types.Type)
...
 	}
 	omitEmpty := strings.Contains(reflect.StructTag(m.Tags).Get("json"), "omitempty")
-	if err := g.generateDefault(m.CommentLines, m.Type, omitEmpty); err != nil {
+	def, err := g.generateDefault(m.CommentLines, m.Type, omitEmpty)
+	if err != nil {
 		return fmt.Errorf("failed to generate default in %v: %v: %v", parent, m.Name, err)
 	}
+	if def != nil {
+		g.Do("Default: $.$,\n", *def)
+		if enumType, isEnum := g.enumContext.EnumType(m.Type); isEnum && !enumType.IsValid(*def) {
```
alexzielenski, line 686: should this check be part of `generateDefault` implementation instead? Just wondering since it is called from multiple places and might be cleaner that way

apelisse: it would mean we would have to push even more parameters to `generateDefault`, I don't have a strong opinion.

apelisse: Also, I don't see `IsValid` being called in other places?

Jefftree:
@alexzielenski I don't have much context on this, is it something we still want to get in?

PR #410, opened 2023-06-15
WIP: v2 builder: reference shared parameters instead of duplication

sttts:


pkg/builder/openapi.go
```diff
@@ -100,6 +102,7 @@ func newOpenAPI(config *common.Config) openAPI {
 				Responses:   config.ResponseDefinitions,
 				Paths:       &spec.Paths{Paths: map[string]spec.PathItem{}},
 				Info:        config.Info,
+				Parameters:  config.SharedParameters,
```
Jefftree, line 105: I know this is a WIP PR, but the parameters shouldn't be defined in the config imo. We should be able to calculate the set of parameters purely through the building process.

apelisse:
I don't see what's terrible about this approach to be honest?

sttts:
Replaced by https://github.com/kubernetes/kube-openapi/pull/411

PR #412, opened 2023-07-14
validate: dont short circuit validator on MinProperties/MaxProperties error

alexzielenski:
Currently `object_validator` does not explore any of the fields of an object if the number of fields violates any MinProperties/MaxProperties condition. This is inconsistent with slice_validator, which validates MinItems/MaxItems after visiting any of the items.

This prevents us from seeing all errors if MinProperties/MaxProperties is violated.

This is the only location in validation where a part of the input object is not explored/short circuited. Having this property would greatly simplify logic for map-type aware DeepEqual logic for CRD validation ratcheting. The map-type aware DeepEqual uses the tree traversed by SchemaValidator to inform which schema is associated with which subobject. Unfortunately if there is a MaxProperties/MinProperties error a portion of the tree is missing.

This change makes the validator keep going after seeing an error with number of properties and validate each of the subfields anyway, just as it does for arrays.

### DIFF
--- a/pkg/validation/validate/object_validator_test.go
+++ b/pkg/validation/validate/object_validator_test.go
@@ -18,6 +18,9 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	kubeopenapierrors "k8s.io/kube-openapi/pkg/validation/errors"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"k8s.io/kube-openapi/pkg/validation/strfmt"
 )
 
 func itemsFixture() map[string]interface{} {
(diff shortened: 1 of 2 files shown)


apelisse:
That looks a lot better, a test would be good?

sttts:
/approve

Leaving lgtm to @apelisse after a test is added.

alexzielenski:
added test

/label tide/merge-method-squash

pkg/validation/validate/object_validator_test.go
```diff
@@ -80,3 +83,69 @@ func TestObjectValidator_EdgeCases(t *testing.T) {
...
+		},
+	}
+
+	obj := map[string]interface{}{
+		"field": "hello, world",
+	}
+	res := s.Validate(obj)
+
+	assert.ElementsMatch(t, []*kubeopenapierrors.Validation{
+		kubeopenapierrors.TooFewProperties(s.Path, s.In, *s.MinProperties, int64(len(obj))),
```
apelisse, line 124: Hard-coding in tests is a lot better IMO
```suggestion
		kubeopenapierrors.TooFewProperties("some.path[5]", "body", 20, 1),
```

apelisse (commented):
/lgtm

thanks

PR #415, opened 2023-07-16
builder/parameters: switch to fnv and exclude fieldValidation query parameter

sttts:
- no need for a cryptographic hash. fnv-1 is a magnitude faster.
- pre-1.28 `kubectl`'s incomplete OpenAPI implementation needs a non-shared `fieldValidation` query parameter.
- pre-1.27 `kubectl` needs a non-shared `dryRun` query parameter (compare https://github.com/kubernetes/kubernetes/pull/114294/commits).

### DIFF
--- a/pkg/builder/parameters.go
+++ b/pkg/builder/parameters.go
@@ -17,10 +17,10 @@ limitations under the License.
 package builder
 
 import (
-	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"hash/fnv"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/builder/parameters.go
```diff
@@ -64,6 +64,12 @@ func collectSharedParameters(sp *spec.Swagger) (namesByJSON map[string]string, r
 		if (p.In == "query" || p.In == "path") && p.Name == "name" {
 			return nil // ignore name parameter as they are never shared with the Kind in the description
 		}
+		if p.In == "query" && p.Name == "fieldValidation" {
+			return nil // keep fieldValidation parameter unshared because kubectl uses it (until 1.27) to detect SSA support
```
liggitt, line 68:
```suggestion
			return nil // keep fieldValidation parameter unshared because kubectl uses it (until 1.27) to detect server-side field validation support
```

liggitt:
change lgtm, one nit on the comment

PR #416, opened 2023-08-02
Implement experimental marshaler for OpenAPI V3

Jefftree:
```
goos: linux
goarch: amd64
pkg: k8s.io/kube-openapi/pkg/spec3
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
                                               │  old1.txt   │              new1.txt              │
                                               │   sec/op    │   sec/op     vs base               │
OpenAPIV3Serialize/appsv1spec.json-24            30.60m ± 2%   12.41m ± 3%  -59.43% (p=0.002 n=6)
OpenAPIV3Serialize/authorizationv1spec.json-24   2.541m ± 1%   1.143m ± 1%  -55.04% (p=0.002 n=6)
geomean                                          8.818m        3.766m       -57.29%

                                               │   old1.txt    │              new1.txt               │
                                               │     B/op      │     B/op      vs base               │
OpenAPIV3Serialize/appsv1spec.json-24            10.586Mi ± 1%   5.175Mi ± 3%  -51.11% (p=0.002 n=6)
OpenAPIV3Serialize/authorizationv1spec.json-24    931.7Ki ± 0%   538.6Ki ± 0%  -42.19% (p=0.002 n=6)
geomean                                           3.103Mi        1.650Mi       -46.83%

                                               │  old1.txt   │              new1.txt              │
                                               │  allocs/op  │  allocs/op   vs base               │
OpenAPIV3Serialize/appsv1spec.json-24            34.00k ± 0%   20.03k ± 0%  -41.08% (p=0.002 n=6)
OpenAPIV3Serialize/authorizationv1spec.json-24   3.021k ± 0%   2.025k ± 0%  -32.96% (p=0.002 n=6)
geomean                                          10.13k        6.369k       -37.15%
```

Slightly less improvement than OpenAPI V2 which is somewhat expected because of the smaller size of OpenAPI V3.

### DIFF
--- a/pkg/handler3/handler_test.go
+++ b/pkg/handler3/handler_test.go
@@ -39,7 +39,7 @@ var returnedOpenAPI = []byte(`{
    "title": "Kubernetes",
    "version": "v1.23.0"
   },
-  "paths": {"x-kubernetes-extension": "val"}}`)
+  "paths": {}}`)
 
 func TestRegisterOpenAPIVersionedService(t *testing.T) {
 	var s *spec3.OpenAPI
--- a/pkg/handler3/handler_test.go
+++ b/pkg/handler3/handler_test.go
@@ -39,7 +39,7 @@ var returnedOpenAPI = []byte(`{
    "title": "Kubernetes",
    "version": "v1.23.0"
   },
-  "paths": {}}`)
+  "paths": {"x-kubernetes-extension": "val"}}`)
 
 func TestRegisterOpenAPIVersionedService(t *testing.T) {
 	var s *spec3.OpenAPI
--- a/pkg/spec3/fuzz.go
+++ b/pkg/spec3/fuzz.go
@@ -169,6 +169,21 @@ var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
 		c.Fuzz(&v.ResponseProps)
 		c.Fuzz(&v.VendorExtensible)
 	},
+	func(v *Operation, c fuzz.Continue) {
+		c.FuzzNoCustom(v)
+		// Do not fuzz null values into the array.
+		for i, val := range v.SecurityRequirement {
+			if val == nil {
+				v.SecurityRequirement[i] = make(map[string][]string)
+			}
+
+			for k, v := range val {
+				if v == nil {
+					val[k] = make([]string, 0)
+				}
+			}
+		}
+	},
 	func(v *spec.Extensions, c fuzz.Continue) {
 		numChildren := c.Intn(5)
 		for i := 0; i < numChildren; i++ {
--- a/pkg/spec3/parameter.go
+++ b/pkg/spec3/parameter.go
@@ -36,6 +36,9 @@ type Parameter struct {
 
 // MarshalJSON is a custom marshal function that knows how to encode Parameter as JSON
 func (p *Parameter) MarshalJSON() ([]byte, error) {
+	if internal.UseOptimizedJSONMarshalingV3 {
+		return internal.DeterministicMarshal(p)
+	}
 	b1, err := json.Marshal(p.Refable)
 	if err != nil {
 		return nil, err
@@ -49,6 +52,18 @@ func (p *Parameter) MarshalJSON() ([]byte, error) {
 		return nil, err
 	}
 	return swag.ConcatJSON(b1, b2, b3), nil
+}
+
+func (p *Parameter) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	var x struct {
+		Ref            string                 `json:"$ref,omitempty"`
+		ParameterProps parameterPropsOmitZero `json:",inline"`
+		spec.Extensions
+	}
+	x.Ref = p.Refable.Ref.String()
+	x.Extensions = internal.SanitizeExtensions(p.Extensions)
+	x.ParameterProps = parameterPropsOmitZero(p.ParameterProps)
+	return opts.MarshalNext(enc, x)
 }
 
 func (p *Parameter) UnmarshalJSON(data []byte) error {
@@ -113,4 +128,20 @@ type ParameterProps struct {
 	Example interface{} `json:"example,omitempty"`
 	// Examples of the parameter's potential value. Each example SHOULD contain a value in the correct format as specified in the parameter encoding
 	Examples map[string]*Example `json:"examples,omitempty"`
+}
+
+type parameterPropsOmitZero struct {
+	Name            string                `json:"name,omitempty"`
+	In              string                `json:"in,omitempty"`
+	Description     string                `json:"description,omitempty"`
+	Required        bool                  `json:"required,omitzero"`
+	Deprecated      bool                  `json:"deprecated,omitzero"`
+	AllowEmptyValue bool                  `json:"allowEmptyValue,omitzero"`
+	Style           string                `json:"style,omitempty"`
+	Explode         bool                  `json:"explode,omitzero"`
+	AllowReserved   bool                  `json:"allowReserved,omitzero"`
+	Schema          *spec.Schema          `json:"schema,omitzero"`
+	Content         map[string]*MediaType `json:"content,omitempty"`
+	Example         interface{}           `json:"example,omitempty"`
+	Examples        map[string]*Example   `json:"examples,omitempty"`
 }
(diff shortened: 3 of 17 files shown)


pkg/spec3/path.go
```diff
@@ -35,15 +35,41 @@ type Paths struct {
 
 // MarshalJSON is a custom marshal function that knows how to encode Paths as JSON
 func (p *Paths) MarshalJSON() ([]byte, error) {
```
Jefftree, line 37: This function was updated because we did not filter for the "/" prefix before.

apelisse: you could have split that in a separate PR/commit maybe? Maybe if we want to backport? I don't think it matters.

pkg/handler3/handler_test.go
```diff
@@ -39,7 +39,7 @@ var returnedOpenAPI = []byte(`{
    "title": "Kubernetes",
    "version": "v1.23.0"
   },
-  "paths": {}}`)
+  "paths": {"x-kubernetes-extension": "val"}}`)
```
apelisse, line 42: Why is that happening?

pkg/spec3/fuzz.go
```diff
@@ -169,6 +169,21 @@ var OpenAPIV3FuzzFuncs []interface{} = []interface{}{
 		c.Fuzz(&v.ResponseProps)
 		c.Fuzz(&v.VendorExtensible)
 	},
+	func(v *Operation, c fuzz.Continue) {
```
apelisse, line 172: I need to think twice about this, but I don't have time right now.

apelisse (commented):
Thanks,
I've compared this to the v2 versino of the same PR and looked at each individual comments I got on that other PR. I checked all the fields to make sure nothing was missing. Looks good to me, two minor questions/comments.

pkg/spec3/fuzz.go
Jefftree: This specifically was copied from the v2 fuzzer https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/spec/fuzz.go#L68-L80

apelisse:
I'm personally good with this, lgtm & approve for me.
Anybody else wants to have a look @sttts @liggitt ?

pkg/handler3/handler_test.go
liggitt: agreed... what's this about?

Jefftree: Removed. There was a bug with the round tripping for paths earlier.

pkg/spec3/spec.go
```diff
@@ -48,3 +48,30 @@ func (o *OpenAPI) UnmarshalJSON(data []byte) error {
...
+		Servers      []*Server              `json:"servers,omitempty"`
+		Components   *Components            `json:"components,omitzero"`
+		ExternalDocs *ExternalDocumentation `json:"externalDocs,omitzero"`
+	}
+	x.Version = o.Version
+	x.Info = o.Info
+	x.Paths = o.Paths
+	x.Servers = o.Servers
+	x.Components = o.Components
+	x.ExternalDocs = o.ExternalDocs
```
alexzielenski, line 75: Is the only difference between this struct and `OpenAPI` struct the `omitzero` properties?

If that is the case I think it would be possible to avoid a copy here by casting `*OpenAPI` to `*OpenAPIOmitZero`:

Following toy example seemed to compile at least:
```go

type X struct {
	A string
	B string
}

type Y struct {
	A string
	B string
}

func (x X) ConvertToY() Y {
	return Y(x)
}

func (x *X) ConvertToYPointer() *Y {
	return (*Y)(x)
}
```

pkg/spec3/response.go
```diff
@@ -194,6 +219,18 @@ func (r *Response) MarshalJSON() ([]byte, error) {
 	return swag.ConcatJSON(b1, b2, b3), nil
 }
 
+func (r Response) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	var x struct {
+		Ref string `json:"$ref,omitempty"`
+		spec.Extensions
+		ResponseProps `json:",inline"`
```
alexzielenski, line 226: These extra structs caused by `Refable` and `VendorExtensible` make me sad but do not see a way around them without breaking changes to the type.

pkg/spec3/parameter.go
```diff
@@ -51,6 +54,18 @@ func (p *Parameter) MarshalJSON() ([]byte, error) {
...
 
+func (p *Parameter) MarshalNextJSON(opts jsonv2.MarshalOptions, enc *jsonv2.Encoder) error {
+	var x struct {
+		Ref            string                 `json:"$ref,omitempty"`
+		ParameterProps parameterPropsOmitZero `json:",inline"`
+		spec.Extensions
+	}
+	x.Ref = p.Refable.Ref.String()
+	x.Extensions = internal.SanitizeExtensions(p.Extensions)
+	x.ParameterProps = parameterPropsOmitZero(p.ParameterProps)
```
alexzielenski, line 65: Could changing these `parameterPropsOmitZero`, `operationsPropsOmitZero`, etc. types to a pointer avoid a copy and boost performance at all?

pkg/spec3/path.go
alexzielenski: why was it a problem for paths from the input object to be serialized

alexzielenski (commented):
mostly lgtm with few nits/questions

pkg/spec3/path.go
Jefftree: It wasn't a problem, but would bring inconsistency between jsonv2 and jsonv1 marshalling

pkg/spec3/parameter.go
Jefftree: I tried something like

```
    var x struct {
        Ref            string                 `json:"$ref,omitempty"`
+       ParameterProps *parameterPropsOmitZero `json:",inline"`
        spec.Extensions
    }
    ...
+   tmp := parameterPropsOmitZero(p.ParameterProps)
+   x.ParameterProps = &tmp
    ...
```

and performance numbers look identical. Is this what you mean by changing them to pointers?

apelisse: Yeah, Jordan called out the same thing in my PR for V2, and I ran the same experiment :-)

pkg/spec3/spec.go
apelisse: isn't the conversion making a copy anyway?

apelisse: It sounds like you're suggesting doing:
```suggestion
	type X struct {
		Version      string                 `json:"openapi"`
		Info         *spec.Info             `json:"info"`
		Paths        *Paths                 `json:"paths,omitzero"`
		Servers      []*Server              `json:"servers,omitempty"`
		Components   *Components            `json:"components,omitzero"`
		ExternalDocs *ExternalDocumentation `json:"externalDocs,omitzero"`
	}
	x := X(o)
```

alexzielenski: Yes, except `x = (*X)(&o)`

pkg/spec3/parameter.go
alexzielenski: I meant using like this:

`x.ParameterProps = (*parameterPropsOmitZero)(&p.ParameterProps)`

apelisse: https://github.com/kubernetes/kube-openapi/pull/325#discussion_r1028573872

pkg/spec3/spec.go
apelisse: it's definitely simpler.

pkg/spec3/parameter.go
Jefftree:
```
| Name                                               | Before       | After        |
| (appsv1)OpenAPIV3Deserialize/jsonv2-24             | 13.68m ± 1%  | 13.65m ± 1%  |
| (authorizationv1)OpenAPIV3Deserialize/jsonv2#01-24 | 1.199m ± 1%  | 1.192m ± 2%  |
| (appsv1)OpenAPIV3Deserialize/jsonv2-24             | 5.260Mi ± 0% | 5.260Mi ± 0% |
| (authorizationv1)OpenAPIV3Deserialize/jsonv2#01-24 | 525.4Ki ± 0% | 525.4Ki ± 0% |
| (appsv1)OpenAPIV3Deserialize/jsonv2-24             | 26.89k ± 0%  | 26.89k ± 0%  |
| (authorizationv1)OpenAPIV3Deserialize/jsonv2#01-24 | 2.517k ± 0%  | 2.517k ± 0%  |
```

alexzielenski: Yeah doesn’t seem to do anything

PR #417, opened 2023-08-11
Fix type in handler3/handler_test

Jefftree:


Jefftree:
/assign @apelisse

I wonder why the typos are independently in both files

PR #418, opened 2023-08-15
Use gomod version

apelisse:
Remove the matrix testing that we constantly have to update and let's use the version that is specified in go.mod.

There's also a `go.mod` in `./test/integration/go.mod`, not sure we're going to remember to udpate that one.

cc @alexzielenski @liggitt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,10 @@
 name: ci-pipeline
-
 on:
   push:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-
 jobs:
-
   ci:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,26 +1,18 @@
...
   pull_request:
     branches: [ master ]
-
 jobs:
-
   ci:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        go: [ '1.19', '1.x' ]
```
liggitt, line 15: stable / old-stable would dynamically match both supported versions of go and not need updating, right?

PR #419, opened 2023-08-18
Adds support for Symbol references in defaults

alexzielenski:
gengo's Defaulter-gen has support for symbol references to be used as defaults in the form `+default=ref(package/of/symbol.Name)`

This PR exposes defaults declared as symbols into the OpenAPI

depends on https://github.com/kubernetes/gengo/pull/247


pkg/generators/openapi.go
```diff
@@ -553,19 +557,27 @@ func (g openAPITypeWriter) validatePatchTags(m *types.Member, parent *types.Type
...
+		return nil, &parsedName, nil
+	} else if err := json.Unmarshal([]byte(tag), &i); err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal default: %v", err)
 	}
-	return i, nil
+	return i, nil, nil
 }
 
 func mustEnforceDefault(t *types.Type, omitEmpty bool) (interface{}, error) {
+	// If t implements custom JSON marshalling, all of this logic is likely wrong
```
apelisse, line 575: yeah, I think that's fine.

pkg/generators/openapi_test.go
```diff
@@ -1693,3 +1694,80 @@ Extensions: spec.Extensions{
...
+					"Value": {
+						SchemaProps: spec.SchemaProps{
+							Default: "A Default Value",
+							Type:    []string{"string"},
+							Format:  "",
+						},
+					},
+					"OtherValue": {
+						SchemaProps: spec.SchemaProps{
+							Default: MyConst,
```
apelisse, line 1745: is the variable actually copied into the openapi file that has the schema?

apelisse: you might have to fully qualify the local package when you do that since I don't think the generate openapi is going in the same package as the original types.

pkg/generators/openapi.go
alexzielenski: I think it is related to https://github.com/kubernetes/kubernetes/issues/118771

apelisse: Agreed, sorry, what I meant was "Yeah, that looks like a reasonable way to address some of the things you've been talking about". Yeah, I agree with you.

pkg/generators/openapi_test.go
alexzielenski: Just tested on k8s upstream, it has the correct package since the ImportNamer takes in the path to the output package - it knows the output is different from the package owning the comment.

pkg/generators/openapi.go
apelisse: Should we disregard all defaults for these types, including the ones in comments?

apelisse (commented):
Thanks, I'd feel a lot better if we had an integration test though.

pkg/generators/openapi.go
alexzielenski: I will open a separate Pr for this change to keep things orthogonal

alexzielenski:
added integration test, removed change to `mustEnforceDefault` we were unsure about

test/integration/testdata/golden.v2.json
```diff
@@ -670,6 +670,26 @@
...
+      "type": "string",
+      "default": "SymbolConstant"
+     },
+     "localSymbolReference": {
+      "type": "string",
+      "default": "SymbolConstant"
+     },
+     "pointerConversionSymbolReference": {
+      "type": "string",
+      "default": "apple"
```
apelisse, line 692: Very cool to have the end-to-end test show that we end-up with the properly resolved values injected in the openapi! (big belated thanks to @seans3 for having written the test that makes the end-to-end possible).

test/integration/testdata/defaults/default.go
```diff
@@ -15,8 +15,23 @@ type Defaulted struct {
...
+
+	// +default=ref(ConstantValue)
+	LocalSymbolReference string `json:"localSymbolReference,omitempty"`
+	// +default=ref(k8s.io/kube-openapi/test/integration/testdata/defaults.ConstantValue)
+	FullyQualifiedSymbolReference string `json:"fullyQualifiedSymbolReference,omitempty"`
+	// +default=ref(k8s.io/kube-openapi/test/integration/testdata/enumtype.FruitApple)
+	ExternalSymbolReference string `json:"externalSymbolReference,omitempty"`
+	// +default=ref(k8s.io/kube-openapi/test/integration/testdata/enumtype.FruitApple)
+	PointerConversionSymbolReference *DefaultedItem `json:"pointerConversionSymbolReference,omitempty"`
+	DefaultedAliasSymbolReference    DefaultedItem  `json:"defaultedAliasSymbolReference,omitempty"`
```
apelisse, line 27: So here we would end-up with what if you had the same comment as the line above. IOW, 
```suggestion
         // +default=ref(k8s.io/kube-openapi/test/integration/testdata/enumtype.FruitApple)
	DefaultedAliasSymbolReference    DefaultedItem  `json:"defaultedAliasSymbolReference,omitempty"`
```
would give "apple" or "SymbolConstant"? Because I find it hard to predict for people who will author these types to predict if it doesn't have the same behavior as above.

apelisse (commented):
That's totally awesome 🌈

test/integration/testdata/defaults/default.go
alexzielenski: What you wrote would give `apple`. The principle is simple and mirrors the same behavior in defaulter-gen: the default defined closest to your field is going to take precedence.

> Because I find it hard to predict for people who will author these types to predict if it doesn't have the same behavior as above.

I don't understand this

apelisse: If it gives apple, then we're fine. I thought there was a difference between pointer/non-pointer or something subtle hidden somewhere.

apelisse:
@thockin !!

PR #420, opened 2023-08-24
Support PostProcessSpec hook in V3

ryantxu:
Currently, the [PostProcessSpec](https://github.com/kubernetes/kube-openapi/blob/master/pkg/common/common.go#L115) hook is only called when creating v2, but does not work for V3.  This PR adds an additional hook and calls it while creating the spec.

While building an agregated api-server I am able to apply my special needs to the V2 spec, but not v3.  The call is initiated from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/routes/openapi.go#L68 so I think it makes most sense to hook in here, but also open to any other suggestions!


pkg/common/common.go
```diff
@@ -111,9 +111,12 @@ type Config struct {
 	// It is an optional function to customize model names.
 	GetDefinitionName func(name string) (string, spec.Extensions)
 
-	// PostProcessSpec runs after the spec is ready to serve. It allows a final modification to the spec before serving.
+	// PostProcessSpec runs after the V2 spec is ready to serve. It allows a final modification to the spec before serving.
 	PostProcessSpec func(*spec.Swagger) (*spec.Swagger, error)
 
+	// PostProcessSpec3 runs after the V3 spec is ready to serve. It allows a final modification to the spec before serving.
+	PostProcessSpec3 func(*spec3.OpenAPI) (*spec3.OpenAPI, error)
```
Jefftree, line 118: Would you be able to put this in OpenAPIV3Config below and use it from there?

ryantxu: that would make sense!  but....   but in apiserver, both the v2+v3 configs are `*openapicommon.Config`

https://github.com/kubernetes/apiserver/blob/master/pkg/server/config.go#L197
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/routes/openapi.go#L35

Jefftree: That's a good reason to separate out the configs. Thanks for bringing this up.

ryantxu:
@Jefftree -- thoughts on what to do with this PR?   I suspect changing [*openapicommon.Config
](https://github.com/kubernetes/apiserver/blob/master/pkg/server/config.go#L197C2-L197C39) to Config3 (or similar) is a massive project :)

Is the approach here a reasonable path?

Jefftree:
I briefly looked and it shouldn't be too hard to change to Config3. Will get to it soon

ryantxu:
Closing here -- this will not be required with https://github.com/kubernetes/kubernetes/pull/121010

Thanks @Jefftree

Jefftree:
@ryantxu The PR linked only decouples the OpenAPI V2 and V3 config. Parts of this PR would still be needed to support PostProcessSpec3 function right?

ryantxu:
Ya -- except it would be a hook on the v3 config.  I'll reopen and update.  I am just not clear how to easily test/verify until the other changes are sorted

PR #429, opened 2023-09-18
account for types with custom unmarshalling during default enfocement

alexzielenski:
Some types have surprising defaults such as `v1.Time` discussed in https://github.com/kubernetes/kubernetes/issues/118771#issuecomment-1599488644.

Had further discussion about a solution and we agreed in https://github.com/kubernetes/kubernetes/issues/109427#issuecomment-1723313763 that removing implicit `{}` defaults for struct fields with custom type is the best way to correct the schemas.

Original fix was included in https://github.com/kubernetes/kube-openapi/pull/419#discussion_r1312056111 but split off to keep things tidy.

/cc @apelisse

### DIFF
--- a/pkg/generators/openapi_test.go
+++ b/pkg/generators/openapi_test.go
@@ -1785,3 +1785,118 @@ type Blah struct {
 	}
 
 }
+
+// Show that types with unmarshalJSON in their hierarchy do not have struct
+// defaults enforced, and that aliases and embededd types are respected
+func TestMustEnforceDefaultStruct(t *testing.T) {
+	callErr, funcErr, assert, _, funcBuffer, imports := testOpenAPITypeWriter(t, `
+package foo
+
+type Time struct {
+	value interface{}
+}
+
+
+type TimeWithoutUnmarshal struct {
+	value interface{}
+}
+
+func (_ TimeWithoutUnmarshal) OpenAPISchemaType() []string { return []string{"string"} }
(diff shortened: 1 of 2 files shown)


apelisse:
Can we have a test for this?

pkg/generators/openapi_test.go
```diff
@@ -1785,3 +1785,118 @@ type Blah struct {
...
+						},
+					},
+					"StructUnmarshal": {
+						SchemaProps: spec.SchemaProps{
+							Ref: ref("base/foo.Time"),
+						},
+					},
+					"NoUnmarshal": {
+						SchemaProps: spec.SchemaProps{
+							Default: map[string]interface{}{},
```
alexzielenski, line 1880: @apelisse I don't think there are current upstream examples of this but if you override `OpenAPISchemaType` without unmarshal JSON you still get the struct default.

I'm thinking maybe should switch current code to check for OpenAPISchemaType instead of UnmarshalJSON, since we are trying to prevent inconsistent `type` + `default`. WDYT?

alexzielenski: Also curious why `OpenAPISchemaType` isnt just a marker comment

apelisse: Because we haven't had the need to do it, and we probably don't want people to annotate the type when we can guess it from the actual type?

apelisse:
We should put that on hold while we determine if we want to keep `mustEnforceDefault` or not.

PR #430, opened 2023-09-20
Do not import github.com/google/go-cmp/cmp outside tests.

dominiquelefevre:
Go-cmp relies heavily on reflection in ways that casue the golang compiler to disable the dead code elimination. Even declaring a variable of type cmp.Option disables the DCE.

Pkg/validation/spec has SwaggerDiffOptions that is only used in tests, but is declared in the package itself. This disables the DCE in every user of k8s.io/client-go. Move SwaggerDiffOptions to swagger_test.go where it belongs.


apelisse:
All of `spec/fuzz.go` is used for tests and should never be imported in an actual program, not just the diff options.
Should we move the whole file in a `spectest` package?

dominiquelefevre:
> All of `spec/fuzz.go` is used for tests and should never be imported in an actual program, not just the diff options.
> Should we move the whole file in a `spectest` package?

Oh, that's absolutely reasonable. Renamed the whole file to `fuzz_test.go` so that it is included into test builds only.

apelisse:
Are we 100% sure that none of this is used outside of this repo? Will this break kubernetes/kubernetes?

dominiquelefevre:
> Are we 100% sure that none of this is used outside of this repo? Will this break kubernetes/kubernetes?

I've grepped the k8s source for `SwaggerFuzzFuncs` and `SwaggerDiffOptions`. Kube-openapi's pkg/validation/spec is the only place where I find them.

Moreover, I've tested the patch with a project of my own that uses a bunch of k8s.io/* libs and even k8s.io/apiserver. It compiles just fine.

dominiquelefevre:
ping @apelisse

apelisse:
I don't see this being used anywhere, so it looks good to me.
/assign @liggitt

dominiquelefevre:
Hi, anybody out here? @sttts please have a look at the patch.

PR #431, opened 2023-09-27
Separate OpenAPI V2 and V3 config in V3 builder

Jefftree:
Decouple v2 and v3 configs to support https://github.com/kubernetes/kube-openapi/pull/420 properly


Jefftree:
/assign @apelisse 
Would appreciate your thoughts on this, thanks!

Jefftree:
k/k PR is a bit more involved: https://github.com/kubernetes/kubernetes/pull/121010

pkg/builder3/openapi.go
```diff
@@ -315,12 +315,12 @@ func (o *openAPI) buildOpenAPISpec(webServices []common.RouteContainer) error {
 // BuildOpenAPISpec builds OpenAPI v3 spec given a list of route containers and common.Config to customize it.
 //
 // Deprecated: BuildOpenAPISpecFromRoutes should be used instead.
-func BuildOpenAPISpec(webServices []*restful.WebService, config *common.Config) (*spec3.OpenAPI, error) {
+func BuildOpenAPISpec(webServices []*restful.WebService, config *common.OpenAPIV3Config) (*spec3.OpenAPI, error) {
```
apelisse, line 318: how breaking is this? We have to wonder how much we break now when we make these changes. I'm assuming that only the apiserver is using that code though.

Jefftree: cc @ryantxu since this will affect your workflow.

My understanding is that only apiserver uses this code, and the v3 path is less used than v2. It's also technically "wrong" to use the v2 config for v3 and I don't want to encourage people going down/sticking to that path. Things like `PostProcessSpec(*spec.Swagger) ...` (https://github.com/kubernetes/kube-openapi/blob/master/pkg/common/common.go#L114-L115) that takes a v2 spec is ignored when building v3.

I'm happy to create a new function if we want though.

ryantxu: FWIW, the standard codegen tools do not produce code that calls this directly.  The breaking change will be in https://github.com/kubernetes/kubernetes/pull/121010 😬 
image

But honestly it feels more like a bug fix since the v2 options do not cleanly apply to v3

pkg/builder3/openapi.go
```diff
@@ -174,9 +174,9 @@ func (o *openAPI) buildRequestBody(parameters []common.Parameter, consumes []str
 	return nil, nil
 }
 
-func newOpenAPI(config *common.Config) openAPI {
+func newOpenAPI(config *common.OpenAPIV3Config) openAPI {
 	o := openAPI{
-		config: common.ConvertConfigToV3(config),
```
apelisse, line 179: Should we consider removing this function?

apelisse:
Yeah, I think the breaking change is fine here. LGTM, just one questino.

pkg/builder3/openapi.go
Jefftree: Updated

PR #432, opened 2023-09-28
Add security requirement to root OpenAPI V3

Jefftree:
While looking into k8s authentication, noticed this field should be used but is missing from the OpenAPI V3. See https://spec.openapis.org/oas/v3.0.0#openapi-object and https://spec.openapis.org/oas/v3.0.0#securityRequirementObject

/assign @apelisse


apelisse:
Do we have corresponding fields for this in OpenAPI v2?

Jefftree:
Yes this field is already present and used in v2. `swagger.json` contains an extra 

```
  "security": [
    {
      "BearerToken": []
    }
  ],
```

that is not present in the v3 json. How many people rely on this -shrug-, but it's good to be consistent.

Jefftree:
@apelisse fixed the CI :)

PR #433, opened 2023-09-28
fix staticcheck - combine var declaration and assignment

Jefftree:
/assign @apelisse


apelisse:
the best change ever!
Do we not run staticchecks automatically in the CI?

Jefftree:
@apelisse Yeah this is quite a lot haha. We can't do much about the json-experiment-library

Jefftree:
Will rethink this PR since there are quite a lot of changes involved.

PR #435, opened 2023-10-16
Feature: Added support for OpenAPI validations marker comments

Schnides123:
This PR adds the ability to annotate API fields with basic OpenAPI schema validations via marker comments, as part of the [KEP-4153: Declarative Validation](https://github.com/kubernetes/enhancements/issues/4153) feature.

/assign @alexzielenski


pkg/generators/markers_test.go
```diff
@@ -0,0 +1,156 @@
...
+				"+maximum=20.0",
+				"+minLength=20",
+				"+maxLength=30",
+				`+pattern="asdf"`,
+				"+multipleOf=1.0",
+				"+minItems=1",
+				"+maxItems=2",
+				"+uniqueItems=true",
+				"not+Minimum=0.0",
+				// "+",
```
alexzielenski, line 51: Would like to see a test case showing what happens when `+` is used

pkg/generators/markers_test.go
```diff
@@ -0,0 +1,156 @@
...
+package generators_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/kube-openapi/pkg/generators"
+)
+
+func TestParseCommentTags(t *testing.T) {
```
alexzielenski, line 26: Unit tests are good. Probably would like to see more cases fleshed out. Also we want integration tests added to https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/openapi_test.go#L0-L1

pkg/generators/openapi.go
```diff
@@ -353,10 +353,49 @@ func (g openAPITypeWriter) generateCall(t *types.Type) error {
 	return g.Error()
 }
 
+func (g openAPITypeWriter) generateValueValidations(vs *ValueValidations) {
+
+	if vs == nil {
+		return
+	}
+	if vs.Minimum != nil {
+		g.Do("Minimum: util.PointerToFloat64($.Minimum$),\n", vs)
```
alexzielenski, line 362: `util` here may not be added correctly to imports. May have to support it to the generator args with `vs` and invoke the `|raw` filter on it which ensures to qualify the package correctly, and add it to imports.

alexzielenski, line 362: An alternative to using these helper functions is to do as you did in the test by abusing the array literal notation:

```
&[]float64{20.0}[0]
```

/cc @apelisse WDYT is preferable

pkg/generators/markers.go
```diff
@@ -0,0 +1,273 @@
...
+	UniqueItems      *bool    `json:"uniqueItems,omitempty"`
+	MultipleOf       *float64 `json:"multipleOf,omitempty"`
+	MaxProperties    *int64   `json:"maxProperties,omitempty"`
+	MinProperties    *int64   `json:"minProperties,omitempty"`
+}
+
+type CommentTags struct {
+	ValueValidations
+
+	// Future markers can all be parsed into this centralized struct...
```
alexzielenski, line 33: I like that a this struct is a good place to move our existing tags to over time rather than have an ever-expanding piecemeal solution

pkg/generators/openapi.go
```diff
@@ -353,10 +353,49 @@ func (g openAPITypeWriter) generateCall(t *types.Type) error {
...
+	if vs.UniqueItems != nil {
+		g.Do("UniqueItems: true,\n", nil)
+	}
+}
+
 func (g openAPITypeWriter) generate(t *types.Type) error {
 	// Only generate for struct type and ignore the rest
 	switch t.Kind {
 	case types.Struct:
+		overrides, err := ParseCommentTags(t.CommentLines)
```
alexzielenski, line 394: Should we also validate the provided tags? Are there other validations that might be reasonable to have of the marker comments?

- Max{X} > Min{X}
- Pattern requires string field, etc.
- Optional, Required, or Defaulted (future PR)

/cc @apelisse @thockin

pkg/generators/openapi.go
```diff
@@ -376,8 +415,9 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 				"SchemaProps: spec.SchemaProps{\n", args)
 			g.generateDescription(t.CommentLines)
 			g.Do("Type:$.type|raw${}.OpenAPISchemaType(),\n"+
-				"Format:$.type|raw${}.OpenAPISchemaFormat(),\n"+
-				"},\n"+
+				"Format:$.type|raw${}.OpenAPISchemaFormat(),\n", args)
+			g.generateValueValidations(&overrides.ValueValidations)
```
alexzielenski, line 419: Each case that adds a new `g.generateValueValidations` should have an associated unit test to show this case is being rendered properly.

existing nit: this code is kinda messy, a good future PR might clean this logic up a bit

alexzielenski (changes requested):
Really like the approach. needs a lot more testing, though. 

When we reach a more mature PR state we should open a k/k PR with these changes to show the diff with the generated OpenAPI

pkg/generators/markers.go
```diff
@@ -0,0 +1,217 @@
+package generators
```
alexzielenski, line 1: Missing boilerplate e.g. `Copyright 2023 The Kubernetes Authors....`. Copy the block from another file

pkg/generators/markers.go
```diff
@@ -0,0 +1,217 @@
...
+
+	var commentTags CommentTags
+	if err = json.Unmarshal(out, &commentTags); err != nil {
+		return CommentTags{}, fmt.Errorf("failed to unmarshal marker comments: %w", err)
+	}
+
+	return commentTags, nil
+}
+
+func parseMarkers(markerComments []string) (map[string]any, error) {
```
apelisse, line 81: Don't we already have something like this somewhere else? We already parse a lot of markers from comments, have you looked in `openapi.go`?

pkg/generators/markers_test.go
apelisse: Also we have some good "integration" tests that show that the final generate openapi is good given the actual comments and stuff. They give a lot of confidence.

pkg/generators/openapi.go
apelisse: Yeah, we could either verify these here or somewhere else, but to keep things simple I would just plain fail here.

I'll just say, in general the OpenAPI/json-schema is fairly loose with what one can do. so it's totally acceptable to have:
```
{
   maximum: 5
   maxLength: 10
}
```
which means, this can be of any type, but if it's a number it can't be bigger than 5 and if it's a string it can't be longer than 10. If it's an array, or map, it can be of arbitrary length. That being said, there are definitely things that still don't make sense like ...
```
{
   type: string
   maximum: 5
}
```
And we're a lot more likely to have these use-cases, so I would just fail these.

pkg/generators/markers.go
```diff
@@ -0,0 +1,217 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"k8s.io/gengo/types"
+)
+
+// Subset of spec.SchemaProps that may be placed into a marker comment
+type ValueValidations struct {
```
apelisse, line 14: Shouldn't we use this: https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/spec/schema.go#L161-L197?

pkg/generators/markers.go
alexzielenski: We use the helpers defined here currently: https://github.com/kubernetes/gengo/blob/master/types/comments.go

But they do not serve this use case well. In KEP-4153 we discussed some CEL ergonomics. So specifically these new marker parsers allow us to interpret the following as you owuld expect:

```
// +k8s:validation:cel[0]:rule="self.pylons >= oldSelf.pylons"
// +k8s:validation:cel[0]:message="YOU MUST CONSTRUCT ADDITIONAL PYLONS"
// +k8s:validation:cel[0]:fieldPath=".pylons"
// +k8s:validation:cel[0]:reason="Invalid"
// +k8s:validation:cel[1]:rule="self.base.owner == 'us'"
// +k8s:validation:cel[1]:message="ALL YOUR BASE ARE BELONG TO US"
// +k8s:validation:cel[1]:fieldPath=".base"
// +k8s:validation:cel[1]:reason="Forbidden"
```

With the existing parser, this would have to be expressed on a single line.

pkg/generators/markers.go
alexzielenski: Yeah probably can. I think this was used because original implementation meant to serialize the fields and we wanted to exclude missing fields. But I don't think that's a problem anymore. 

@Schnides123 remove ValueValidations and use schemaprops again?

apelisse:
I would still add an intergration test. Create a new directory for value validation in https://github.com/kubernetes/kube-openapi/tree/master/test/integration/testdata, add some types in there with various validation markers, carefully follow the [README](https://github.com/kubernetes/kube-openapi/blob/master/test/integration/README.md) including the notes at the bottom. Thanks!

pkg/generators/markers.go
```diff
@@ -0,0 +1,242 @@
...
+	Pattern          *string  `json:"pattern,omitempty"`
+	MaxItems         *int64   `json:"maxItems,omitempty"`
+	MinItems         *int64   `json:"minItems,omitempty"`
+	UniqueItems      *bool    `json:"uniqueItems,omitempty"`
+	MultipleOf       *float64 `json:"multipleOf,omitempty"`
+	MaxProperties    *int64   `json:"maxProperties,omitempty"`
+	MinProperties    *int64   `json:"minProperties,omitempty"`
+}
+
+type CommentTags struct {
```
Jefftree, line 48: What's our plan with existing supported comments like +default and +nullable? Is the plan for all tags to adhere to the +k8s:validation:<field> convention?

test/integration/testdata/valuevalidation/alpha.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"k8s.io/kube-openapi/pkg/common"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// Dummy type to test the openapi-gen API rule checker.
+// The API rule violations are in format of:
+// -> +k8s:validation:[validation rule]=[value]
+
+// +k8s:validation:maxProperties=5
```
apelisse, line 12: Wait, when did we decide to use `k8s:validation:` prefix? That's not useful. it should be `+maxProperties`.

pkg/generators/markers.go
apelisse: Thanks for raising that point, I don't think we want it AT ALL.

Schnides123: +default and +nullable won't be modified in any way. The prefixes are only being applied to the comment types added in the current PR

pkg/generators/markers_test.go
Schnides123: added a test for +

test/integration/testdata/valuevalidation/alpha.go
Schnides123: removed the prefix, added a whitelist to prevent collision with other marker comments

pkg/generators/markers.go
Schnides123: prefixes have been removed

Schnides123:
I have been told to add the prefix back in, so the prefix is now back in

pkg/generators/markers.go
```diff
@@ -0,0 +1,242 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+
+	"k8s.io/gengo/types"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// Subset of spec.SchemaProps that may be placed into a marker comment
+type ValueValidations struct {
```
alexzielenski, line 32: Remove unused struct

pkg/generators/markers.go
```diff
@@ -0,0 +1,242 @@
...
+	return commentTags, nil
+}
+
+func parseMarkers(markerComments []string) (map[string]any, error) {
+	markers := types.ExtractCommentTags("+", markerComments)
+
+	// Parse the values as JSON
+	result := map[string]any{}
+	for key, value := range markers {
+		if !strings.HasPrefix(key, "k8s:validation:") {
```
alexzielenski, line 105: Change to argument + constant

pkg/generators/markers_test.go
```diff
@@ -0,0 +1,140 @@
...
+				"+k8s:validation:multipleOf=1.0",
+				"+k8s:validation:minItems=1",
+				"+k8s:validation:maxItems=2",
+				"+k8s:validation:uniqueItems=true",
+				"exclusiveMaximum=true",
+				"not+k8s:validation:Minimum=0.0",
+			},
+			expected: generators.CommentTags{
+				spec.SchemaProps{
+					Maximum:     &[]float64{20.0}[0],
```
alexzielenski, line 56: use `k8s.io/utils/ptr.To`

pkg/generators/openapi.go
```diff
@@ -353,10 +354,65 @@ func (g openAPITypeWriter) generateCall(t *types.Type) error {
...
+	}
+	if len(vs.Pattern) > 0 {
+		p, err := json.Marshal(vs.Pattern)
+		if err != nil {
+			panic(err)
+		}
+		g.Do("Pattern: $.$,\n", string(p))
+	}
+	if vs.MultipleOf != nil {
+		g.Do("MultipleOf: &[]float64{$.MultipleOf$}[0],\n", vs)
```
alexzielenski, line 394: use `ptr.To(LITERAL)`

alexzielenski (changes requested):

test/integration/testdata/valuevalidation/alpha.go
alexzielenski, line 12: need to add k8s:open-api gen tags

and enable test integration_test_suite.go, and add route for it in testuti/testutil.go

This process should be made far simpler in a future PR

pkg/generators/markers.go
```diff
@@ -0,0 +1,225 @@
...
+	//    "foo": map[string]any{
+	//      "bar": map[string]any{
+	//        "baz": map[string]any{
+	//          "qux": "quux",
+	//        },
+	//      },
+	//    },
+	//  }
+	//
+	markers, err := parseMarkers(comments, markerPrefix)
```
alexzielenski, line 58: Could we also plumb the prefix out as an argument into `ParseCommentTags`?

pkg/generators/openapi.go
```diff
@@ -353,10 +355,74 @@ func (g openAPITypeWriter) generateCall(t *types.Type) error {
...
+	if vs.MinProperties != nil {
+		g.Do("MinProperties: $.ptrTo|raw$[int64]($.spec.MinProperties$),\n", args)
+	}
+	if vs.MaxProperties != nil {
+		g.Do("MaxProperties: $.ptrTo|raw$[int64]($.spec.MaxProperties$),\n", args)
+	}
+	if len(vs.Pattern) > 0 {
+		p, err := json.Marshal(vs.Pattern)
+		if err != nil {
+			panic(err)
```
alexzielenski, line 399: Replace panic with error return and thread through to callsities

pkg/generators/openapi.go
```diff
@@ -376,8 +442,9 @@ func (g openAPITypeWriter) generate(t *types.Type) error {
 				"SchemaProps: spec.SchemaProps{\n", args)
 			g.generateDescription(t.CommentLines)
 			g.Do("Type:$.type|raw${}.OpenAPISchemaType(),\n"+
-				"Format:$.type|raw${}.OpenAPISchemaFormat(),\n"+
-				"},\n"+
+				"Format:$.type|raw${}.OpenAPISchemaFormat(),\n", args)
+			g.generateValueValidations(&overrides.SchemaProps)
```
alexzielenski, line 446: I see the following new calls to `generateValueValidations`

- [x] Custom V2 Type/Format and V3 Definition Functions
- [x] Custom V2 Type/Format and V3 oneOf types
- [ ] Custom V2 Type/Format (and No V3 Definition or OneOf)
- [ ] V3 OneOf Types
- [x] Regular Behavior (no custom types or override functions implemented)

The checked boxes are those that an integration exists already. This is function is pretty complicated. I wonder if we can simplify this function in a future PR. In the meantime, I'd like to see the rest of the cases covered in the integration tests to make sure the output compiles in those cases.

alexzielenski (changes requested):
Looking really good. Few more tests before I think it is ready to go

pkg/generators/markers.go
```diff
@@ -0,0 +1,225 @@
...
+	for key, value := range markers {
+		if !strings.HasPrefix(key, prefix) {
+			// we only care about validation markers for now
+			continue
+		}
+
+		newKey := strings.TrimPrefix(key, prefix)
+
+		// Skip ref markers
+		if len(value) == 1 && regexp.MustCompile(`^(?:\s*ref\s{0,1}\().*\)\s*`).MatchString(value[0]) {
```
alexzielenski, line 79: Is this possible try `"k8s.io/gengo/examples/defaulter-gen/generators".ParseSymbolReference` against the value to see if it is a ref instead? To avoid duplciating this logic

pkg/generators/markers.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, errors.Join(errs...)
+	}
+
+	return nested, nil
+}
+
+// Recursively puts a value into the given keypath, creating intermediate maps
+// and slices as needed. If a key is of the form `foo[bar]`, then bar will be
+// treated as an index into the array foo. If bar is not a valid integer, then
+// the key will be ignored.
```
alexzielenski, line 151: Could we make it return an error instead of ignoring non-int array subscripts?

pkg/generators/markers.go
```diff
@@ -0,0 +1,225 @@
...
+
+	var commentTags CommentTags
+	if err = json.Unmarshal(out, &commentTags); err != nil {
+		return CommentTags{}, fmt.Errorf("failed to unmarshal marker comments: %w", err)
+	}
+
+	return commentTags, nil
+}
+
+func parseMarkers(markerComments []string, prefix string) (map[string]any, error) {
```
alexzielenski, line 65: add simple godoc comment for function

alexzielenski (changes requested):

pkg/generators/markers.go
```diff
@@ -36,7 +36,87 @@ type CommentTags struct {
...
+		}
+	case types.Slice, types.Array:
+		if c.Minimum != nil || c.Maximum != nil || c.MultipleOf != nil || c.MinProperties != nil || c.MaxProperties != nil || c.MinLength != nil || c.MaxLength != nil || c.Pattern != "" {
+			return fmt.Errorf("invalid validation for %s type", t.Name)
+		}
+	case types.Map:
+		if c.Minimum != nil || c.Maximum != nil || c.MultipleOf != nil || c.MinItems != nil || c.MaxItems != nil || c.MinLength != nil || c.MaxLength != nil || c.Pattern != "" {
+			return fmt.Errorf("invalid validation for %s type", t.Name)
+		}
+	}
```
alexzielenski, line 71: For the enforcement of which validation can go on which type i think youd be better served by not repeating conditions, and also more specific error messages:

```go
reolvedType := resolveAliasAndPtrType(&t)
typeString, _ := openapi.OpenAPITypeFormat(reolvedType.String()) // will be empty for complicated types
isArray := resolvedType.Kind == types.Slice || resolvedType.Kind == types.Array
isMap := resolveType.Kind == types.Map
isString := typeString == "string"
isInt := typeString == "integer"
isFloat := typeString == "number"
isBool := typeString == "boolean"

if c.MaxItems != nil && !isArray {
    allErrors = append(allErrors, fmt.Errorf("maxItems may only be specified on arrays"))
}

...
```

alexzielenski (changes requested):

pkg/generators/markers.go
Schnides123: I think the logic does already return an error, the comment just needs to be updated.

pkg/generators/markers.go
```diff
@@ -59,6 +139,11 @@ func ParseCommentTags(comments []string, prefix string) (CommentTags, error) {
 		return CommentTags{}, fmt.Errorf("failed to unmarshal marker comments: %w", err)
 	}
 
+	// Validate the parsed comment tags
+	if err := commentTags.Validate(*t); err != nil {
```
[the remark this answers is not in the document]
Schnides123: Are you sure? I feel like there really isn't a use-case for generating comment tags that wouldn't pass validation. Splitting it out might be convenient for plumbing the type out of the ParseCommentTags arguments, but I don't foresee us having any situations where we'd want to create a commentTags instance and then not validate it immediately after.

pkg/generators/openapi.go
alexzielenski: resolved

alexzielenski (approved):
/approve

Looks really good! Great starting point for the first alpha. Thanks for the extensive testing! 

I pinned k/k against this branch and tested there are no dep changes (other than kube-openapi) and that running the code generator (without using the markers) is a no-op.

alexzielenski:
/hold

can you squash

pkg/generators/markers.go
alexzielenski: resolved

alexzielenski (approved):

PR #436, opened 2023-10-25
Optimize CRD building by reusing Definitions

Jefftree:


alexzielenski:
@Jefftree is this ready for review? curious if we still want this in

Jefftree:
Will add some testing and report back

PR #442, opened 2023-12-29
Fix integration test route builder for go-restful > 3.8

kapouer:
It would produce empty paths without leading slash


Jefftree:
Thanks!
/lgtm
/approve

PR #443, opened 2024-01-03
bump go to 1.21

Jefftree:
/assign @alexzielenski 

Since k/k is already on 1.21.

Integration `go.sum` changes were from running a `go mod tidy` in that directory.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module k8s.io/kube-openapi
 
-go 1.19
+go 1.21
 
 require (
 	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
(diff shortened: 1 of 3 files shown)


alexzielenski:
It looks like k8s 1.26 is still in support and [uses Go 1.19](https://github.com/kubernetes/kubernetes/blob/v1.26.12/go.mod#L9).

Is it possible that in the future we would need to bump the k8s dep on kube-openapi for a bugfix? If thats the case I think making this change would require the older versions of k8s use go1.21 also

Jefftree:
That's a good point, although we'd most likely cut a new branch and add a specific commit off of the version of the kube-openapi dependency from the k8s version? Fast forwarding to the latest version is usually undesired since we don't provide backwards compatibility guarantees and could include unwanted features when we're targeting bug fixes. I haven't seen any cases of a bug-fix backport but if you feel that's an important use case I'm okay with delaying this PR.

alexzielenski:
Yes I see now we have a separate branch for releases that needed backports. Didn't realize that was our process for this repo.

/lgtm

PR #444, opened 2024-01-05
removed error-join shim after go version update

Schnides123:


alexzielenski:
Thanks for cleaning this up!

/lgtm
/approve

PR #445, opened 2024-01-06
Adds support for CEL marker comments

alexzielenski:
Adds CEL marker comments support to kube-openapi. Previous extension support only allowed strings. This PR makes use of new marker comment system introduced in #435 to emit CEL vendor extensions.

Multiline syntax is not yet supported. Proposed syntax is 1:1 with CEL CRD fields:

```go
	// +k8s:validation:cel[0]:rule="self.length() > 0"
	// +k8s:validation:cel[0]:message="string message"
	// +k8s:validation:cel[1]:rule="self.length() % 2 == 0"
	// +k8s:validation:cel[1]:messageExpression="self + ' hello'"
	// +optional
	CELField string `json:"celField"`
```

There are linter errors to help keep the indexing straight

/assign @jefftree
/cc @Schnides123

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -295,8 +295,9 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
 	rest := k[1:]
 
 	if idxIdx := strings.Index(key, "["); idxIdx > -1 {
+		subscript := strings.Split(key[idxIdx+1:], "]")[0]
 		key := key[:idxIdx]
-		index, err := strconv.Atoi(strings.Split(key[idxIdx+1:], "]")[0])
+		index, err := strconv.Atoi(subscript)
 		if err != nil {
 			// Ignore key
 			return nil, fmt.Errorf("expected integer index in key %v, got %v", key, key[idxIdx+1:])
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -306,9 +306,15 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
 		var arrayDestination []any
 		if existing, ok := m[key]; !ok {
 			arrayDestination = make([]any, index+1)
-		} else {
+		} else if existing, ok := existing.([]any); !ok {
+			// Error case. Existing isn't of correct type. Can happen if
+			// someone is subscripting a field that was previously not an array
+			return nil, fmt.Errorf("expected []any at key %v, got %T", key, existing)
+		} else if index >= len(existing) {
 			// Ensure array is big enough
-			arrayDestination = append(existing.([]any), make([]any, index-len(existing.([]any))+1)...)
+			arrayDestination = append(existing, make([]any, index-len(existing)+1)...)
+		} else {
+			arrayDestination = existing
 		}
 
 		m[key] = arrayDestination
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -30,9 +30,41 @@ import (
 	"k8s.io/kube-openapi/pkg/validation/spec"
 )
 
+type CELTag struct {
+	Rule              string `json:"rule,omitempty"`
+	Message           string `json:"message,omitempty"`
+	MessageExpression string `json:"messageExpression,omitempty"`
+	OptionalOldSelf   *bool  `json:"optionalOldSelf,omitempty"`
+}
+
+func (c *CELTag) Validate() error {
+	if c == nil || *c == (CELTag{}) {
+		return fmt.Errorf("empty CEL tag is not allowed")
+	}
+
+	var errs []error
+	if c.Rule == "" {
+		errs = append(errs, fmt.Errorf("rule cannot be empty"))
+	}
+	if c.Message == "" && c.MessageExpression == "" {
+		errs = append(errs, fmt.Errorf("message or messageExpression must be set"))
+	}
+	if c.Message != "" && c.MessageExpression != "" {
+		errs = append(errs, fmt.Errorf("message and messageExpression cannot be set at the same time"))
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	return nil
+}
+
 // CommentTags represents the parsed comment tags for a given type. These types are then used to generate schema validations.
 type CommentTags struct {
 	spec.SchemaProps
+
+	CEL []CELTag `json:"cel,omitempty"`
 
 	// Future markers can all be parsed into this centralized struct...
 	// Optional bool `json:"optional,omitempty"`
@@ -85,6 +117,14 @@ func (c CommentTags) Validate() error {
 	}
 	if c.MultipleOf != nil && *c.MultipleOf == 0 {
 		err = errors.Join(err, fmt.Errorf("multipleOf cannot be 0"))
+	}
+
+	for i, celTag := range c.CEL {
+		celError := celTag.Validate()
+		if celError == nil {
+			continue
+		}
+		err = errors.Join(err, fmt.Errorf("invalid CEL tag at index %d: %w", i, celError))
 	}
 
 	return err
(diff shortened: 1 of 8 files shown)


pkg/generators/markers.go
```diff
@@ -87,6 +113,10 @@ func (c CommentTags) Validate() error {
 		err = errors.Join(err, fmt.Errorf("multipleOf cannot be 0"))
 	}
 
+	for _, celTag := range c.CEL {
+		err = errors.Join(err, celTag.Validate())
```
Jefftree, line 117: I think it would be also useful to indicate the index of the expression that failed. If a user accidentally skips an index, it's quite difficult to pinpoint.

pkg/generators/markers.go
```diff
@@ -295,8 +325,9 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
 	rest := k[1:]
 
 	if idxIdx := strings.Index(key, "["); idxIdx > -1 {
+		subscript := strings.Split(key[idxIdx+1:], "]")[0]
```
Jefftree, line 328: Just want to double check, at the moment the array is only used for CEL right? Do we have other cases of array usage?

pkg/generators/markers.go
```diff
@@ -305,20 +336,25 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
...
-			arrayDestination[index] = destination
-			return putNestedValue(destination, rest, v)
+			arrayDestination[index], err = putNestedValue(destination, rest, v)
 		} else if dst, ok := arrayDestination[index].(map[string]any); ok {
 			// Already exists case, correct type
-			return putNestedValue(dst, rest, v)
+			arrayDestination[index], err = putNestedValue(dst, rest, v)
 		}
 
 		// Already exists, incorrect type. Error
```
Jefftree, line 360: This comment is inaccurate now that we're not returning on line 354 and 357? I'm also a bit confused why we're ignoring the error returned from those lines now?

pkg/generators/markers_test.go
```diff
@@ -131,20 +131,102 @@ func TestParseCommentTags(t *testing.T) {
...
+				`+k8s:validation:cel[0]:rule="oldSelf == self"`,
+				`+k8s:validation:cel[0]:message="immutable field"`,
+				`+k8s:validation:cel[2]:rule="self > 5"`,
+				`+k8s:validation:cel[2]:message="must be greater than 5"`,
+			},
+			expectedError: `invalid marker comments: empty CEL tag is not allowed`,
+		},
+		{
+			t: types.Float64,
+			// This should parse, but return an error in validation since
```
Jefftree, line 178: Is this comment referring to the previous `skipped CEL rule`?

pkg/generators/markers_test.go
```diff
@@ -131,20 +131,102 @@ func TestParseCommentTags(t *testing.T) {
...
+						Rule:            "self > 5",
+						Message:         "must be greater than 5",
+						OptionalOldSelf: ptr.To(true),
+					},
+				},
+			},
+		},
+		{
+			t: types.Float64,
+			// This should parse, but return an error in validation since
```
Jefftree, line 204: same question as above

pkg/generators/markers_test.go
```diff
@@ -167,315 +249,254 @@ func TestParseCommentTags(t *testing.T) {
...
+		comments     []string
+		t            *types.Type
+		errorMessage string
 	}{
 		{
 			name: "invalid minimum type",
-			commentParams: map[string]any{
-				"minimum": 10.5,
+			comments: []string{
+				`+k8s:validation:minimum=10.5`,
```
Jefftree, line 260: I think I missed the discussion on adding this prefix, is it going to cause any backwards incompatibilities with the rename?

pkg/generators/markers_test.go
```diff
@@ -131,20 +131,102 @@ func TestParseCommentTags(t *testing.T) {
...
 			name: "invalid: invalid value",
 			comments: []string{
 				"+k8s:validation:",
 			},
 			expectedError: `failed to parse marker comments: cannot have empty key for marker comment`,
 		},
 		{
-			t:    &numKind,
+			t: types.Float64,
+			// temporary test. ref support may be added in the future
```
Jefftree, line 143: noop? Might be good to add something else to the test case to ensure we have an output and the ref is ignored.

Jefftree (commented):
Mostly looks good. Still familiarizing myself with parts of CEL so lmk if any comments don't make sense.

pkg/generators/markers.go
alexzielenski: correct this isnt used for anything else

pkg/generators/markers_test.go
alexzielenski: This isn't a rename, I changed the way the test works so that it expects the comments before parsing rather than after parsing

pkg/generators/markers_test.go
alexzielenski: not a no-op, just the expected output is the empty `CommentTags`, so we are checking that the ref is ignored

pkg/generators/markers.go
```diff
@@ -30,10 +30,36 @@ import (
...
 )
 
+type CELTag struct {
+	Rule              string `json:"rule,omitempty"`
+	Message           string `json:"message,omitempty"`
+	MessageExpression string `json:"messageExpression,omitempty"`
+	OptionalOldSelf   *bool  `json:"optionalOldSelf,omitempty"`
+}
+
+func (c *CELTag) Validate() error {
```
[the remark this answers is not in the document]
alexzielenski: added

pkg/generators/markers_test.go
alexzielenski: Yeah, moved it

pkg/generators/markers_test.go
alexzielenski: removed

pkg/generators/markers.go
alexzielenski: nice catch. I was throwing out the error for the recursie call. Ive straightened out the error handling a bit.

pkg/generators/markers.go
```diff
@@ -305,25 +346,37 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
...
 			arrayDestination = make([]any, index+1)
-		} else {
+		} else if existing, ok := existing.([]any); !ok {
+			// Error case. Existing isn't of correct type. Can happen if
+			// someone is subscripting a field that was previously not an array
+			return nil, fmt.Errorf("expected []any at key %v, got %T", key, existing)
+		} else if index >= len(existing) {
 			// Ensure array is big enough
-			arrayDestination = append(existing.([]any), make([]any, index-len(existing.([]any))+1)...)
+			arrayDestination = append(existing, make([]any, index-len(existing)+1)...)
```
Jefftree, line 355: This is quite neat, though realistically we'd probably expect users to populate the array in order since we'd error out on empty indices. No changes requested, just some food for thought.

pkg/generators/markers.go
```diff
@@ -305,25 +346,37 @@ func putNestedValue(m map[string]any, k []string, v any) (map[string]any, error)
 		var arrayDestination []any
 		if existing, ok := m[key]; !ok {
 			arrayDestination = make([]any, index+1)
-		} else {
+		} else if existing, ok := existing.([]any); !ok {
+			// Error case. Existing isn't of correct type. Can happen if
```
Jefftree, line 350: I'm trying to understand this case better. Can you describe how it could be previously not an array? I'm just picturing assigning an item as both an array and not an array. Eg:

```
// +k8s:validation:cel="string message"
// +k8s:validation:cel[0]:rule="self.length() > 0"
```

pkg/generators/markers.go
Jefftree: awesome, thanks!

test/integration/testdata/golden.v2.json
```diff
@@ -1265,20 +1265,52 @@
...
       "minLength": 1,
       "pattern": "^a.*b$"
+     },
+     "celField": {
+      "type": "string",
+      "default": "",
+      "x-kubernetes-validations": [
+       {
+        "message": "string message",
+        "rule": "self.length() \u003e 0"
```
Schnides123, line 1275: Do we want characters encoded like this?

pkg/generators/markers.go
alexzielenski: thats basically how

test/integration/testdata/golden.v2.json
alexzielenski: nice catch. definitely not

alexzielenski: According to [the docs](https://pkg.go.dev/encoding/json#Marshal) this is expected behavior:

> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. So that the JSON will be safe to embed inside HTML <script> tags, the string is encoded using HTMLEscape, which replaces "<", ">", "&", U+2028, and U+2029 are escaped to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029". This replacement can be disabled when using an Encoder, by calling SetEscapeHTML(false).

Example:

```go
func TestMarshalString(t *testing.T) {
	myString := "2 > 1"
	myJSON, err := json.Marshal(myString)
	if err != nil {
		t.Fatal(err)
	}
	if `"2 \u003e 1"` != string(myJSON) {
		t.Fatalf("expected %q, got %q", `"2 \u003e 1"`, string(myJSON))
	}
}
```


Given that it is valid JSON and parses the same way, I think it's fine.

pkg/generators/markers_test.go
```diff
@@ -112,7 +112,7 @@ func TestParseCommentTags(t *testing.T) {
 				"+k8s:validation:maximum=20.0",
 				"+k8s:validation:minimum=30.0",
 			},
-			expectedError: `cannot unmarshal array into Go struct field CommentTags.minimum of type float64`,
+			expectedError: `failed to parse marker comments: cannot have multiple values for key 'minimum'`,
```
Jefftree: Nice, outputting a more precise error message is great

Jefftree: question: Is this still part of the PR?

alexzielenski: yes

pkg/generators/markers.go
```diff
@@ -240,15 +240,21 @@ func lintArrayIndices(markerComments []string) error {
 	prevIndex := -1
 	for _, line := range markerComments {
 		line = strings.Trim(line, " ")
-		subscriptIdx := strings.Index(line, "[")
+		split := strings.Split(line, "=")
+		if len(split) < 2 {
+			// Ignore malformed key in linter.
```
Jefftree, line 245: no error?

Jefftree: actually, just realized this is possible with bool, so in a sense it's not malformed?

The error would still get caught later on when validating empty CEL tags, not sure if you'd like to error earlier though?

Test case:

```
				`+k8s:validation:cel[0]:rule="oldSelf == self"`,
				`+k8s:validation:cel[6]:optionalOldSelf`,
```

alexzielenski: > The error would still get caught later on when validating empty CEL tags, not sure if you'd like to error earlier though?

The linter is not meant to catch those types of errors. Just avoid common mistakes labelling array indices

> actually, just realized this is possible with bool, so in a sense it's not malformed?

That's a good catch. Ill add a test and fix that case

alexzielenski: fixed & added tests

alexzielenski:
squashed, @Jefftree can you re-lgtm?

PR #446, opened 2024-01-17
add raw string/multiline syntax to marker comments

alexzielenski:
adds ability to specify a raw/multiline string in kube-openapi marker comments to make longer strings more ergonomic to use

Syntax Example:

```
+k8s:validation:key> value line 1
+k8s:validation:key> value line 2
```

yields `key=" validation line 2\n validation line 2"`

Using `=` to specify multiple key-value pairs is now an error for `k8s:validation` prefixed items. Lists must be specified using explicit indices

/assign @Jefftree 
/cc @Schnides123

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -231,13 +231,24 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
 	return commentTags, nil
 }
 
-// ExtractCommentTags parses comments for lines of the form:
+var (
+	allowedKeyCharacterSet = `[:_a-zA-Z0-9\[\]\-]`
+	valueEmpty             = regexp.MustCompile(fmt.Sprintf(`^(%s*)$`, allowedKeyCharacterSet))
+	valueAssign            = regexp.MustCompile(fmt.Sprintf(`^(%s*)=(.*)$`, allowedKeyCharacterSet))
+	valueRawString         = regexp.MustCompile(fmt.Sprintf(`^(%s*)>(.*)$`, allowedKeyCharacterSet))
+)
+
+// extractCommentTags parses comments for lines of the form:
 //
-//	'marker' + "key=value".
+//	'marker' + "key=value"
+//
+//	or to specify truthy boolean keys:
+//
+//	'marker' + "key"
 //
 // Values are optional; "" is the default.  A tag can be specified more than
-// one time and all values are returned.  If the resulting map has an entry for
-// a key, the value (a slice) is guaranteed to have at least 1 element.
+// one time and all values are returned.  Returns a map with an entry for
+// for each key and a value.
 //
 // Similar to version from gengo, but this version support only allows one
 // value per key (preferring explicit array indices), supports raw strings
@@ -255,15 +266,11 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
 // assigned key
 // (in contrast to types.ExtractCommentTags which allows array-typed
 // values to be specified using `=`).
-var (
-	allowedKeyCharacterSet = `[:_a-zA-Z0-9\[\]\-]`
-	valueEmpty             = regexp.MustCompile(fmt.Sprintf(`^(%s*)$`, allowedKeyCharacterSet))
-	valueAssign            = regexp.MustCompile(fmt.Sprintf(`^(%s*)=(.*)$`, allowedKeyCharacterSet))
-	valueRawString         = regexp.MustCompile(fmt.Sprintf(`^(%s*)>(.*)$`, allowedKeyCharacterSet))
-)
-
 func extractCommentTags(marker string, lines []string) (map[string]string, error) {
 	out := map[string]string{}
+
+	// Used to track the the line immediately prior to the one being iterated.
+	// If there was an invalid or ignored line, these values get reset.
 	lastKey := ""
 	lastIndex := -1
 	lastArrayKey := ""
@@ -273,12 +280,15 @@ func extractCommentTags(marker string, lines []string) (map[string]string, error
 	for _, line := range lines {
 		line = strings.Trim(line, " ")
 
-		// Make sure last key gets reset if we `continue`
+		// Track the current value of the last vars to use in this loop iteration
+		// before they are reset for the next iteration.
 		previousKey := lastKey
 		previousArrayKey := lastArrayKey
 		previousIndex := lastIndex
 
+		// Make sure last vars gets reset if we `continue`
 		lastIndex = -1
+		lastArrayKey = ""
 		lastKey = ""
 
 		if len(line) == 0 {
@@ -362,6 +372,9 @@ func extractCommentTags(marker string, lines []string) (map[string]string, error
 		out[key] = value
 		lastKey = key
 
+		// Lint the array subscript for common mistakes. This only lints the last
+		// array index used, (since we do not have a need for nested arrays yet
+		// in markers)
 		if arrayPath, index, hasSubscript, err := extractArraySubscript(key); hasSubscript {
 			// If index is non-zero, check that that previous line was for the same
 			// key and either the same or previous index
@@ -530,6 +543,14 @@ func putNestedValue(m map[string]any, k []string, v any) error {
 	}
 }
 
+// extractArraySubscript extracts the left array subscript from a key of
+// the form  `foo[bar][baz]` -> "bar".
+// Returns the key without the subscript, the index, and a bool indicating if
+// the key had a subscript.
+// If the key has a subscript, but the subscript is not a valid integer, returns an error.
+//
+// This can be adapted to support multidimensional subscripts probably fairly
+// easily by retuning a list of ints
 func extractArraySubscript(str string) (string, int, bool, error) {
 	subscriptIdx := strings.Index(str, "[")
 	if subscriptIdx == -1 {
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -472,8 +472,8 @@ func nestMarkers(markers map[string]any) (map[string]any, error) {
 	for key, value := range markers {
 		var err error
 		keys := strings.Split(key, ":")
-		nested, err = putNestedValue(nested, keys, value)
-		if err != nil {
+
+		if err = putNestedValue(nested, keys, value); err != nil {
 			errs = append(errs, err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/generators/markers_test.go
```diff
@@ -325,6 +332,58 @@ func TestParseCommentTags(t *testing.T) {
 				},
 			},
 		},
+		{
+			t:    types.Float64,
+			name: "raw string rule",
+			comments: []string{
+				`+k8s:validation:cel[0]:rule> raw string rule`,
```
Jefftree, line 339: question: has this design with `>` representing multiline been flushed out somewhere or is this still experimental?

pkg/generators/markers.go
```diff
@@ -293,57 +296,150 @@ func lintArrayIndices(markerComments []string, prefix string) error {
 	return nil
 }
 
+// ExtractCommentTags parses comments for lines of the form:
+//
+//	'marker' + "key=value".
```
Jefftree, line 301: can you give an example in the docs of input and output? We're changing the return type compared to `types.ExtractCommentTags` and it would be great to have that documented.

pkg/generators/markers.go
```diff
@@ -293,57 +296,150 @@ func lintArrayIndices(markerComments []string, prefix string) error {
...
+				valueBytes, err := json.Marshal(matches[2])
+				if err != nil {
+					return nil, fmt.Errorf("invalid value for key %v: %w", key, err)
+				}
+
+				value = string(valueBytes)
+			} else if key != previousKey {
+				// Successive usages of the same key of a raw string must be
+				// consecutive
+				return nil, fmt.Errorf("concatenations to key '%s' must be consecutive with its assignment", key)
```
Jefftree, line 374: love the detailed error messages (Y). Non blocking but would be nice to see test cases with these errors

jpbetz:
I'm okay with this as an approach. I haven't reviewed the implementation in detail.  Since this is not API facing, I think we just need to get agreement on the go tag format internally.

@thockin @deads2k  do you have any thoughts on this format?  It seems to me that it expands our go tag support to handle the cases we need to handle, and avoids the "super long oneline gotags" that kubebuilder suffers from

pkg/generators/markers.go
```diff
@@ -293,57 +296,150 @@ func lintArrayIndices(markerComments []string, prefix string) error {
...
+//
+// Similar to version from gengo, but this version support only allows one
+// value per key (preferring explicit array indices), supports raw strings
+// with concatenation, and limits the usable characters allowed in a key
+// (for simpler parsing).
+func extractCommentTags(marker string, lines []string) (map[string]string, error) {
+	allowedKeyCharacterSet := `[:_a-zA-Z0-9\[\]\-]`
+	valueEmpty := regexp.MustCompile(fmt.Sprintf(`^\+%s(%s*)$`, marker, allowedKeyCharacterSet))
+	valueAssign := regexp.MustCompile(fmt.Sprintf(`^\+%s(%s*)=(.*)$`, marker, allowedKeyCharacterSet))
+	valueRawString := regexp.MustCompile(fmt.Sprintf(`^\+%s(%s*)>(.*)$`, marker, allowedKeyCharacterSet))
```
jpbetz, line 315: The marker can change but we only have a few of them in practice.. but we've got to compile regex on each `extractCommentTags` invocation.  That's unfortunate.  I see we're doing a `strings.HasPrefix(line, marker) ` of each line anyway as a pre-filter.  Could we remove the prefix and then use a regex pattern that doesn't need to contain the marker, allowing us to only compile the regexes once (maybe in a sync.Once to avoid doing it in an init and holding the memory if we don't need it?)

pkg/generators/markers_test.go
```diff
@@ -123,16 +122,24 @@ func TestParseCommentTags(t *testing.T) {
...
+				`+k8s:validation:minimum=asdf`,
+			},
+			expectedError: `failed to parse marker comments: failed to parse value for key minimum as JSON: invalid character 'a' looking for beginning of value`,
+		},
+		{
+			t:    types.Float64,
+			name: "invalid: invalid value type",
 			comments: []string{
-				"+k8s:validation:minimum=asdf",
+				`+k8s:validation:minimum="asdf"`,
```
jpbetz, line 135: Do we have bare strings like in use today that we'll need to migrate?

jpbetz (changes requested):
It's going to be great to have this.  Two minor comments then LGTM.  I'd like to get one other lead to review the syntax, but I think it balances the tradeoffs well.

pkg/generators/markers_test.go
alexzielenski: Nope, the `+k8s:validation` prefix is not used anywhere

pkg/generators/markers.go
alexzielenski: great point, ive now factored this out

pkg/generators/markers.go
alexzielenski: Added

pkg/generators/markers.go
alexzielenski: Added

pkg/generators/markers_test.go
alexzielenski: It's the result of some offline discussions I've had with a few folks. The KEP makes reference to multiline support being desired, but no specific syntax was perscibed

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
...
-		// There will always be at least the first element in the split.
-		// If there is no = sign, we want to use the whole line as the key in the
-		// case of boolean marker comments.
-		key := strings.Split(line, "=")[0]
-		subscriptIdx := strings.Index(key, "[")
-		if subscriptIdx == -1 {
-			prevLine = ""
-			prevIndex = -1
+		lastIndex = -1
+		lastKey = ""
```
jpbetz, line 282: Is there a reason not to reset `lastArrayKey = ""` here too?  (I didn't try to prove to myself if it doesn't matter, but it's absence caught my eye).

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
...
 		line = strings.Trim(line, " ")
 
-		// If there was a non-prefixed marker inserted in between, then
-		// reset the previous line and index.
-		if !strings.HasPrefix(line, "+"+prefix) {
-			prevLine = ""
-			prevIndex = -1
-			continue
-		}
+		// Make sure last key gets reset if we `continue`
```
jpbetz, line 276:
```suggestion
		// Make sure last vars get reset if we `continue`
```

And move comment directly above `last*` vars?  (This comment did save me time when reading this)

jpbetz (commented):
just a couple nits

pkg/generators/markers_test.go
jpbetz: > The KEP makes reference to multiline support being desired, but no specific syntax was perscibed

I think this is OK.  It's developer facing. It's not part of our forever APIs. We could change it if we really wanted to.  Number of call sites in k/k will probably be in the 100s..

Jefftree: sgtm

pkg/generators/markers.go
alexzielenski: moved

pkg/generators/markers.go
alexzielenski: nope, was a mistake. ive added it

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
...
+// have the syntax:
+//
+//	'marker' + "key>value"
+//	'marker' + "key>value"
+//
+// Successive usages of the same raw string key results in concatenating each
+// line with `\n` in between. It is an error to use `=` to assing to a previously
+// assigned key
+// (in contrast to types.ExtractCommentTags which allows array-typed
+// values to be specified using `=`).
```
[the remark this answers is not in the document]
alexzielenski: moved

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
 	return commentTags, nil
 }
 
-// lintArrayIndices attempts to prevent:
-// 1. Skipped indices
-// 2. Non-consecutive indices
-func lintArrayIndices(markerComments []string, prefix string) error {
+// ExtractCommentTags parses comments for lines of the form:
```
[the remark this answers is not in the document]
alexzielenski: changed

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
...
+//
+//	'marker' + "key=value".
+//
+// Values are optional; "" is the default.  A tag can be specified more than
+// one time and all values are returned.  If the resulting map has an entry for
+// a key, the value (a slice) is guaranteed to have at least 1 element.
```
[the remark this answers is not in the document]
alexzielenski: changed

pkg/generators/markers.go
```diff
@@ -231,119 +231,189 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
 	return commentTags, nil
 }
 
-// lintArrayIndices attempts to prevent:
-// 1. Skipped indices
-// 2. Non-consecutive indices
-func lintArrayIndices(markerComments []string, prefix string) error {
+// ExtractCommentTags parses comments for lines of the form:
+//
+//	'marker' + "key=value".
```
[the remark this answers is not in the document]
alexzielenski: I didnt take exactly this but added an explicit example for the assignment-less case

jpbetz:
LGTM once test issue is cleared (some kind of duplication)?

pkg/generators/markers.go
```diff
@@ -231,119 +231,202 @@ func ParseCommentTags(t *types.Type, comments []string, prefix string) (CommentT
...
+//
+//	or to specify truthy boolean keys:
+//
+//	'marker' + "key"
+//
+// Values are optional; "" is the default.  A tag can be specified more than
+// one time and all values are returned.  Returns a map with an entry for
+// for each key and a value.
+//
+// Similar to version from gengo, but this version support only allows one
```
thockin, line 253: Previous paragraph says "A tag can be specified more than one time and all values are returned." but this says "only allows one value".

Am I misunderstanding?

thockin (commented):
Thanks!

pkg/generators/markers.go
alexzielenski: Gah, that was missed when modifying the original comment. 

The correct doc should read only one value per key. I'll correct that

PR #449, opened 2024-02-13
feature: Allow validation comment tags to be merged through aliases

alexzielenski:
We have a number of shared Go types used in k8s API. It would be useful if we could use aliases to override validation of a base type without creating a new type (so it is transparent to Go). 

One example is NodeSelectorTerm:

https://github.com/kubernetes/kubernetes/blob/2ce04fc04bf2cbbbacf2f184fd9ebd4e99d65430/pkg/apis/core/types.go#L2830-L2847

There are separate validations for `MatchExpressions` vs `MatchField` both with type `NodeSelectorRequirement` despite the type being the same. 

https://github.com/kubernetes/kubernetes/blob/7972f0309ce8bad3292f3291718361367b2e58fe/pkg/apis/core/validation/validation.go#L4384-L4396

It wouldn't be good to write complex CEL `.all` quantification expressions due to the fact that the error messages wouldnt be reported in the correct locations.

This PR makes it so we can define separate types overlaying a common type through aliases, and our validation comment tags still function. The tags specified closest to the the field take precedence, and we follow the type and aliases thereafter.

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -62,8 +62,20 @@ func (c *CELTag) Validate() error {
 	return nil
 }
 
-// CommentTags represents the parsed comment tags for a given type. These types are then used to generate schema validations.
-type CommentTags struct {
+// commentTags represents the parsed comment tags for a given type. These types are then used to generate schema validations.
+// These only include the newer prefixed tags. The older tags are still supported,
+// but are not included in this struct. Comment Tags are transformed into a
+// *spec.Schema, which is then combined with the older marker comments to produce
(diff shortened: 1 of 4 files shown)


alexzielenski:
/cc @Jefftree 
curious what you think

jpbetz:
Putting validation on an alias type helps a lot here.

Given how hard it is to expose the validation data in OpenAPI, I'd be okay with "internal only" declarative validation for these cases, at least until we have more time to look into ways to either merge or inject validation into shared types.

PR #451, opened 2024-02-14
Add support for common name formats to validation markers

Schnides123:


pkg/generators/markers.go
```diff
@@ -91,9 +91,17 @@ func (c commentTags) ValidationSchema() (*spec.Schema, error) {
 		SchemaProps: c.SchemaProps,
 	}
 
-	if len(c.CEL) > 0 {
+	if res.AllOf != nil {
+		res.AllOf = append([]spec.Schema{}, res.AllOf...)
+	}
```
alexzielenski, line 96: can remove i think

pkg/generators/markers.go
```diff
@@ -164,6 +183,13 @@ func (c commentTags) Validate() error {
 		err = errors.Join(err, fmt.Errorf("invalid CEL tag at index %d: %w", i, celError))
 	}
 
+	if c.Format != "" {
+		_, ok := NameFormats[c.Format]
```
alexzielenski, line 187: We have accepted formats that are not name formats. See https://github.com/kubernetes/kubernetes/blob/69b0699610099f239296c2659a5f936a6573bf59/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/formats.go#L26-L51 for official list. These shouldnt desugar into CEL and should be emitted into the spec.Schema `Format` field

pkg/generators/markers_test.go
```diff
@@ -512,6 +513,75 @@ func TestParseCommentTags(t *testing.T) {
 	}
 }
 
+func TestFormat(t *testing.T) {
+
+	formatNames := []string{
```
alexzielenski, line 518: Can we use the `NameFormats` official list rather than duplicating here

pkg/generators/markers_test.go
```diff
@@ -512,6 +513,75 @@ func TestParseCommentTags(t *testing.T) {
...
+	}{}
+
+	for _, formatName := range formatNames {
+
+		cases = append(cases, struct {
+			t             *types.Type
+			name          string
+			comments      []string
+			expected      *spec.Schema
+			expectedError string
```
alexzielenski, line 544: Add error cases for invalid format and non-string format

alexzielenski (commented):
looking great. requested a few changes

alexzielenski:
This PR depends on https://github.com/kubernetes/kubernetes/pull/124622

PR #452, opened 2024-02-16
refactor: make CommentTags internal

alexzielenski:
I think it more straightforward to convert the commentTags into a spec.Schema and generate from that. Especially as we begin to add more "virtual" markers to the struct rather than direct spec.SchemaProps. Generating from an intermediate `spec.SchemaProps` I think will make things more maintainable in the long run

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -62,8 +62,20 @@ func (c *CELTag) Validate() error {
 	return nil
 }
 
-// CommentTags represents the parsed comment tags for a given type. These types are then used to generate schema validations.
-type CommentTags struct {
+// commentTags represents the parsed comment tags for a given type. These types are then used to generate schema validations.
+// These only include the newer prefixed tags. The older tags are still supported,
+// but are not included in this struct. Comment Tags are transformed into a
+// *spec.Schema, which is then combined with the older marker comments to produce
+// the generated OpenAPI spec.
+//
+// List of tags not included in this struct:
+//
+// - +optional
+// - +default
+// - +listType
+// - +listMapKeys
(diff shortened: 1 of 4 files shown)


pkg/generators/markers.go
```diff
@@ -73,8 +85,36 @@ type CommentTags struct {
...
 }
 
+// Returns the schema for the given CommentTags instance.
+// This is the final authoritative schema for the comment tags
+func (c commentTags) ValidationSchema() (*spec.Schema, error) {
+	res := spec.Schema{
+		SchemaProps: c.SchemaProps,
+	}
+
+	if res.AllOf != nil {
```
Jefftree, line 95: What is this line for? Isn't AllOf already part of the schemaprops?

alexzielenski: remnant from a planned change. removed

pkg/generators/markers_test.go
```diff
@@ -369,16 +392,20 @@ func TestParseCommentTags(t *testing.T) {
...
+					Extensions: map[string]interface{}{
+						"x-kubernetes-validations": []interface{}{
+							map[string]interface{}{
+								"rule":    "oldSelf == self",
+								"message": "cant change",
+							},
+							map[string]interface{}{
+								"rule":            "self > 5",
+								"message":         "must be greater than 5",
+								"optionalOldSelf": true,
```
Jefftree, line 406: We're not losing any information now that this isn't a pointer right?

alexzielenski: correct. since the extensions map is intended to represent the JSON structure of things, I also think it is better that it is not a pointer

PR #454, opened 2024-02-23
bugfix: emit enum in schema when used as list or map element

### LINKED ISSUE #450: Bug: +enum type used in array is not reflected in Schema
I've been testing our some of our marker comments on native types. I've found an issue where having an array with +enum does not result in the `items` of the array validating the `enum`.

/cc @jiahuif 

Example:

https://github.com/kubernetes/kubernetes/blob/7ec1a89a509906dad9fd6a4635d7bfc157b47790/staging/src/k8s.io/api/core/v1/types.go#L331-L335

PersistentVolumeSpec.AccessModes above is a `[]PersistentVolumeAccessMode`. This type [is tagged with `+enum`](https://github.com/kubernetes/kubernetes/blob/7ec1a89a509906dad9fd6a4635d7bfc157b47790/staging/src/k8s.io/api/core/v1/types.go#L778C1-L792C1).

Unfortunately the generated OpenAPI [does not include](https://github.com/kubernetes/kubernetes/blob/master/pkg/generated/openapi/zz_generated.openapi.go) (can't direct link line in large file) the enum values in the `items` for this field:

```go

func schema_k8sio_api_core_v1_PersistentVolumeSpec(ref common.ReferenceCallback) common.OpenAPIDefinition {
	return common.OpenAPIDefinition{
		Schema: spec.Schema{
				Description: "PersistentVolumeSpec is the specification of a persistent volume.",
				Type:        []string{"object"},
					...
					"accessModes": {
							Extensions: spec.Extensions{
								"x-kubernetes-list-type": "atomic",
							},
						},
							Description: "accessModes contains all ways the volume can be mounted. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#access-modes",
							Type:        []string{"array"},
							Items: &spec.SchemaOrArray{
								Schema: &spec.Schema{
										Default: "",
										Type:    []string{"string"},
										Format:  "",
									},
								},
							},
						},
					},
...
}
```

alexzielenski:
fixes #450

Doing this I feel like we are getting closer to need a refactor since when we add a feature there is so much conditional code we end up missing spots. But I am wondering how tolerant we are of possibly changing the generated schema?

/assign @Jefftree

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -942,6 +942,10 @@ func (g openAPITypeWriter) generateMapProperty(t *types.Type) error {
 	typeString, format := openapi.OpenAPITypeFormat(elemType.String())
 	if typeString != "" {
 		g.generateSimpleProperty(typeString, format)
+		if enumType, isEnum := g.enumContext.EnumType(t.Elem); isEnum {
+			// original type is an enum, add "Enum: " and the values
+			g.Do("Enum: []interface{}{$.$},\n", strings.Join(enumType.ValueStrings(), ", "))
+		}
 		g.Do("},\n},\n},\n", nil)
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


Jefftree:
/lgtm
/hold (in case you want more eyes)

What kind of changes were you thinking? Emitting additional validation information in the OpenAPI seems like a net positive to me.

alexzielenski:
/hold cancel

I meant that the code generator feels like it could use a refactor. It is very painful to add/maintain features

PR #455, opened 2024-02-23
bugfix: dont throw error if +required is set with omitempty

alexzielenski:
I had included a bug into previous PR #448 which throws an error if +required is set on an `omitempty` field which ran counter to my interests.

Took another stab at it, this time including tests :)

/assign @Jefftree 
sorry for the churn.

/hold

Am also going to try to pin dep after this one so will wait for green PR there before merging this

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -81,19 +81,25 @@ func hasOpenAPITagValue(comments []string, value string) bool {
 	return false
 }
 
-func hasRequiredTag(m *types.Member) bool {
-	return types.ExtractCommentTags(
-		"+", m.CommentLines)[tagRequired] != nil
-}
-
-// hasOptionalTag returns true if the member has +optional in its comments or
-// omitempty in its json tags.
-func hasOptionalTag(m *types.Member) bool {
+// isOptional returns error if the member has +optional and +required in
+// its comments. If +optional is present it returns true. If +required is present
+// it returns false. Otherwise, it returns true if `omitempty` JSON tag is present
+func isOptional(m *types.Member) (bool, error) {
 	hasOptionalCommentTag := types.ExtractCommentTags(
 		"+", m.CommentLines)[tagOptional] != nil
-	hasOptionalJsonTag := strings.Contains(
-		reflect.StructTag(m.Tags).Get("json"), "omitempty")
-	return hasOptionalCommentTag || hasOptionalJsonTag
+	hasRequiredCommentTag := types.ExtractCommentTags(
+		"+", m.CommentLines)[tagRequired] != nil
+	if hasOptionalCommentTag && hasRequiredCommentTag {
+		return false, fmt.Errorf("member %s cannot be both optional and required", m.Name)
+	} else if hasRequiredCommentTag {
+		return false, nil
+	} else if hasOptionalCommentTag {
+		return true, nil
+	}
+
+	// If neither +optional nor +required is present in the comments,
+	// infer optional from the json tags.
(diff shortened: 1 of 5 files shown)


pkg/generators/openapi.go
```diff
@@ -81,19 +81,24 @@ func hasOpenAPITagValue(comments []string, value string) bool {
 	return false
 }
 
-func hasRequiredTag(m *types.Member) bool {
-	return types.ExtractCommentTags(
-		"+", m.CommentLines)[tagRequired] != nil
-}
-
 // hasOptionalTag returns true if the member has +optional in its comments or
```
Jefftree, line 84: Please update the docs for this

alexzielenski: I had the same thought behind the implementation in my initial PR which ended up having this bug. I would need to split up `hasOptionalTag` into `hasOptionalCommentTag` and `hasOptoinalJSONStructTag` which ended up feeling less clean when I compared them

pkg/generators/union.go
```diff
@@ -163,7 +163,7 @@ func parseUnionStruct(t *types.Type) (*union, []error) {
 		if types.ExtractCommentTags("+", m.CommentLines)[tagUnionDiscriminator] != nil {
 			errors = append(errors, u.setDiscriminator(jsonName)...)
 		} else {
-			if !hasOptionalTag(&m) {
+			if optional, err := hasOptionalTag(&m); !optional || err != nil {
```
Jefftree, line 166: Do we have users for this code path? Should the error be passed forward?

alexzielenski: The error should already be logged by the other code paths

pkg/generators/openapi.go
```diff
@@ -81,19 +81,25 @@ func hasOpenAPITagValue(comments []string, value string) bool {
...
-func hasRequiredTag(m *types.Member) bool {
-	return types.ExtractCommentTags(
-		"+", m.CommentLines)[tagRequired] != nil
-}
-
-// hasOptionalTag returns true if the member has +optional in its comments or
-// omitempty in its json tags.
-func hasOptionalTag(m *types.Member) bool {
+// hasOptionalTag returns error if the member has +optional and +required in
+// its comments. If +optional is present it returns false. If +required is present
```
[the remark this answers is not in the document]
alexzielenski: renamed

alexzielenski:
added another commit to fix another bug that was exposed when i pulled the changes into k/k. Newly evaluated enums sometimes alias each other and cause duplicate entries.

pkg/generators/openapi.go
```diff
@@ -81,19 +81,25 @@ func hasOpenAPITagValue(comments []string, value string) bool {
...
-	return types.ExtractCommentTags(
-		"+", m.CommentLines)[tagRequired] != nil
-}
-
-// hasOptionalTag returns true if the member has +optional in its comments or
-// omitempty in its json tags.
-func hasOptionalTag(m *types.Member) bool {
+// isOptional returns error if the member has +optional and +required in
+// its comments. If +optional is present it returns false. If +required is present
+// it returns true. Otherwise, it returns true if `omitempty` JSON tag is present
```
Jefftree, line 86:
```suggestion
// its comments. If +optional is present it returns true. If +required is present
// it returns false. Otherwise, it returns true if `omitempty` JSON tag is present
```

pkg/generators/union.go
Jefftree: sounds good

pkg/generators/openapi.go
alexzielenski: gah, ty

PR #456, opened 2024-02-24
Support defining a custom schema from raw json

ryantxu:
The ability to add [OpenAPIV3Definition()](https://github.com/kubernetes/kube-openapi/tree/master/pkg/generators#custom-openapi-type-definitions) is essential when trying to integrate existing complex types, however it requires importing this library.

For simple types, this can be avoided using:
```
    func (_ Time) OpenAPISchemaType() []string { return []string{"string"} }
    func (_ Time) OpenAPISchemaFormat() string { return "date-time" }
```

This PR proposes adding a new function that lets you return a raw spec directly
```
func (FooV3Raw) OpenAPISchemaJSON() json.RawMessage {
	return json.RawMessage(`{
		"description": "something custom",
		"type": "object",
		"required": ["type", "value"],
		"properties": {
			"type": {
				"type": "string"
			},
			"value": {
				"type": "string"
			}
		}
	}`)
}
```

TODO:
- [ ] add this to the docs README

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -271,6 +271,16 @@ func hasOpenAPIV3DefinitionMethod(t *types.Type) bool {
 	return false
 }
 
+func hasOpenAPISchemaJSONMethod(t *types.Type) bool {
+	for mn, mt := range t.Methods {
+		if mn != "OpenAPISchemaJSON" {
+			continue
+		}
+		return methodReturnsValue(mt, "encoding/json", "RawMessage")
+	}
+	return false
+}
+
 func hasOpenAPIDefinitionMethod(t *types.Type) bool {
 	for mn, mt := range t.Methods {
 		if mn != "OpenAPIDefinition" {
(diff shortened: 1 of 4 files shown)


ryantxu:
For context -- we are aiming to include values from https://github.com/grafana/grafana-plugin-sdk-go in apis exposed using kube-openapi.

Any better suggestions for function naming?

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -103,6 +104,12 @@ func schema_test_integration_testdata_custom_FooV3OneOf(ref common.ReferenceCall
 	})
 }
 
+func schema_test_integration_testdata_custom_FooV3Raw(ref common.ReferenceCallback) common.OpenAPIDefinition {
+	schema := spec.Schema{}
+	_ = schema.UnmarshalJSON(custom.FooV3Raw{}.OpenAPISchemaJSON())
```
ryantxu, line 109: ignoring the error is not great... panic seems worse

pkg/generators/openapi.go
```diff
@@ -271,6 +271,16 @@ func hasOpenAPIV3DefinitionMethod(t *types.Type) bool {
 	return false
 }
 
+func hasOpenAPISchemaJSONMethod(t *types.Type) bool {
+	for mn, mt := range t.Methods {
+		if mn != "OpenAPISchemaJSON" {
```
ryantxu, line 276: any better suggestions for a name?

ryantxu:
this is not working as i hoped it might... will think some more and revist if it still makes sense

PR #458, opened 2024-02-26
Use gengo/v2 in kube-openapi

thockin:
This is needed to support Go workspaces in k/k.

This is done as several small commits so each step can be seen. It has the unfortunate effect of not compiling at each step.

@liggitt @alexzielenski @jpbetz


cmd/openapi-gen/args/args.go
```diff
@@ -18,59 +18,61 @@ package args
...
+func (args *Args) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&args.OutputDir, "output-dir", "",
+		"the base directory under which to generate results")
+	fs.StringVar(&args.OutputPkg, "output-pkg", "",
+		"the base Go import-path under which to generate results")
+	fs.StringVar(&args.OutputFile, "output-file", "generated.openapi.go",
+		"the name of the file to be generated")
+	fs.StringVar(&args.GoHeaderFile, "go-header-file", "",
+		"the path to a file containing boilerplate header text; the string \"YEAR\" will be replaced with the current 4-digit year")
+	fs.StringVarP(&args.ReportFilename, "report-filename", "r", "-",
```
liggitt, line 59: nit: use the `c.ReportFilename` current field value (defaulted in `New()`) rather than stomping a hard-coded string `-` here?

pkg/generators/config.go
```diff
@@ -51,36 +49,31 @@ func DefaultNameSystem() string {
...
 
-	return generator.Packages{
-		&generator.DefaultPackage{
-			PackageName: filepath.Base(arguments.OutputPackagePath),
-			PackagePath: arguments.OutputPackagePath,
-			HeaderText:  header,
-			GeneratorFunc: func(c *generator.Context) (generators []generator.Generator) {
+	return []generator.Target{
+		&generator.SimpleTarget{
+			PkgName:       filepath.Base(args.OutputDir),
```
liggitt, line 68: is `filepath.Base(args.OutputDir)` actually correct for finding the package name, or is it best-effort / usually-correct? if the latter, comment here to note that limitation and possibly for a follow-up improvement

would `path.Base(args.OutputPackage)` be more likely to be correct (though still possibly best-effort) - `path` instead of `filepath`, `customArgs.OutputPackage` instead of `arguments.OutputBase`?

liggitt (commented):
just a couple comments

is the k/k PR updated with a temporary `replace` of this to make sure it works properly in context there?

liggitt:
hmm, integration test CI looks unhappy

thockin:
Yeah, I am looking at it.

TL;DR: it seems to run openapi-gen and then re-run it witrh --verify.  I must be misunderstanding it.

Jefftree:
We have a separate go.mod file for the integration test directory. I'm assuming the go.mod there needs to be updated as well

thockin:
Test should pass now.

The integration test at HEAD is B-R-O-K-E-N.  Running it man ually leaves uncommitted changes, but the test passes.  The test itself runs openapi-gen (which generates a delta) then runs a verify against the fresh delta, which passes.  The test could literally never fail (unless it exits non-zero).

I don't know why the delta exists.  It seems to be a regression in formatting?  See https://github.com/kubernetes/kube-openapi/pull/458/commits/d1213fd88d6d30090194811a5edbb2b6392237e8

If the delta is expected, it's fixed here.  If not, someone show me why?

This PR now changes the test to generate code into the temp dir and diff, just like it does for the reports and other files.

thockin:
I can't really fix k/k until this merges, but almost all of these commits came straight out of my k/k tree

/remove hold

cmd/openapi-gen/args/args.go
thockin: fixed in next push

pkg/generators/config.go
thockin: Correct on Pkg vs Dir.

filepath touches a few more places.

PR #460, opened 2024-02-27
bugfix: resolve non-determinism in deduplicating enum comments

alexzielenski:
649db6989aaecdd64cc4ea16b76e0d0067664001 included a bugfix for duplicated enum values, but did not account for the re-exported enum values not having identical comments. This led to non-deterministic iteration order over the enum types and an output that would change descriptions almost every time you ran it

This PR is a simple fix by applying a deterministic tie breaker (length). If length is equal, then we take the alphabetically lower godoc.


thockin:
Confirmedt that this fixes my problem, and I no longer see it flipping.

Issue #461, opened 2024-02-27
Bugs from recent changes - go.mod version invalid for go 1.20 and new interface is missing input -i


vsoch:
I attempting using master here as of the changes today, and ran into two issues. First, the go.mod specifies 1.21.3, which is invalid for older verisons of go (possibly OK for newer but we are using 1.20):

```console
17.96 /tmp/kube-openapi/go.mod:3: invalid go version '1.21.3': must match format 1.23
17.96 make: *** [Makefile:350: /workspace/bin/openapi-gen] Error 1
------
```

When I tweak that, the binary that builds is missing the `-i` for input file parameter, a new change in the UI that doesn't seem to be documented. Here is the UI before (two days ago):

![image](https://github.com/kubernetes/kube-openapi/assets/814322/d35c8dd1-63c3-4219-8373-b9c7efe297d9)

Note the `-i` for input dirs.

![image](https://github.com/kubernetes/kube-openapi/assets/814322/36cffa59-537e-4ef2-93dd-4b0f9254fee3)

And now after - there is no `-i`

![image](https://github.com/kubernetes/kube-openapi/assets/814322/0929de99-89e6-4f70-90d1-d7aa5716d6f2)

Noted here:

![image](https://github.com/kubernetes/kube-openapi/assets/814322/5e373f24-87cb-4e4d-bf29-5e3be92e0557)

There aren't releases (that I can pin to use) or even recent tags (the most recent with release- is from [3 years ago](https://github.com/kubernetes/kube-openapi/tree/release-1.22)) so I don't have any choice but to pin the commit from two days ago (which works). It would be great if there were guidance on either of the above. We can't just update the entire version of go (at least right now) and pinning a commit is probably not best. 

Ping @thockin for guidance who I think made the recent changes. Thank you!

thockin:
Hihi.

We considered whether we should make a "v2" of this and decided not to, but the implication of that is that downstreams who use kube-openapi directly are exposed to this breaking change.

First, the X.Y vs. X.Y.Z change.  Unfortunately, Go is doing this by default.  We could manually reset it to "1.21" for now but eventually we won't want to keep patching that by hand.  @alexzielenski or @jpbetz or @liggitt may have an opinion, but I am OK with that.

[...]

PR #462, opened 2024-02-27
Drop go.mod version to 1.20

thockin:
Older version of Go can't parse "X.Y.Z".

xref #461 

@vsoch


go.mod
```diff
@@ -1,6 +1,9 @@
 module k8s.io/kube-openapi
 
-go 1.21.3
+// Keep this as X.Y (instead of X.Y.Z, which Go 1.21 wants, but Go 1.20 does
+// not support) until we think it is reasonable for consumers to be expected to
+// use Go 1.21 or higher.
+go 1.21
```
liggitt, line 6: go 1.21 is really hardnosed about this, and propagates up go versions from dependencies' go directives. 

That means if gengo/v2 says it requires go1.21.x, go 1.21 will refuse to build this module unless it also says it requires go1.21.x

if we ever have to bump to HEAD kube-openapi versions in k/k release branches, go1.21 will refuse to build *those* unless we also bump the release branch k/k go.mod files up to go1.21, which we can't currently do for compat reasons (this is hopefully getting ironed out by go1.23 in https://github.com/golang/go/issues/65573)

If possible, we should leave kube-openapi (and by extension, gengo/v2, I guess) declaring a minimum go version of the oldest in-support k/k branch. That's go1.20 currently. Did anything in gengo/v2 or kube-openapi require anything from go1.21 or go1.22?

thockin: It's PROBABLY ok to use 1.20, but hard to test.  If I revert go.mod to say "1.20"

```
$ head go.mod 
module k8s.io/kube-openapi

// Keep this as X.Y (instead of X.Y.Z, which Go 1.21 wants, but Go 1.20 does
// not support) until we think it is reasonable for consumers to be expected to
// use Go 1.21 or higher.
go 1.20

require (
	github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46
	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a

$ go build ./cmd/openapi-gen/
go: updates to go.mod needed; to update it:
	go mod tidy
```

liggitt: to revert this to go1.20 you also have to revert gengo/v2 to go1.20 ... it's viral

thockin:
PTAL

PR #465, opened 2024-03-22
bugfix: change log level from error to info

caozhuozi:
I'm not sure why the code uses `Errorf` here while the result of `parseSymbolReference` is `ok`.

I found this issue when I was using `openapi-gen.go` to generate the `/test/integration/testdata/defaults` package (to have a try) and the following error logs appeared.
```
E0322 16:21:36.249229   11020 openapi.go:707] k8s.io/kube-openapi/test/integration/testdata/defaults.ConstantValue, k8s.io/kube-openapi/test/integration/testdata/defaults
E0322 16:21:36.249997   11020 openapi.go:707] k8s.io/kube-openapi/test/integration/testdata/defaults.ConstantValue, k8s.io/kube-openapi/test/integration/testdata/defaults
E0322 16:21:36.250239   11020 openapi.go:707] k8s.io/kube-openapi/test/integration/testdata/enumtype.FruitApple, k8s.io/kube-openapi/test/integration/testdata/defaults
E0322 16:21:36.250397   11020 openapi.go:707] k8s.io/kube-openapi/test/integration/testdata/enumtype.FruitApple, k8s.io/kube-openapi/test/integration/testdata/defaults
E0322 16:21:36.250527   11020 openapi.go:707] k8s.io/kube-openapi/test/integration/testdata/defaults.ConstantValue, k8s.io/kube-openapi/test/integration/testdata/defaults

```
But the files were generated successfully.

Hope I don't have any misunderstanding here.

PR #468, opened 2024-04-11
Use klog.Info for defaultFromComments instead of Error

tarvitz:
I found running openapi generator produces information like the following:

```bash
$ ./hack/update-openapi.sh
Generating openapi code for 1 targets
E0411 14:46:21.484926   54800 openapi.go:707] go.company.name/slacker/meta/pkg/apis/configs/v1alpha1.ReviewersPingLate, go.company.name/slacker/meta/pkg/apis/configs/v1alpha1
```
Seems there should be Infof instead of Errorf


Jefftree:
Duplicate of https://github.com/kubernetes/kube-openapi/pull/465.
/close

PR #469, opened 2024-04-12
Allow marker comments to extend validations for subschemas

alexzielenski:
This extends current marker syntax to support extending the subschemas of a struct inside `items`, `properties`, or `additionalProperties`. This is useful in cases like declarative validation for native types where we have many shared types whose validations change at point-of-use, so it does not make sense to annotate the type. And using an alias is too disruptive.

Example:

```go
package foo

// +k8s:openapi-gen=true
// +k8s:validation:properties:field:maxLength=10
// +k8s:validation:properties:aliasMap:additionalProperties:pattern>^foo$
type Blah struct {
	// +k8s:validation:items:cel[0]:rule="self.length() % 2 == 0"
        // +k8s:validation:items:cel[0]:message="length must be even"
	Field MyAlias `json:"field"`

	// +k8s:validation:additionalProperties:maxLength=10
	AliasMap MyAliasMap `json:"aliasMap"`
}
		
type MyAliasMap map[string]MyAlias
type MyAlias []string
```

In the above example we can override subschemas at multiple different levels, with type checking supported.

This yields schema:

```yaml
type: object
allOf:
    properties:
        field: 
            items: {maxLength: 10}
        aliasMap: {pattern: ^foo$}
properties:
    field:
        type: array
        items:
            type: string
        allOf:
            items:
                x-kubernetes-validations:
                - rule: self.length() % 2 == 0
                  message: length must be even
    aliasMap:
        additionalProperties: {type: string}
        allOf:
            additionalProperties:
                maxLength: 10
```

The value validations are placed in an allOf inside the object they were attached to and override subproperties. New structure is not allowed to be placed in these tags, only validations. We validate that any named properties are also in the structure of the Go type.

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -75,7 +75,29 @@ func (c *CELTag) Validate() error {
 // - +listMapKeys
 // - +mapType
 type commentTags struct {
-	spec.SchemaProps
+	Nullable         *bool         `json:"nullable,omitempty"`
+	Format           *string       `json:"format,omitempty"`
+	Maximum          *float64      `json:"maximum,omitempty"`
+	ExclusiveMaximum *bool         `json:"exclusiveMaximum,omitempty"`
+	Minimum          *float64      `json:"minimum,omitempty"`
+	ExclusiveMinimum *bool         `json:"exclusiveMinimum,omitempty"`
+	MaxLength        *int64        `json:"maxLength,omitempty"`
+	MinLength        *int64        `json:"minLength,omitempty"`
+	Pattern          *string       `json:"pattern,omitempty"`
+	MaxItems         *int64        `json:"maxItems,omitempty"`
+	MinItems         *int64        `json:"minItems,omitempty"`
+	UniqueItems      *bool         `json:"uniqueItems,omitempty"`
+	MultipleOf       *float64      `json:"multipleOf,omitempty"`
+	Enum             []interface{} `json:"enum,omitempty"`
+	MaxProperties    *int64        `json:"maxProperties,omitempty"`
+	MinProperties    *int64        `json:"minProperties,omitempty"`
+
+	// Nested commentTags for extending the schemas of subfields at point-of-use
+	// when you cant annotate them directly. Cannot be used to add properties
+	// or remove validations on the overridden schema.
+	Items                *commentTags            `json:"items,omitempty"`
+	Properties           map[string]*commentTags `json:"properties,omitempty"`
+	AdditionalProperties *commentTags            `json:"additionalProperties,omitempty"`
 
 	CEL []CELTag `json:"cel,omitempty"`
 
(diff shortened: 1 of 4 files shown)


pkg/generators/markers.go
```diff
@@ -75,7 +75,29 @@ func (c *CELTag) Validate() error {
...
+	MinItems         *int64        `json:"minItems,omitempty"`
+	UniqueItems      *bool         `json:"uniqueItems,omitempty"`
+	MultipleOf       *float64      `json:"multipleOf,omitempty"`
+	Enum             []interface{} `json:"enum,omitempty"`
+	MaxProperties    *int64        `json:"maxProperties,omitempty"`
+	MinProperties    *int64        `json:"minProperties,omitempty"`
+
+	// Nested commentTags for extending the schemas of subfields at point-of-use
+	// when you cant annotate them directly. Cannot be used to add properties
+	// or remove validations on the overriden schema.
```
Jefftree, line 97: overridden

Jefftree (commented):
I feel like there are quite a few implicit changes here, would you be able to separate them out into another PR/commit to explain the relevance? The changes move the codebase in a good direction but I'm a bit confused about the interdependency between them and the PR description goal.

OTOH

- a set of fields being transformed into ptrs
- deterministic output
- validations such as minItems/etc should only be attached corresponding types

test/integration/pkg/generated/openapi_generated.go
```diff
@@ -1015,9 +1015,7 @@ func schema_test_integration_testdata_valuevalidation_Foo(ref common.ReferenceCa
 	return common.OpenAPIDefinition{
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
-				Type:          []string{"object"},
-				MinProperties: ptr.To[int64](1),
-				MaxProperties: ptr.To[int64](5),
+				Type: []string{"object"},
```
jpbetz, line 1018: I must have missed something.  Why is this change needed?

jpbetz (commented):
Looks good. A few very minor comments.

test/integration/pkg/generated/openapi_generated.go
alexzielenski: ValidateType was adjusted to continue validation for non-opaque struct types (mostly to enable recursive validation of the new subschema fields):

https://github.com/kubernetes/kube-openapi/pull/469/files/626a771d522a239689b64c1b50ee0ebf37215d8b#diff-72f8260914152fe9557b05d756ec94854184645e991ff755855628428d6d7a18L180-R294

The existing test for this had used `minProperties` and `maxPropertes` on a struct (rather than map), which now results in error.

I somewhat arbitrarily chose to continue this restricting (rather than widening those fields to be used for structs) because it seemed to me there did not seem to be much of a use case. Happy to allow them for structs again if someone else things it could be useful

pkg/generators/markers.go
```diff
@@ -220,16 +333,57 @@ func (c commentTags) ValidateType(t *types.Type) error {
...
+	if c.Items != nil && !isArray {
+		err = errors.Join(err, fmt.Errorf("items can only be used on array types"))
+
+		if err == nil {
+			err = errors.Join(err, c.Items.ValidateType(t))
+		}
+	}
+	if c.Properties != nil {
+		if resolvedType.Kind != types.Struct {
+			err = errors.Join(err, fmt.Errorf("properties can only be used on struct types"))
```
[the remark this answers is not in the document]
alexzielenski: Added

test/integration/pkg/generated/openapi_generated.go
alexzielenski: Ended up allowing MinProperties/MaxProperties since it seemed unnecessary restriction and to keep this PR focused

alexzielenski:
> I feel like there are quite a few implicit changes here, would you be able to separate them out into another PR/commit to explain the relevance? The changes move the codebase in a good direction but I'm a bit confused about the interdependency between them and the PR description goal.
> 
> OTOH
> 
> * a set of fields being transformed into ptrs
> * deterministic output
> * validations such as minItems/etc should only be attached corresponding types

I split the deterministic output change into a separate commit (no test changes, since theres only 1 possible element so it is more of a future-proofing change)

I also removed the MinProperties/MaxProperties behavior change, so diff is much smaller.

ready for another pass

pkg/generators/markers.go
```diff
@@ -86,9 +108,75 @@ type commentTags struct {
...
+	if c.Format != nil {
+		format = *c.Format
+	}
+	isExclusiveMaximum := c.ExclusiveMaximum != nil && *c.ExclusiveMaximum
+	isExclusiveMinimum := c.ExclusiveMinimum != nil && *c.ExclusiveMinimum
+	isUniqueItems := c.UniqueItems != nil && *c.UniqueItems
+	pattern := ""
+	if c.Pattern != nil {
+		pattern = *c.Pattern
+	}
```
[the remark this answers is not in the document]
alexzielenski: It allows us to disambiguate empty from unset. It doesnt affect anything but I thought that while I was copying the fields over it'd be better to have this ability.

alexzielenski:
/hold 

@jpbetz While trying to use this I've ran into the problem that CEL expressions inside `allOf` are not validated, since it is not part of the "structural" portion of the schema. Is this something that will ever change in the future? I'm also curious why this is the case, since CEL expressions are definitely value validations and don't impact schema structure unlike other X-extensions

alexzielenski:
Ive tested this extensively as part of [#](https://github.com/kubernetes/kubernetes/pull/123163)

To use CEL within these still requires https://github.com/kubernetes/kubernetes/pull/124381 but I think this is ready to merge

/cc @jpbetz 
/remove-hold

Jefftree:
lgtm from me, I'll leave lgtm with @jpbetz if he wants another pass

pkg/generators/openapi_test.go
```diff
@@ -2358,10 +2358,6 @@ func TestMarkerComments(t *testing.T) {
 		package foo
 
 		// +k8s:openapi-gen=true
-		// +k8s:validation:maxProperties=10
-		// +k8s:validation:minProperties=1
-		// +k8s:validation:exclusiveMinimum
-		// +k8s:validation:exclusiveMaximum
```
jpbetz, line 2364: Sorry if this is obvious, but why are we removing these tags from this test input?

jpbetz:
One question about the removed test cases then this looks ready to merge.

pkg/generators/openapi_test.go
alexzielenski: added back

jpbetz:
/approve
/lgtm
Leaning on @Jefftree's review here, but it all looked reasonable to me.

PR #474, opened 2024-04-30
Add dependabot

lucacome:
Adds dependabot to keep Go and GitHub Actions dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


liggitt:
oof... this makes kube-openapi proactively bump all dependencies in ways that impact Kubernetes' API surface and makes us unable to actually consume kube-openapi updates. I don't think we want this done globally until we insulate ourselves from API breaks as a result of dependency bumps

xref https://github.com/kubernetes/kubernetes/pull/126787/files#r1727343696

PR #475, opened 2024-04-30
Bump github.com/NYTimes/gziphandler to v1.1.1

lucacome:
Bumps github.com/NYTimes/gziphandler to the latest tag


lucacome:
can we merge this?

PR #478, opened 2024-06-20
int or string prototype [WIP]

Jefftree:
Prototype of adding x-int-or-string

See https://github.com/kubernetes/kubernetes/pull/125612 for generation


Jefftree:
@alexzielenski fyi

PR #481, opened 2024-06-30
Bubble up buildOperations error.

godwinpang:
Errors encountered in `buildOperations` should not fail silently. 
I ran into an issue where `BuildOpenAPISpecFromRoutes` (https://github.com/kubernetes/kube-openapi/blob/b456828f718bab62dc3013d192665eb3d17f8fe9/pkg/builder/openapi.go#L52) would return indeterministic results due to `buildOperations` errors not being bubbled up.


austince (approved):

Jefftree:
Thanks!
/lgtm
/approve

godwinpang:
@Jefftree fixed the gofmt issue, could I get a followup lgtm?

PR #482, opened 2024-07-04
Render code blocks properly in OpenAPI

Jefftree:
Render code blocks properly in OpenAPI.

Partially solves https://github.com/kubernetes/website/issues/35712

/cc @sftim @robert-cronin
/assign @jpbetz @sttts

### DIFF
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -50,6 +50,8 @@ const (
 	tagValueFalse = "false"
 )
 
+const codeBlockDelimiter = "```"
+
 // Used for temporary validation of patch struct tags.
 // TODO: Remove patch struct tag validation because they we are now consuming OpenAPI on server.
 var tempPatchTags = [...]string{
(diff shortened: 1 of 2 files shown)


sftim (commented):
Do we want to handle code blocks inside block quotes? I imagine not, but we might do. Calling this out as a case we _could_ choose to handle.

Jefftree:
The motivation was code blocks but imo anything within the block quote `\`\`\`` should NOT be interpreted as markdown and should be outputted as is with no reformatting done. That's essentially what this PR is targeting

sftim:
Markdown blockquotes look like:

<pre>
> This is a block quote
> 
> Within the block quote there could be a multiline code snippet
> ```
> echo "Hello, world
> ```

</pre>

Jefftree:
Ah sorry I mixed up the terminology. 

I'm not sure if we have any examples of use cases with blockquotes, so right now I'm leaning probably not. We can re-evaluate if we see a future use case.

sftim:
@jefftree I'm happy to see this land. I can review it but you'd need to either explain how to test this, or give me an explain-it-like-I-use-Go-more-than-once-every-5-years guide to setting up a local dev environment (yes those overlap a bit).

PR #500, opened 2024-08-12
Bump gengo version

thockin:
Fix one breakage after gengo refactored named function results.

I get a test-fail for ./pkg/internal/third_party/go-json-experiment/json when I run locally, but that also fails as far back as when it was added.

@jpbetz


thockin:
Sorry, had to re-push with test/integration/go.mod update

Issue #501, opened 2024-08-12
kube-openapi tests fail at newer versions of go


jpbetz:
Since we depend on kube-openapi from k8s/k8s and other projects using newer versions of go, we need to bump the go version of `kube-openapi` and address this test failure to ensure it is not going to cause problems downstream.

How to reproduce:

```
gvm use go1.22.2
go test -race ./cmd/... ./pkg/...


?   	k8s.io/kube-openapi/cmd/openapi-gen	[no test files]
?   	k8s.io/kube-openapi/cmd/openapi-gen/args	[no test files]
?   	k8s.io/kube-openapi/cmd/openapi2smd	[no test files]
?   	k8s.io/kube-openapi/pkg/builder3/util	[no test files]
ok  	k8s.io/kube-openapi/pkg/aggregator	(cached)
?   	k8s.io/kube-openapi/pkg/common	[no test files]
?   	k8s.io/kube-openapi/pkg/common/restfuladapter	[no test files]
ok  	k8s.io/kube-openapi/pkg/builder	(cached)
ok  	k8s.io/kube-openapi/pkg/builder3	(cached)
ok  	k8s.io/kube-openapi/pkg/cached	(cached)
?   	k8s.io/kube-openapi/pkg/util/jsontesting	[no test files]
?   	k8s.io/kube-openapi/pkg/util/proto/testing	[no test files]
?   	k8s.io/kube-openapi/pkg/util/sets	[no test files]
?   	k8s.io/kube-openapi/pkg/validation/strfmt/bson	[no test files]
ok  	k8s.io/kube-openapi/pkg/generators	7.610s
ok  	k8s.io/kube-openapi/pkg/generators/rules	(cached)
ok  	k8s.io/kube-openapi/pkg/handler	(cached)
ok  	k8s.io/kube-openapi/pkg/handler3	(cached)
ok  	k8s.io/kube-openapi/pkg/idl	(cached) [no tests to run]
ok  	k8s.io/kube-openapi/pkg/internal	(cached)
--- FAIL: TestMarshal (0.06s)
    --- FAIL: TestMarshal/Structs/OmitZero/NonZero (0.00s)
        arshal_test.go:3690: arshal_test.go:1346: Marshal output mismatch:
            got  {
            	"Bool": true,
            	"String": " ",
            	"Bytes": "",
            	"Int": 1,
            	"Uint": 1,
            	"Map": {},
            	"StructScalars": {
            		"Bool": false,
            		"String": "",
[...]

PR #503, opened 2024-08-16
Patch go-json-experiment json to pass tests on go1.22

Jefftree:
Per comment https://github.com/kubernetes/kube-openapi/issues/501#issuecomment-2289891359 and https://github.com/kubernetes/kube-openapi/issues/501#issuecomment-2291112778.

Fixes https://github.com/kubernetes/kube-openapi/issues/501

This is manual patch to include the test fix from https://github.com/go-json-experiment/json/commit/c2a874acdab3b26ac88290cff3b256169f420a48. We cannot directly fast forward to that commit because it would increase our minimum supported go version to 1.21.

Tested manually via `go test`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,16 @@ on:
     branches: [ master ]
 jobs:
   ci:
+    strategy:
+      matrix:
+        go-version: [ '1.20', '1.21', '1.22', '1.23' ]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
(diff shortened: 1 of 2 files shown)


Jefftree:
cc @jpbetz @liggitt @dsnet

liggitt:
make sure the passes on both the current minimum version of go for this repo (go 1.20) and on go 1.22 / 1.23

liggitt:
nit: avoid mentioning the other repo / commit sha in the commit message here or github may ping that commit with a comment every time this commit is picked / merged anywhere

Jefftree:
Removed the commit referencing another commit sha and expanded the testing to test for all go versions 1.20-1.23

PR #504, opened 2024-08-21
WIP: rules: enforce that ListMeta is called "metadata"

pohly:
This catches the following mistake:

   type ResourceSliceList struct {
       metav1.ListMeta `json:"listMeta" protobuf:"bytes,1,opt,name=listMeta"`
       ...
   }

The "listMeta" name matches the Go field name, but the convention is to call that JSON field "metadata".


pkg/generators/rules/naming_convention.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
Jefftree, line 2: nit: 2024

pkg/generators/api_linter.go
```diff
@@ -136,6 +136,7 @@ type apiLinter struct {
 func newAPILinter() *apiLinter {
 	return &apiLinter{
 		rules: []APIRule{
+			&rules.NamingConvention{},
```
Jefftree, line 139: Should we be more specific that this is only referring to ListMeta naming convention?

pkg/generators/rules/naming_convention.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Name returns the name of APIRule
+func (n *NamingConvention) Name() string {
+	return "naming_convention"
+}
+
+// Validate evaluates API rule on type t and returns a list of field names in
+// the type that violate the rule. Empty field name [""] implies the entire
+// type violates the rule.
+func (n *NamingConvention) Validate(t *types.Type) ([]string, error) {
```
Jefftree, line 42: can you add a test for this?

Jefftree (commented):
Minor comments.

cc @liggitt for changes affect api linting

pkg/generators/api_linter.go
pohly: I asked on Slack whether there should be other conventions that this should enforce:
https://kubernetes.slack.com/archives/C0409NGC1TK/p1724256359105049?thread_ts=1723854248.858719&cid=C0409NGC1TK

If there aren't then renaming makes sense.

pkg/generators/rules/naming_convention.go
pohly: Absolutely. It's WIP because I wanted to be sure first that this is going in the right direction.

pkg/generators/api_linter.go
pohly: It's also for ObjectMeta. That wasn't in my original commit, but I've included it in the revised one.

pkg/generators/rules/naming_convention.go
```diff
@@ -0,0 +1,77 @@
...
+	typeName := m.Type.String()
+	switch typeName {
+	case "k8s.io/apimachinery/pkg/apis/meta/v1.ListMeta",
+		"k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta":
+		jsonTag, ok := reflect.StructTag(m.Tags).Lookup("json")
+		if !ok {
+			return false
+		}
+		jsonName := strings.Split(jsonTag, ",")[0]
+		if jsonName != "metadata" {
```
pohly, line 71: "metadata" is excluded from name checking in https://github.com/kubernetes/kube-openapi/blob/76de80e0abd99757a62e1f638f2da51e5f225baf/pkg/generators/rules/names_match.go#L39-L40

The effect is that this mistake is not caught:
```
Items []ResourceSlice `json:"metadata" protobuf:"bytes,2,rep,name=metadata"`
```

An alternative to adding this NamingConvention rule would be to extend the NamesMatch rule:
- Check for ListMeta and ObjectMeta there.
- Those fields must be called "metadata".
- Other fields that are called "metadata" must have matching Go and JSON names.

pohly: Done, PR updated accordingly.

I couldn't resist also addressing this false positive:
```
SomeInt int `json:""`
```

This wasn't reported because "" was also in the `jsonNameBlacklist`.

pohly:
Using this in Kubernetes might be tricky, updating the dependency wasn't straight-forward - see https://github.com/kubernetes/kubernetes/pull/126917 (not including this PR yet).

liggitt:
try rebasing, some other fixups went in last week (https://github.com/kubernetes/kube-openapi/pull/507, https://github.com/kubernetes/kube-openapi/pull/508)

could also wait until https://github.com/kubernetes/kubernetes/pull/126787 gets in and handles some of the regen cleanup

pkg/generators/rules/names_match.go
```diff
@@ -109,14 +100,30 @@ func (n *NamesMatch) Validate(t *types.Type) ([]string, error) {
...
 			}
 		}
 	}
 	return fields, nil
 }
 
+func nameIsOkay(member types.Member, jsonName string) bool {
+	if jsonName == "" {
+		return member.Type.Kind == types.Struct ||
+			member.Type.Kind == types.Pointer && member.Type.Underlying.Kind == types.Struct
```
pohly, line 114: Ooops. This should be member.Type.Elem.

/cancel lgtm

pohly: https://github.com/kubernetes/kubernetes/blob/5ac315faf428567ec023aa7d8c2b89b1d9c9211a/vendor/k8s.io/gengo/v2/types/types.go#L327-L335

pohly: Noticed just in time when looking at the gengo API changes, but missed it earlier when writing the code... sorry!

PR #489, opened 2024-07-09
Bump all Go dependencies

lucacome:
Bumps all the dependencies to the latest version available. Hopefully after this dependabot will start behaving...

PR #507, opened 2024-08-22
disable dependabot

Jefftree:
/assign @liggitt


liggitt:
it's possible we can exclude specific deps we know hit our API surface, but disabling to start sounds good until we recover from https://github.com/kubernetes/kube-openapi/pull/489/files#r1727377659

PR #508, opened 2024-08-22
downgrade govalidator, go-restful, jsonreference

Jefftree:
Per https://github.com/kubernetes/kube-openapi/pull/489/files#r1727377659, revert bumps to govalidator, go-restful, jsonreference

/assign @liggitt


go.mod
```diff
@@ -4,9 +4,9 @@ go 1.20
 
 require (
 	github.com/NYTimes/gziphandler v1.1.1
-	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2
-	github.com/emicklei/go-restful/v3 v3.12.1
-	github.com/go-openapi/jsonreference v0.21.0
+	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
+	github.com/emicklei/go-restful/v3 v3.11.0
+	github.com/go-openapi/jsonreference v0.20.1
```
liggitt, line 9: is `go mod tidy` happy with this? just want to make sure github.com/go-openapi/swag won't force up jsonreference or govalidator

Jefftree: Yes, go-openapi/swag has no dependencies on either: https://github.com/go-openapi/swag/blob/master/go.mod

liggitt:
can you dry-run bumping k/k to the results of this to make sure this does what we need?

```
hack/pin-dependency.sh github.com/kubernetes/kube-openapi=github.com/Jefftree/kube-openapi downgrade-dep
```

then make sure lint-dependencies and update-vendor are happy, and that govalidator didn't jump?

PR #511, opened 2024-10-05
use our own fork of go-yaml

aojea:
We have our own fork of go-yaml https://github.com/kubernetes-sigs/yaml/pull/76 so we can simplify the dependency graph in kubernetes if we end the path in kubernetes-sigs/yaml

```
go mod why gopkg.in/yaml.v3
# gopkg.in/yaml.v3
k8s.io/apimachinery/pkg/util/managedfields
k8s.io/kube-openapi/pkg/util/pro

```

Ref https://github.com/kubernetes-sigs/structured-merge-diff/pull/264

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/tools v0.24.0
 	google.golang.org/protobuf v1.34.2
-	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/gengo/v2 v2.0.0-20240826214909-a7b603a56eb7
 	k8s.io/klog/v2 v2.130.1
--- a/cmd/openapi2smd/openapi2smd.go
+++ b/cmd/openapi2smd/openapi2smd.go
@@ -22,7 +22,7 @@ import (
 	"os"
 
 	openapi_v2 "github.com/google/gnostic-models/openapiv2"
-	yaml "gopkg.in/yaml.v2"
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -48,4 +46,5 @@ require (
 	golang.org/x/sync v0.8.0 // indirect
 	golang.org/x/sys v0.23.0 // indirect
 	golang.org/x/text v0.17.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
aojea, line 49: this is coming from

```
o mod why gopkg.in/yaml.v3
# gopkg.in/yaml.v3
k8s.io/kube-openapi/cmd/openapi2smd
github.com/google/gnostic-models/openapiv2
gopkg.in/yaml.v3
```

aojea: I See, and the type is required, breaking this PR 😢 
```
Error: pkg/util/proto/document_v3.go:291:31: cannot use s.GetDefault().ToRawInfo() (value of type *"gopkg.in/yaml.v3".Node) as *"sigs.k8s.io/yaml/goyaml.v3".Node value in argument to parseV3Interface

```

aojea: @jpbetz should we move `github.com/google/gnostic-models/openapiv2` to our fork too?

aojea: @Jefftree tried to mean, move the yaml imported version, not the package :)

sttts:
/retest

What do we do with the 1.20 CI failure?

aojea:
it is legit, I will investigate more

aojea:
@sttts it should be fixed now, I somehow added unnecessarely the dependency to `test/integration/go.mod`

PR #517, opened 2024-11-24
remove old github.com/golang/protobuf

huww98:
gnostic-models has not been updated for over 1 year. Just use its latest main branch, as in gnostic repo

This should be one step forward to remove old `github.com/golang/protobuf` from the k8s ecosystem. We are not in a hurry, but this cleanup needs to be done sooner or later.


Jefftree:
Looks good, I just need to double check the kubernetes/kubernetes version bump that it doesn't cause any dependency issues.

Jefftree:
https://github.com/kubernetes/kubernetes/pull/128982

huww98:
There is a request https://github.com/google/gnostic-models/issues/8 . But receives no response for over 1 year. Maybe we can try again

Jefftree:
> can we request a tag on the gnostic-models repo so we don't have to switch from a tag to an untagged commit?

ack, I'll work on getting a new version released and probably also bumping some of the extremely outdated packages.

Jefftree:
/hold cancel
Published v0.6.9 for gnostic-models: https://github.com/google/gnostic-models/releases/tag/v0.6.9.

@huww98 can you update this PR to point to the new tag?

huww98:
> can you update this PR to point to the new tag?

Done

.github/workflows/ci.yml
```diff
@@ -34,10 +34,10 @@ jobs:
         go test ./...
     # We set the maximum version of the go directive as 1.20 here
     # because the oldest go directive that exists on our supported
-    # release branches in k/k is 1.20.
+    # release branches in k/k is 1.22.8.
```
liggitt, line 37: Oldest go directive in 1.29 is still go 1.21: https://github.com/kubernetes/kubernetes/blob/release-1.29/go.mod#L9

.github/workflows/ci.yml
```diff
@@ -34,10 +34,10 @@ jobs:
         go test ./...
     # We set the maximum version of the go directive as 1.20 here
```
liggitt, line 35:
```suggestion
    # We set the maximum version of the go directive as 1.21 here
```

huww98: OK. Sorry. Maybe we should just remove these redundant information. To make future updates easier.

go.mod
```diff
@@ -1,15 +1,14 @@
 module k8s.io/kube-openapi
 
-go 1.20
+go 1.21
```
Jefftree, line 3: @liggitt are we able to safely bump this? Per https://github.com/kubernetes/kube-openapi/issues/463 I'd imagine we wanted to keep 1.20 and directly skip to 1.23 once that's the lowest supported k/k version.

Jefftree: Edit: https://github.com/kubernetes/kube-openapi/pull/517#discussion_r1859757625, looks like the oldest go directive is k/k v1.29 @ go1.21

liggitt:
lgtm

PR #519, opened 2024-12-07
Fully ignore unknown tags in ParseCommentTags

jpbetz:
Fix `generators.ParseCommentTags` to not fail on parsing on tags that it ignores anyway.

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -20,9 +20,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 
 	"k8s.io/gengo/v2/types"
 	openapi "k8s.io/kube-openapi/pkg/common"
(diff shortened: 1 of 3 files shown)


thockin:
> I've made this opt-in since a full unit test run demonstrated that we have code that depends on the existing behavior.

Is that VALID, GOOD code or just code that exercised the existing behavior?

pkg/generators/openapi_test.go
```diff
@@ -2386,7 +2387,7 @@ func TestMarkerComments(t *testing.T) {
 			// +k8s:validation:pattern="^foo$[0-9]+"
 			StringValue string
 
-			// +k8s:validation:maxitems=10
+			// +k8s:validation:maxItems=10
```
yongruilin, line 2390: Wondering why this passed the test before?

jpbetz: Because maxItems is never used to check anything in the test :(  If the test had tried to exceed maxItems, it would have been allowed.

I could just remove this validation rules to simplify the test case..

yongruilin: hmmm.. I think it is because the `json.Unmarshal` is [not case-sensitive](https://pkg.go.dev/encoding/json#Unmarshal). So when unmarshal the `commentTags` here https://github.com/kubernetes/kube-openapi/blob/99599406b04fd8d4b08b519eef9461744667043d/pkg/generators/markers.go#L411 , `maxItems` and `maxitems` both are parsed.

thockin: We should not rely on that, I think the JSON we use in k/k is not case-insensitve

yongruilin: k/k is using https://github.com/kubernetes-sigs/json which is case sensitive

jpbetz:
> > I've made this opt-in since a full unit test run demonstrated that we have code that depends on the existing behavior.
> 
> Is that VALID, GOOD code or just code that exercised the existing behavior?

I dug deeper.  One of the problem cases was for the `[` delimited strings Alex added support for with the multiline cel support, and the other was a check for empty string.  I've fixed this to handle both and dropped the options.

pkg/generators/openapi_test.go
jpbetz: I'll follow up on this in a separate issue.

pkg/generators/markers.go
```diff
@@ -61,6 +63,30 @@ func (c *CELTag) Validate() error {
 	return nil
 }
 
+func isKnownTagCommentKey(key string) bool {
```
thockin, line 66: Could you comment this - Would be nice to explain the expected "grammar" for a key here.

thockin:
Edit PR comment to remote words about options?

jpbetz:
> Edit PR comment to remote words about options?

Thanks. Yes, that's misleading. Fixed.

thockin (commented):
LGTM

pkg/generators/markers.go
jpbetz: Sure, added.

thockin: not pushed?

jpbetz: oops, pushed now

thockin:
Thanks!

/lgtm
/approve

PR #522, opened 2025-02-13
Fix Handler Leak

stasos24:
Original source code:
https://github.com/kubernetes/kube-openapi/blob/2c72e554b1e7755b5fbee01cc910063070d5b4ec/pkg/validation/spec/ref.go#L70-L77

Fixed Handler leaking before return

### DIFF
--- a/pkg/validation/spec/ref.go
+++ b/pkg/validation/spec/ref.go
@@ -72,7 +72,7 @@ func (r *Ref) IsValidURI(basepaths ...string) bool {
 		if err != nil {
 			return false
 		}
-
+		rr.Body.Close()
 		return rr.StatusCode/100 == 2
 	}
 


Jefftree:
/assign @sttts @liggitt 
(for approval)

pkg/validation/spec/ref.go
```diff
@@ -72,7 +72,7 @@ func (r *Ref) IsValidURI(basepaths ...string) bool {
 		if err != nil {
 			return false
 		}
-
+		rr.Body.Close()
```
liggitt, line 75: thanks for the leak fix... it looks like IsValidURI is unused in this repo or in kubernetes/kubernetes, and I don't think we actually *want* code paths that can be triggered to make network requests from here ... I'm wondering if we should delete IsValidURI entirely

PR #529, opened 2025-03-18
Fix idl package test naming

Jefftree:
Discovered from https://github.com/kubernetes/kube-openapi/pull/528 and https://github.com/kubernetes/kube-openapi/actions/runs/13925483617/job/38968758746?pr=528. `Example<string>` is a reserved keyword https://go.dev/blog/examples so the examples should be in the same package as the type definition in `pkg/idl/doc.go`

### DIFF
--- a/pkg/idl/listtype_test.go
+++ b/pkg/idl/listtype_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package idl_test
+package idl
 
 // This example shows how to use the listType map attribute and how to
 // specify a key to identify elements of the list. The listMapKey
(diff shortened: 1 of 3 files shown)


yongruilin:
/lgtm
Why it wasn't captured by ci before?

pkg/idl/listtype_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package idl_test
+package idl
```
yongruilin, line 17: How about doing this instead by importing?
```
import (
	_ "k8s.io/kube-openapi/pkg/idl"
)
```

PR #530, opened 2025-03-18
Add test for go 1.24

Jefftree:


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   ci:
     strategy:
       matrix:
-        go-version: [ '1.21', '1.22', '1.23' ]
+        go-version: [ '1.21', '1.22', '1.23', '1.24']
```
yongruilin, line 11: /lgtm
ooc, how to decide we stop supporting the old versions?

Jefftree: Based on go version of k/k still within the support window https://kubernetes.io/releases/

PR #531, opened 2025-03-18
Add validation rules for streaming

serathius:
Borrowed code from https://github.com/kubernetes/enhancements/issues/5116#issuecomment-2731281325 
/assign @liggitt


serathius:
@jpbetz is OOO
/assign @sttts

Issue #533, opened 2025-04-24
go-yaml/yaml as a direct dependency was archived


caozhuozi:
https://github.com/go-yaml/yaml/commit/944c86a7d29391925ed6ac33bee98a0516f1287a

I added this a note in https://github.com/kubernetes/kubernetes/issues/120276
https://github.com/kubernetes/kubernetes/issues/120276#issuecomment-2828028967

dims:
also see https://github.com/kubernetes/kubernetes/pull/131454

PR #534, opened 2025-04-24
Switch away from archived gopkg.in/yaml.v3

dims:
fixes: https://github.com/kubernetes/kube-openapi/issues/533

https://github.com/go-yaml/yaml has a archived notice.

### DIFF
--- a/pkg/validation/spec/gnostic_test.go
+++ b/pkg/validation/spec/gnostic_test.go
@@ -29,10 +29,10 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
-	"gopkg.in/yaml.v3"
 	jsontesting "k8s.io/kube-openapi/pkg/util/jsontesting"
 	. "k8s.io/kube-openapi/pkg/validation/spec"
 	"sigs.k8s.io/randfill"
+	"sigs.k8s.io/yaml/goyaml.v3"
 )
 
 func gnosticCommonTest(t testing.TB, fuzzer *randfill.Filler) {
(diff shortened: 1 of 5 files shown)


pkg/validation/spec/gnostic_test.go
```diff
@@ -294,158 +292,6 @@ func BenchmarkGnosticConversion(b *testing.B) {
...
-				t.Error(err)
-				return
-			} else if len(nodes.Content) != 1 {
-				t.Errorf("unexpected yaml parse result")
-				return
-			}
-
-			root := nodes.Content[0]
-
-			parsed, err := openapi_v2.NewSecurityDefinitionsItem(root, compiler.NewContextWithExtensions("$root", root, nil, nil))
```
dims, line 349: NewSecurityDefinitionsItem's first parameter has to the the `yaml.Node` from `gopkg.in/yaml.v3` so we cannot use this method.

pkg/validation/spec/gnostic_test.go
```diff
@@ -294,158 +292,6 @@ func BenchmarkGnosticConversion(b *testing.B) {
...
-				return
-			} else if len(nodes.Content) != 1 {
-				t.Errorf("unexpected yaml parse result")
-				return
-			}
-
-			root := nodes.Content[0]
-
-			ctx := compiler.NewContextWithExtensions("$root", root, nil, nil)
-			parsed, err := openapi_v2.NewParameter(root, ctx)
```
dims, line 420: `NewContextWithExtensions` and `NewParameter`'s  uses `yaml.Node` from `gopkg.in/yaml.v3` in their method signature so we cannot use this method.

PR #537, opened 2025-05-14
Add OpenAPIModelNamer and opt-in generator support

jpbetz:
This enables kube-openapi to use model names defined by the models instead of the `<go-package>.<go-type-name>` of the Go structs that defines the model.

This is entirely opt-in.  Existing behavior is retained.

`openapi-gen` users may opt-in with the `--use-openapi-model-names`.

Models may opt-in by implementing the `OpenAPIModelNamer` interface introduced by this change:

```go
type OpenAPIModelNamer interface {
	OpenAPIModelName() string
}
```

For example:

```go
func (in Pod) OpenAPIModelName() string {
	return "io.k8s.api.core.v1.Pod"
}
```

When used, the "canonical type names" and "REST friendly names" are the same model name.

This is simpler than the existing behavior where the name formats differ needlessly require conversion:

- Canonical type names
  - e.g. `k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.CustomResourceDefinition`
- REST friendly names
  - e.g. `io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinition`


jpbetz:
/assign @Jefftree 
cc @aojea @BenTheElder 
/sig api-machinery
/triage accepted

pkg/generators/openapi.go
```diff
@@ -339,8 +343,18 @@ func (g openAPITypeWriter) generateCall(t *types.Type) error {
 	// Only generate for struct type and ignore the rest
 	switch t.Kind {
 	case types.Struct:
+		if namer.IsPrivateGoName(t.Name.Name) { // skip private types
```
Jefftree, line 346: Did you encounter one of these types or is this just precautionary?

jpbetz: `int64Amount` was leaking out from quantity into `pkg/generated/openapi/zz_generated.openapi.go`

Jefftree:
/lgtm
/hold

in case you want additional eyes

jpbetz:
/hold cancel
I'd like to merge this part and nudge the k/k/ part again.

PR #538, opened 2025-05-23
bugfix: walk on parameter references

tomleb:
The walker doesn't walk on parameter references (#/parameters/<id>) which may contain definition references (#/definitions/<id>) that we want to keep.

This happens often with DeleteOptions and the DeleteOptions definition ends up being filtered out when looking at a OpenAPI V2 document.


Jefftree:
/lgtm
/approve

thanks!

PR #540, opened 2025-05-26
[WIP] Added quantity format

sreeram-venkitesh:
Part of https://github.com/kubernetes/kubernetes/issues/130639 and along the same lines of https://github.com/kubernetes/kube-openapi/pull/526.

I tested this locally but I'm not sure if we should add `k8s.io/apimachinery` in vendor here.

I have a corresponding PR in k/k which I'm opening along with these changes. Would love to hear some feedback.

### DIFF
--- a/pkg/validation/strfmt/quantity.go
+++ b/pkg/validation/strfmt/quantity.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


cici37:
I think the introduction of "k8s.io/apimachinery/pkg/api/resource" import here would cause loop dependency issue later which we may wanna to avoid.

pkg/validation/strfmt/quantity.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type Quantity string
+
+func init() {
+	quantity := Quantity("")
+	Default.Add("quantity", &quantity, isQuantity)
```
liggitt, line 24: I wonder if this type definition / registration could be done from outside kube-openapi, e.g. defined in kubernetes/kubernetes or apiextensions-apiserver and then registered

If so, that would avoid the kube-openapi → k8s.io/apimachinery dependency

PR #545, opened 2025-06-26
Drop go 1.21 from supported versions

Jefftree:
SInce k/k 1.29 is not supported anymore, I'd like to drop 1.21 and bump the minimum required version to 1.22. This also let's us depend on `go.yaml.in/yaml/v3` with a go.mod version set to 1.22.

/assign @liggitt @dims


dims:
/approve
/lgtm

thanks @Jefftree !

PR #548, opened 2025-07-08
feat(strfmt): add k8s-ip-sloppy string format

lalitc375:
This PR introduces a new string format `k8s-ip-sloppy` for validating IP addresses with slightly more lenient rules, matching the behavior of Kubernetes.

The validation is based on `k8s.io/utils/net.ParseIPSloppy`, which is equivalent to Go's `net.ParseIP` prior to Go 1.17. This version notably allows leading zeros in IPv4 address octets (e.g., "1.2.3.01"), a format that is no longer considered valid by the standard library but is maintained in Kubernetes for backward compatibility.

### DIFF
--- a/pkg/validation/strfmt/kubernetes-extensions.go
+++ b/pkg/validation/strfmt/kubernetes-extensions.go
@@ -17,6 +17,8 @@ package strfmt
 import (
 	"encoding/json"
 	"regexp"
+
+	netutils "k8s.io/utils/net"
 )
 
 const k8sPrefix = "k8s-"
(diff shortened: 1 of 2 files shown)


lalitc375:
Discussed this PR with @jpbetz  and @liggitt offline. k9s:format=k8s-ip (earlier k8s-ip-sloppy) can be represented using existing ipv4 and ipv6  open api formats. Similar to the following.

```yaml
                endpoint:
                  type: string
                  oneOf:
                    - format: ipv4
                    - format: ipv6
```

PR #564, opened 2025-09-10
Compatiblity fix: Always use OpenAPIModelName() functions if they are avalable

jpbetz:
I discovered in k/k that https://github.com/kubernetes/kube-openapi/pull/563 is not fully backward compatible. Before migrating sample-apiserver to use the new model names approach, it's openapi serving was broken.

The problem was once OpenAPIModelName() functions were generated for the core types, sample-apiserver's openapi had a mix of types with OpenAPIModelName() functions (the core types) and without (the sample-apiserver types).

So to be fully compatible, this PR detects OpenAPIModelName() functions and always uses them when available so that projects that import types from core or other projects continue to work even when not using `+k8s:openapi-model-package` for their own types.


Jefftree:
Looks like the CI failure is legitimate. Do the golden files need to be updated?

timuthy:
Just came across this PR, as we're facing exactly this issue when upgrading the Kubernetes dependencies of our aggregated API server to 1.35 (github.com/gardener/gardener).

I’ve used an `openapi-gen` version that includes this fix, but the issue of broken OpenAPI serving still occurs.

- This is the generated openapi file with mixed types: https://github.com/timuthy/gardener/blob/0bcf56e05ccb344b964f62de888aeec2e5d6888f/pkg/apiserver/openapi/openapi_generated.go
- This is how we call the openapi-gen tool: https://github.com/gardener/gardener/blob/662a2e7609c952cb08cce6935181029297d14cfa/hack/update-codegen.sh#L410

@jpbetz do you have any hints or examples where you got this working? I couldn't find this mixed setup in the sample-apiserver project.

PR #578, opened 2026-03-12
Add +k8s:maxBytes and +k8s:minBytes OpenAPI representation

BasavarajBankolli:
### What this PR does

Adds support for the `+k8s:maxBytes` and `+k8s:minBytes` validation markers in kube-openapi.

These markers are mapped to OpenAPI `maxLength` and `minLength` respectively, allowing byte-based validation constraints to be expressed in API definitions.

## Implementation

- Added `MaxBytes` and `MinBytes` fields to `commentTags`
- Added validation for negative values and range conflicts
- Prevented simultaneous use of `maxLength` with `maxBytes` and `minLength` with `minBytes`
- Mapped byte markers to length constraints during schema generation

Fixes: https://github.com/kubernetes/kubernetes/issues/137613

### DIFF
--- a/pkg/generators/markers.go
+++ b/pkg/generators/markers.go
@@ -113,6 +113,8 @@ type commentTags struct {
 	ExclusiveMinimum *bool         `json:"exclusiveMinimum,omitempty"`
 	MaxLength        *int64        `json:"maxLength,omitempty"`
 	MinLength        *int64        `json:"minLength,omitempty"`
+	MaxBytes         *int64        `json:"maxBytes,omitempty"`
+	MinBytes         *int64        `json:"minBytes,omitempty"`
 	Pattern          *string       `json:"pattern,omitempty"`
 	MaxItems         *int64        `json:"maxItems,omitempty"`
 	MinItems         *int64        `json:"minItems,omitempty"`
@@ -187,7 +189,13 @@ func (c *commentTags) ValidationSchema() (*spec.Schema, error) {
 		}
 		transformedAdditionalProperties = &spec.SchemaOrBool{Schema: additionalProperties, Allows: true}
 	}
+	if c.MaxBytes != nil && c.MaxLength == nil {
+		c.MaxLength = c.MaxBytes
+	}
 
+	if c.MinBytes != nil && c.MinLength == nil {
+		c.MinLength = c.MinBytes
+	}
 	res := spec.Schema{
 		SchemaProps: spec.SchemaProps{
 			Nullable:         isNullable,
@@ -249,6 +257,18 @@ func (c commentTags) Validate() error {
 	if c.MaxLength != nil && *c.MaxLength < 0 {
 		err = errors.Join(err, fmt.Errorf("maxLength cannot be negative"))
 	}
+	if c.MinBytes != nil && *c.MinBytes < 0 {
+		err = errors.Join(err, fmt.Errorf("minBytes cannot be negative"))
+	}
+	if c.MaxBytes != nil && *c.MaxBytes < 0 {
+		err = errors.Join(err, fmt.Errorf("maxBytes cannot be negative"))
+	}
+	if c.MaxLength != nil && c.MaxBytes != nil {
+		err = errors.Join(err, fmt.Errorf("maxLength and maxBytes cannot both be specified"))
+	}
+	if c.MinLength != nil && c.MinBytes != nil {
+		err = errors.Join(err, fmt.Errorf("minLength and minBytes cannot both be specified"))
+	}
 	if c.MinItems != nil && *c.MinItems < 0 {
 		err = errors.Join(err, fmt.Errorf("minItems cannot be negative"))
 	}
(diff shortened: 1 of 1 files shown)


BasavarajBankolli:
@jefftree @yujuhong could you please review this PR?

Jefftree:
Sweeping open PRs, is this something we want merged @aaron-prindle @jpbetz?

jpbetz:
I'll defer to @aaron-prindle who is tracking how we map JSONSchema/OpenAPI with k8s: tags.

pkg/generators/markers.go
```diff
@@ -249,6 +257,18 @@ func (c commentTags) Validate() error {
 	if c.MaxLength != nil && *c.MaxLength < 0 {
 		err = errors.Join(err, fmt.Errorf("maxLength cannot be negative"))
 	}
+	if c.MinBytes != nil && *c.MinBytes < 0 {
+		err = errors.Join(err, fmt.Errorf("minBytes cannot be negative"))
+	}
+	if c.MaxBytes != nil && *c.MaxBytes < 0 {
+		err = errors.Join(err, fmt.Errorf("maxBytes cannot be negative"))
+	}
+	if c.MaxLength != nil && c.MaxBytes != nil {
```
aaron-prindle: We can remove this check, maxLength should be allowed on a field with maxBytes

pkg/generators/markers.go
```diff
@@ -249,6 +257,18 @@ func (c commentTags) Validate() error {
...
+		err = errors.Join(err, fmt.Errorf("maxLength and maxBytes cannot both be specified"))
+	}
+	if c.MinLength != nil && c.MinBytes != nil {
```
aaron-prindle, line 257: We can remove this check, minLength should be allowed on a field with minBytes

pkg/generators/markers.go
```diff
@@ -187,7 +189,13 @@ func (c *commentTags) ValidationSchema() (*spec.Schema, error) {
 		}
 		transformedAdditionalProperties = &spec.SchemaOrBool{Schema: additionalProperties, Allows: true}
 	}
+	if c.MaxBytes != nil && c.MaxLength == nil {
```
aaron-prindle: This mapping is semantically incorrect with respect to the k/k implementation of the tgs. OpenAPI `minLength`/`maxLength` count characters not bytes (this was the rational for splitting out `minBytes`/`maxBytes`, see kubernetes/kubernetes#137291). 

I believe we ideally we would emit a CEL rule here using something like`size(bytes(self))` to check the byte count and match the k/k Go validator exactly logic.

pkg/generators/markers.go
aaron-prindle, line 360: I believe we would also want a check like this for minBytes and maxBytes as currently the tag only targets strings

aaron-prindle:
Currently only maxBytes is implemented upstream in k/k, this PR would have to wait until minBytes is also implemented

aaron-prindle:
Can you also add test cases for this in `markers_test` similar to [min|max]Length in https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/markers_test.go

aaron-prindle:
Thanks for the PR here @BasavarajBankolli. Currently the PR here opts to map `maxBytes`/`minBytes` directly onto OpenAPI's `maxLength`/`minLength`.  This option makes k/k native type validation and OpenAPI-side validation disagree (OpenAPI counts characters; the k/k `MaxBytes` validator counts bytes).

The initial issue here: https://github.com/kubernetes/kubernetes/issues/137613 left the OpenAPI representation as an explicit open design question ("new x-kubernetes tag? CEL rule?" which has not gotten consensus as you noted here: https://github.com/kubernetes/kubernetes/issues/137613#issuecomment-4041495233). We likely want to get consensus on this before implementation here.  I've added some comments here assuming the CEL method but realistically we want to get some sig-apimachinery folks to confirm on the issue there the approach before implementing.

BasavarajBankolli:
Thanks for the detailed feedback!

Updated the PR to:

* remove the direct maxBytes/minBytes -> maxLength/minLength mapping
* remove the maxLength/maxBytes conflict checks
* add string-type validation checks for maxBytes
* reduce scope to maxBytes only since minBytes is not yet implemented upstream in k/k
* add marker tests similar to maxLength/minLength

pkg/generators/markers.go
```diff
@@ -187,7 +189,13 @@ func (c *commentTags) ValidationSchema() (*spec.Schema, error) {
...
+		c.MaxLength = c.MaxBytes
```
liggitt: maxBytes != maxLength (characters) ... I don't think we can conflate them like this

pkg/generators/markers.go
```diff
@@ -187,7 +189,13 @@ func (c *commentTags) ValidationSchema() (*spec.Schema, error) {
...
+	}
 
+	if c.MinBytes != nil && c.MinLength == nil {
+		c.MinLength = c.MinBytes
```
liggitt: minBytes != minLength (characters) ... I don't think we can conflate them like this

PR #590, opened 2026-04-26
drop mailru/easyjson by upgrading go-openapi/swag to v0.25.4

dims:
go-openapi/swag v0.25.4 refactored into sub-modules that no longer depend on github.com/mailru/easyjson. The root package re-exports all the same symbols so no import changes are needed.

The go directive bumps to 1.24.0 (required by swag sub-modules), which also surfaces two pre-existing vet errors (non-constant format strings) fixed here.


jpbetz:
/approve
/lgtm

Excellent. Thanks!

dims:
looks like we need explicit approvals from

/assign @sttts @liggitt 

(thanks @jpbetz!)

sttts:
@dims how deep have you looked into potential behaviour changes? go-openapi has some reputation of changing semantics without notification by accident. As we use it in our CRD validation there is risk that we break our API (we have forked the go-openapi validator nowadays, so risk is lower than in the past, but worth a check anyway).

dims:
@sttts poked at it with some agents, the main thing that showed up was a change in `IsFloat64AJSONInteger`. which didn't seem too bad to me.

| SHA | Message | Author |
|---|---|---|
| [`9b27996`](https://github.com/go-openapi/swag/commit/9b27996645668e213c4cc21717241fe76aa2c12f) | improve equality comparison for floats | @afansv |
| [`df12219`](https://github.com/go-openapi/swag/commit/df12219bdbf52f470098b14bab30e1ac13a7b212) | IsFloat64JSONInteger check: fixed short-circuit case; benchmark confirms ~14% faster | @fredbi |
| [`24f77a7`](https://github.com/go-openapi/swag/commit/24f77a7beca3ac6d87f2d9a207d45d1689cfb947) | experimental code to compare rounding methods | @fredbi |

The core algorithm change (`uint64` truncation → `math.Round`) is in `9b27996`. The short-circuit fix and benchmark validation are in `df12219`.

See https://gist.github.com/dims/ecd681ea7d4748300d3b6203074e7d70 for the full gist.

Jefftree:
@dims need a rebase on this one since the SHA pinning already merged: https://github.com/kubernetes/kube-openapi/pull/592

dims:
@Jefftree Done!

Jefftree:
/lgtm

pkg/validation will require either @sttts @liggitt

PR #591, opened 2026-04-27
Add apiversion.yaml support

jpbetz:
This implements a declaration file for simplifying codegen management as described in https://github.com/kubernetes/enhancements/issues/5975

This improves openapi-gen to support groupversion.yaml files but still respect +k8s:openapi-gen if not yaml file is present.


jpbetz:
Looks like this is blocked until https://github.com/kubernetes/kube-openapi/pull/592 merges

jpbetz:
I'm becoming less convinced this approach will work. Closing this will I work on the k8s side.

PR #593, opened 2026-04-29
Add API declaration support

jpbetz:
This is needed by https://github.com/kubernetes/kubernetes/pull/138657


jpbetz:
@Jefftree would you review?  This is needed by https://github.com/kubernetes/kubernetes/pull/138657 and we're happy with the general approach. If I can merge this I can pick it up in k/k and then integrate.

pkg/generators/model_names.go
```diff
@@ -40,6 +41,21 @@ func extractOpenAPISchemaNamePackage(comments []string) (string, error) {
 	return v.Value, nil
 }
 
+// resolvePackageModelPackage returns the OpenAPI model package for pkg.
+// When apiversion.yaml is present, spec.modelPackage is authoritative.
```
Jefftree, line 45: I'm worried it's possible to have the existing comments still be around and people unsure about the source of truth. Would it be good practice to error if both are set to ensure migration?

pkg/generators/openapi.go
```diff
@@ -321,8 +340,13 @@ func (g openAPITypeWriter) shouldUseOpenAPIModelName(t *types.Type) bool {
 	if value != "" {
 		return true
 	}
+	// Private types never get a generated OpenAPIModelName method
+	// (model_names-gen skips them), so we can't reference one here.
+	if namer.IsPrivateGoName(t.Name.Name) {
```
Jefftree: Seems like a separate bug, do we need it for this PR?

pkg/generators/model_names.go
jpbetz: I'm willing to try out the strict approach and see how it goes. Updated.

pkg/generators/openapi.go
jpbetz: Removed.

pkg/generators/apidefinitions/loader_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestLoadAPIVersion(t *testing.T) {
+	const manifest = `apiVersion: apidefinitions.k8s.io/v1alpha1
+kind: APIVersion
+metadata:
+  name: apps/v1
```
Jefftree: do we want the gv to be embedded in the name? is corev1 going to be `/v1` as the name? 

I noticed in the KEP we also want to support internal versions. Wondering if that gives too much ambiguity in our parser `splitGroupVersion` to have to handle all those cases in a single string.

pkg/generators/model_names.go
```diff
@@ -40,6 +41,28 @@ func extractOpenAPISchemaNamePackage(comments []string) (string, error) {
...
+	}
+	tagPackage, err := extractOpenAPISchemaNamePackage(pkg.Comments)
+	if err != nil {
+		return "", err
+	}
+	if apiVersion != nil {
+		if len(tagPackage) > 0 {
+			return "", fmt.Errorf("Expected to either a apiversion.yaml or k8s:openapi-model-package tag, but found both. Please remove one.")
+		}
+		return apiVersion.Spec.ModelPackage, nil
```
Jefftree, line 61: Do we need validation at all? This is more of a question, do we need to check for things like certain fields being present?

pkg/generators/apidefinitions/loader.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, fmt.Errorf("%s: %w", filepath.Join(dir, apiVersionFile), err)
+	}
+	if err := validateName(av); err != nil {
+		return nil, fmt.Errorf("%s: %w", filepath.Join(dir, apiVersionFile), err)
+	}
+	return av, nil
+}
+
+func readManifest(dir, filename string) ([]byte, error) {
+	data, err := os.ReadFile(filepath.Join(dir, filename))
```
Jefftree, line 62: This is a read + unmarshal on every type right? Should we cache it?

pkg/generators/apidefinitions/loader.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("expected kind %s but got %s", expectedKind, actualKind)
+	}
+	return nil
+}
+
+func validateName(av *APIVersion) error {
+	if av.Metadata.Name == "" {
+		return fmt.Errorf("metadata.name is required")
+	}
+	if _, _, err := splitGroupVersion(av.Metadata.Name); err != nil {
```
Jefftree: This seems to just enforce non-empty. I think we should be more strict here.

pkg/generators/model_names.go
```diff
@@ -40,6 +41,28 @@ func extractOpenAPISchemaNamePackage(comments []string) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	tagPackage, err := extractOpenAPISchemaNamePackage(pkg.Comments)
+	if err != nil {
+		return "", err
+	}
+	if apiVersion != nil {
+		if len(tagPackage) > 0 {
+			return "", fmt.Errorf("Expected to either a apiversion.yaml or k8s:openapi-model-package tag, but found both. Please remove one.")
```
Jefftree: Can we pass in the package here so it's more actionable for the user?

Also `s/Expected to either/expected either`

pkg/generators/apidefinitions/loader_test.go
jpbetz: This give us names that match how we think of APIVersions elsewhere.  It's a really well established convention.

Consider:

```
apiVersion: v1
kind: Pod
```

and

```
apiVersion: apps/v1
kind: Deployment
```

Here, we're defining those apiVersions, and so we use the names users are already familiar with.

Internal versions are not handled by this kind. This is for versioned types.

pkg/generators/model_names.go
jpbetz: It can only help, I think.

pkg/generators/apidefinitions/loader.go
jpbetz: If it proves beneficial, we can optimize.  Codgen already parses so much data this is really just a drop in the bucket.

pkg/generators/apidefinitions/loader.go
jpbetz: This is tricky.  The code this replaces didn't have any validation/constraints.

I can enforce a reasonable k8s style resource name, but 3rd party systems don't always use k8s style version identifiers, so if we impose restrictions on version we could limit usability of this to mostly in-tree applications.

jpbetz:
@Jefftree Feedback applied. Thanks!

pkg/generators/model_names.go
```diff
@@ -40,6 +41,28 @@ func extractOpenAPISchemaNamePackage(comments []string) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	tagPackage, err := extractOpenAPISchemaNamePackage(pkg.Comments)
+	if err != nil {
+		return "", err
+	}
+	if apiVersion != nil {
+		if len(tagPackage) > 0 {
+			return "", fmt.Errorf("in package %s: expected to either a apiversion.yaml or k8s:openapi-model-package tag, but found both. Please remove one.", pkg.Dir)
```
Jefftree:
```suggestion
			return "", fmt.Errorf("in package %s: expected either a apiversion.yaml or k8s:openapi-model-package tag, but found both. Please remove one.", pkg.Dir)
```

pkg/generators/apidefinitions/loader.go
Jefftree: sounds good.

pkg/generators/apidefinitions/loader.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// LoadAPIVersion reads an apiversion.yaml file, returning nil if absent.
+func LoadAPIVersion(dir string) (*APIVersion, error) {
+	data, err := readManifest(dir, apiVersionFile)
+	if err != nil || data == nil {
+		return nil, err
+	}
+	av := &APIVersion{}
+	if err := yaml.Unmarshal(data, av); err != nil {
```
Jefftree, line 49: I think `UnmarshalStrict` would make more sense here since we don't want to ignore unknown fields (which most likely comes from typo). wdyt?

pkg/generators/apidefinitions/loader.go
Jefftree: I was just thinking basic constraints like "this must contain at most one slash (/)". Agree that making it too restrictive will limit usability, so I'm fine with adding it later if we find people making the same mistakes.

pkg/generators/apidefinitions/loader_test.go
Jefftree: For apiversion 100% agree. Kubernetes [names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) usually follow some variant of RFC 1123 that disallows slashes in names though. This isn't a persisted type so I guess it doesn't matter that much?

Jefftree (commented):
a few nits, otherwise LGTM.

pkg/generators/apidefinitions/loader_test.go
jpbetz: This occurred to me as well. It's worth discussing.

We certainly are making a trade-off between consistency with API naming conventions and consistency of how an identifier for a well understood concept is used.

I bias towards consistency with the identifier because it will be easily grep-able/grok-able and naturally aligns with what the actual `apiVersion` field. The most obvious alternative would be to use CRD style names.  But CRD only have structure for GV (not GVR), so it's no directly applicable.

pkg/generators/apidefinitions/loader.go
jpbetz: I'd like to allow for version skew between the kube-openapi version and the version of k8s.

We might add more fields to this in the future and use them in k8s. I'd like older versions of kube-openapi to continue to work.

This also reminds me, I need to make sure kubebuilder is not impacted by this change..  I don't think kubebuilder process any of the tags we'll be able to drop, but I do want to check.

pkg/generators/apidefinitions/loader.go
jpbetz: Good catch. I've rewritten the name validation to be strict about either being <group>/<version> or <version> with non-empty values for both.

PR #596, opened 2026-05-02
Move modelPackage to group level, provide opt-out

jpbetz:
After integrating with k/k it's clear that group level modelPackage declarations are much cleaner.

Also, all other generators provide opt-out support, so adding it for consistency.


pkg/generators/openapi.go
```diff
@@ -127,14 +127,19 @@ func isOpenAPIEnabledForPackage(pkg *types.Package) bool {
 	if pkg == nil {
 		return false
 	}
+	hasFalse := hasOpenAPITagValue(pkg.Comments, tagValueFalse)
+	hasTrue := hasOpenAPITagValue(pkg.Comments, tagValueTrue)
+	if hasFalse && !hasTrue { // For backward compatability
```
Jefftree, line 132:
```suggestion
	if hasFalse && !hasTrue { // For backward compatibility
```

test/integration/testdata/apidefinitions/v1/apiversion.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: apidefinitions.k8s.io/v1alpha1
+kind: APIVersion
+metadata:
+  name: apidefinitions.test.k8s.io/v1
```
Jefftree, line 4: Is the only purpose of this file now to provide the name? Assuming we'll add to it in the future but it seems like we can get this information from directory name/go package name.

pkg/generators/openapi.go
Jefftree: nit: update doc for apigroup.yaml.

pkg/generators/apidefinitions/types.go
```diff
@@ -32,8 +32,35 @@ type APIVersion struct {
...
+	Spec       APIGroupSpec `json:"spec"`
+}
+
+// APIGroupSpec provides a specification of an APIGroup.
+type APIGroupSpec struct {
+	// ModelPackage is the OpenAPI model package prefix shared by every
+	// version in this group. The per-version model package is
+	// "<ModelPackage>.<version>" (e.g. "io.k8s.api.apps" + "v1" yields
+	// "io.k8s.api.apps.v1").
+	ModelPackage string `json:"modelPackage"`
```
Jefftree: Noticed the omitempty was dropped, don't we still allow it to be empty?

test/integration/testdata/apidefinitions/v1/apiversion.yaml
jpbetz: No, the main purpose is to drive all codegen off of a declaration file instead of a bunch of random tags in doc.go files. OpenAPI is just really simple.

pkg/generators/apidefinitions/types.go
jpbetz: Good catch

pkg/generators/openapi.go
Jefftree: wonder if it's github's issue of not applying the comment/suggestion?

pkg/generators/openapi.go
```diff
@@ -120,21 +120,26 @@ func apiTypeFilterFunc(c *generator.Context, t *types.Type) bool {
 }
 
 // isOpenAPIEnabledForPackage reports whether openapi generation is
-// requested for pkg. apiversion.yaml is authoritative when present;
-// the legacy +k8s:openapi-gen=true tag is consulted only when the yaml
-// is absent.
+// requested for pkg. apigroup.yaml and apiversion.yaml are authoritative when present.
+// The +k8s:openapi-gen=true tag is checked only when the API defintion files
```
Jefftree:
```suggestion
// The +k8s:openapi-gen=true tag is checked only when the API definition files
```

pkg/generators/model_names.go
```diff
@@ -42,24 +43,48 @@ func extractOpenAPISchemaNamePackage(comments []string) (string, error) {
...
-		return apiVersion.Spec.ModelPackage, nil
+		return "", nil
+	}
+	derived := apiGroup.ModelPackageFor(apiVersion.VersionFromName())
+	if tagPackage != "" && tagPackage != derived {
+		return "", fmt.Errorf("in package %s: parent apigroup.yaml derives model package %q for version %q but +k8s:openapi-model-package tag is %q. Please make them match or remove the tag.",
+			pkg.Dir, derived, apiVersion.VersionFromName(), tagPackage)
 	}
-	return tagPackage, nil
+	return derived, nil
```
Jefftree, line 87: walking through this, isn't it possible that derived is empty?

Jefftree:
One edge case question, otherwise LGTM

https://github.com/kubernetes/kube-openapi/pull/596#discussion_r3182593739

pkg/generators/model_names.go
jpbetz: It is. ModelPackageFor handles this and the empty return case is documented.

Jefftree:
/lgtm
/approve
/hold

Feel free to unhold when you want to merge.

jpbetz:
/hold cancel
I've verified this works as needed with downstream changes, pulling it in now

PR #599, opened 2026-05-05
Bump kin-openapi to v0.137.0 and fix schema default validation

arthurbdiniz:
### Motivation

The Debian package for [kin-openapi](https://tracker.debian.org/pkg/golang-github-getkin-kin-openapi) is at a version v0.135, making it incompatible with kube-openapi's current dependency on v0.103.0. This bump brings the integration test dependency to v0.137.0.

### What broke

Starting from versions beyond v0.103.0, kin-openapi's `Validate()` method enforces that default values in OpenAPI v3 schemas conform to their own schema constraints. This is correct per the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md), a default value must be valid against the schema it belongs to.

> default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value **MUST** conform to the defined type for the Schema Object defined at the same level.

The integration test's `builder3/main.go` calls `specForValidator.Validate(loader.Context)` on the generated spec, which now rejects schemas where:

1. A struct's enforced default is `{}` but the schema declares required properties (e.g. `defaults.SubStruct` requires `"S"`, so `{}` is invalid)
2. A field's enforced default (zero value) violates its own validation constraints (e.g. `NumberValue float64` defaults to `0` but has `exclusiveMinimum: 1`)

### The fix

**Generator change (`mustEnforceDefault`):** Previously, non-pointer struct fields always got `map[string]interface{}{}` as their enforced default. This empty object is invalid when the struct has required fields. The new `structZeroDefault` function builds a default that includes zero values for all required (non-optional) members, producing e.g. `{"S": ""}` instead of `{}`. This makes the generated default compliant with the schema's own `required` constraints.

**Test data fixes:**
- `defaults/default.go`: Fixed `+default={"s": "foo", "i": 5}` to `{"S": "foo", "I": 5}`, the JSON keys must match the actual serialized property names (`S` and `I` uppercase), which the stricter validation now checks.
- `valuevalidation/alpha.go`: Made `StringValue` and `NumberValue` optional. Their validation constraints (`minLength: 1`, `exclusiveMinimum: 1`) are incompatible with their zero values (`""`, `0`). A non-optional field whose zero value violates its own constraints is a schema that contradicts itself. Making them optional avoids the enforced default while still testing the validation marker generation.


arthurbdiniz:
This version of kin-openapi requires `go >= 1.25` otherwise CI will fail.

```
go: go.mod requires go >= 1.25 (running go 1.24.13; GOTOOLCHAIN=local)
```

Should we remove version 1.24 from CI?

Jefftree:
1.24 cannot be removed as we need to support the last 3 kubernetes release (back to 1.34 which is still on 1.24). Can we bump to a version that supports 1.24?

arthurbdiniz:
@Jefftree I was able to make it work using version v0.135.0, moved to another pull request to have a correct branch name.

https://github.com/kubernetes/kube-openapi/pull/600

PR #600, opened 2026-05-05
Bump kin-openapi to v0.135.0 and fix schema default validation

arthurbdiniz:
### Motivation

The Debian package for [kin-openapi](https://tracker.debian.org/pkg/golang-github-getkin-kin-openapi) is at a version v0.135, making it incompatible with kube-openapi's current dependency on v0.103.0. This bump brings the integration test dependency to v0.135.0.

### What broke

Starting from versions beyond v0.103.0, kin-openapi's `Validate()` method enforces that default values in OpenAPI v3 schemas conform to their own schema constraints. This is correct per the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md), a default value must be valid against the schema it belongs to.

> default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value **MUST** conform to the defined type for the Schema Object defined at the same level.

The integration test's `builder3/main.go` calls `specForValidator.Validate(loader.Context)` on the generated spec, which now rejects schemas where:

1. A struct's enforced default is `{}` but the schema declares required properties (e.g. `defaults.SubStruct` requires `"S"`, so `{}` is invalid)
2. A field's enforced default (zero value) violates its own validation constraints (e.g. `NumberValue float64` defaults to `0` but has `exclusiveMinimum: 1`)

### The fix

**Generator change (`mustEnforceDefault`):** Previously, non-pointer struct fields always got `map[string]interface{}{}` as their enforced default. This empty object is invalid when the struct has required fields. The new `structZeroDefault` function builds a default that includes zero values for all required (non-optional) members, producing e.g. `{"S": ""}` instead of `{}`. This makes the generated default compliant with the schema's own `required` constraints.


Jefftree:
Seems like we have test data bugs, generator bugs, and the kin-openapi bump. It's a bit hard to review as is, can we split it into multiple PRs?

arthurbdiniz:
Sure @Jefftree, I've split into small PRs.

- https://github.com/kubernetes/kube-openapi/pull/601
- https://github.com/kubernetes/kube-openapi/pull/602
- https://github.com/kubernetes/kube-openapi/pull/603

After we get those merged I will rebase this one and check if I can split more.

Jefftree:
> Sure @Jefftree, I've split into small PRs.
> 
> * [Fix default annotation to match JSON property names #601](https://github.com/kubernetes/kube-openapi/pull/601)
> * [Fix namedmodels --output-pkg path in integration test README #602](https://github.com/kubernetes/kube-openapi/pull/602)
> * [test/integration: make valuevalidation fields with incompatible const… #603](https://github.com/kubernetes/kube-openapi/pull/603)
> 
> After we get those merged I will rebase this one and check if I can split more.

Approved all 3, should be merged soon. Thank you!

arthurbdiniz:
Thanks for the quick response, @Jefftree!

 I've rebased this PR now.
 
 This one is the most complex to review since includes new logic to the generator. Let me know if you have any questions.

Jefftree:
will do a pass tomorrow, thanks again for the PR!

Jefftree:
Tested out k/k OpenAPI changes here: https://github.com/kubernetes/kubernetes/pull/138895/changes/c59e94e03b0f302f7b671168b15aeed829393078, seems reasonable.

This is somewhat breaking as the defaults will be tighter, but makes the schema more accurate with `structZeroDefault`. Highly doubt that any consumer directly uses/benefits from this though, but I figure it's still a net positive change since we keep dependencies up to date and make openapi more accurate. (go.mod changes/additions are in tests only, so k/k vendor isn't affected).

/cc @liggitt @jpbetz

test/integration/go.mod
```diff
@@ -33,8 +33,14 @@ require (
 	github.com/google/gnostic-models v0.7.0 // indirect
 	github.com/google/go-cmp v0.7.0 // indirect
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
-	github.com/invopop/yaml v0.1.0 // indirect
+	github.com/josharian/intern v1.0.0 // indirect
+	github.com/mailru/easyjson v0.7.7 // indirect
```
liggitt, line 37: github.com/mailru/easyjson is being added back in? we're actively trying to eliminate this from upstream k/k and had *just* gotten down to only kustomize referencing this

Jefftree: go mod addition is test only. Not pulled by k/k.

test/integration/testdata/golden.v2.json
```diff
@@ -987,7 +989,10 @@
      "Field": {
       "type": "array",
       "items": {
-       "default": {},
+       "default": {
```
liggitt: this is pre-existing, but we shouldn't be emitting a default for an item inside a slice or a value type in a map ... even if field `Foo []Bar` or `Foo map[string]Bar` is required, that doesn't mean we emit a default for the Bar schema inline.

It looks like the vast majority of the diffs in the k/k commit are of this flavor ... if we fixed this, the impact on Kubernetes core openapi would be much much smaller.

I think generateSliceProperty and generateMapProperty are incorrect in emitting defaults for their item / value types

test/integration/go.mod
liggitt: hmm, ok

test/integration/testdata/golden.v2.json
arthurbdiniz: Hi @liggitt,

If I understood correctly, the issue is that `generateSliceProperty` and `generateMapProperty` are calling `generateDefault` on their element/value types, which emits a `default` on the items/additionalProperties schema.

Per JSON Schema, `default` means `"the value to assume if this is missing"`, which makes sense for object properties but not for array items or map values, since those are always explicitly provided.

- https://github.com/kubernetes/kube-openapi/blob/19c3011e7fa0b84457787ce60d03b5d7d78c6521/pkg/generators/openapi.go#L1151
- https://github.com/kubernetes/kube-openapi/blob/19c3011e7fa0b84457787ce60d03b5d7d78c6521/pkg/generators/openapi.go#L1188

So the fix would be to remove the `generateDefault` calls from those two functions? Keeping defaults only at the field/property level.

Here's the patch I have, let me know if this is what you had in mind and if I should go ahead and update in this PR:

```diff
diff --git a/pkg/generators/openapi.go b/pkg/generators/openapi.go
index 7a3e3bb..fa5ee8f 100644
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -1202,9 +1202,6 @@ func (g openAPITypeWriter) generateMapProperty(t *types.Type) error {
 
        g.Do("Type: []string{\"object\"},\n", nil)
        g.Do("AdditionalProperties: &spec.SchemaOrBool{\nAllows: true,\nSchema: &spec.Schema{\nSchemaProps: spec.SchemaProps{\n", nil)
-       if err := g.generateDefault(t.Elem.CommentLines, t.Elem, false, t.Elem); err != nil {
-               return err
-       }
        typeString, format := openapi.OpenAPITypeFormat(elemType.String())
        if typeString != "" {
                g.generateSimpleProperty(typeString, format)
@@ -1239,9 +1236,6 @@ func (g openAPITypeWriter) generateSliceProperty(t *types.Type) error {
        elemType := resolveAliasAndPtrType(t.Elem)
        g.Do("Type: []string{\"array\"},\n", nil)
        g.Do("Items: &spec.SchemaOrArray{\nSchema: &spec.Schema{\nSchemaProps: spec.SchemaProps{\n", nil)
-       if err := g.generateDefault(t.Elem.CommentLines, t.Elem, false, t.Elem); err != nil {
-               return err
-       }
        typeString, format := openapi.OpenAPITypeFormat(elemType.String())
        if typeString != "" {
                g.generateSimpleProperty(typeString, format)
```

Generation example

`test/integration/testdata/golden.v2.json`
```diff
"Field": {
       "type": "array",
       "items": {
-       "default": {
-        "Port": 0,
-        "Protocol": ""
-       },
        "$ref": "#/definitions/listtype.Item"
       },
```

liggitt: Yeah, I think that's the right patch. Let's merge that in a separate PR and ripple that into a Kubernetes update first, then rebase this PR on the result. That will make the generated changes from this PR much smaller and easier to review.

liggitt: @Jefftree might be able to help with the Kubernetes dependency update if needed

arthurbdiniz: PR open https://github.com/kubernetes/kube-openapi/pull/606, thanks for the feedback.

arthurbdiniz:
Rebased to get https://github.com/kubernetes/kube-openapi/pull/606 changes.

arthurbdiniz:
Updated k/k https://github.com/kubernetes/kubernetes/pull/138932 to check how repo will change with this PR.

liggitt:
looks like this needs a rebase, and can update https://github.com/kubernetes/kubernetes/pull/138932 as well after https://github.com/kubernetes/kubernetes/pull/139001 merges (the diff there will shrink a lot after https://github.com/kubernetes/kubernetes/pull/139001 merges)

pkg/generators/openapi.go
```diff
@@ -967,6 +967,60 @@ func mustEnforceDefault(t *types.Type, omitEmpty bool) (interface{}, error) {
...
+		}
+		optional, err := isOptional(&m)
+		if err != nil {
+			return nil, err
+		}
+		if optional {
+			continue
+		}
+		// Include zero value for this required field
+		memberOmitEmpty := strings.Contains(reflect.StructTag(m.Tags).Get("json"), "omitempty")
```
liggitt: this is a copy of pre-existing checks, but this is not correct... this would false-positive on a field named `json:"omitemptyTest"`

let's fix the existing checks and use the fixed check here:
```diff
@@ -101,7 +101,7 @@ func isOptional(m *types.Member) (bool, error) {

        // If neither +optional nor +required is present in the comments,
        // infer optional from the json tags.
-       return strings.Contains(reflect.StructTag(m.Tags).Get("json"), "omitempty"), nil
+       return hasOmitemptyTag(m), nil
 }

 func apiTypeFilterFunc(c *generator.Context, t *types.Type) bool {
@@ -246,6 +246,11 @@ func getReferableName(m *types.Member) string {
        }
 }

+func hasOmitemptyTag(m *types.Member) bool {
+       jsonTag, _ := reflect.StructTag(m.Tags).Lookup("json")
+       return strings.HasSuffix(jsonTag, ",omitempty") || strings.Contains(jsonTag, ",omitempty,")
+}
+
 func shouldInlineMembers(m *types.Member) bool {
        jsonTag, jsonTagExists := reflect.StructTag(m.Tags).Lookup("json")
        return m.Embedded && jsonTagExists && (jsonTag == "" || strings.HasPrefix(jsonTag, ","))
@@ -1056,7 +1061,7 @@ func (g openAPITypeWriter) generateProperty(m *types.Member, parent *types.Type)
                g.Do("},\n},\n", nil)
                return nil
        }
-       omitEmpty := strings.Contains(reflect.StructTag(m.Tags).Get("json"), "omitempty")
+       omitEmpty := hasOmitemptyTag(m)
        if err := g.generateDefault(m.CommentLines, m.Type, omitEmpty, parent); err != nil {
                return fmt.Errorf("failed to generate default in %v: %v: %v", parent, m.Name, err)
        }
```

arthurbdiniz: I introduced a `hasOmitemptyTag` helper that checks for `,omitempty` as a proper comma-delimited option (suffix or mid-position) instead of a naive substring match. This correctly rejects field names like `"omitemptyTest"` or `"fooOmitempty"`.

I also applied the same fix to the third occurrence in `structZeroDefault` which had the same bug, and added a unit test covering both the valid cases and the false-positive scenarios.

https://github.com/kubernetes/kube-openapi/pull/600/changes/d0477d7c55d525de301e4da55b6867ff5eee08d4

Code also rebased. :)

arthurbdiniz: @liggitt let me know if make sense to raise https://github.com/kubernetes/kube-openapi/commit/d0477d7c55d525de301e4da55b6867ff5eee08d4 into another PR just with the hasOmitemptyTag helper and the fix to `isOptional`.

liggitt: > @liggitt let me know if make sense to raise [d0477d7](https://github.com/kubernetes/kube-openapi/commit/d0477d7c55d525de301e4da55b6867ff5eee08d4) into another PR just with the hasOmitemptyTag helper and the fix to `isOptional`.

if you want to and it is simpler, that's fine

arthurbdiniz: I think it will make this PR cleaner.

Here it is the separate PR https://github.com/kubernetes/kube-openapi/pull/610.

arthurbdiniz:
Rebased to get https://github.com/kubernetes/kube-openapi/pull/610.

liggitt:
this lgtm

https://github.com/kubernetes/kubernetes/pull/139001 merged, so if you want to update https://github.com/kubernetes/kubernetes/pull/138932 we can do a last sanity check of the impact on kubernetes/kubernetes, then merge this

arthurbdiniz:
> this lgtm
> 
> [kubernetes/kubernetes#139001](https://github.com/kubernetes/kubernetes/pull/139001) merged, so if you want to update [kubernetes/kubernetes#138932](https://github.com/kubernetes/kubernetes/pull/138932) we can do a last sanity check of the impact on kubernetes/kubernetes, then merge this

@liggitt, Rebased and updated this and k/k https://github.com/kubernetes/kubernetes/pull/138932

arthurbdiniz:
@Jefftree, anything else needed for this PR? If not, good for another review please.

Jefftree:
/approve

Leaving lgtm with liggitt

liggitt:
sweeping through the k/k PR, I agree the new generated defaults include the required fields for the types, but most of the defaults would not actually pass API validation and be usable (this was true before as well with the empty `default: {}` declarations, it's just even more confusing now since the default *appears* to be a more usable object with some fields included).

Is there a way (in kubernetes/kubernetes API types) to indicate a field should *omit* the default since the serialized zero value is not valid?

arthurbdiniz:
@liggitt, it's possible to add a `+nodefault` comment marker on the field.

When present, the generator skips the enforced default entirely,  no `default` field is emitted in the schema output.

So in k/k API types it would look like:

```go
// +nodefault
Spec SomeSpec `json:"spec"`
```

The generated schema would have the field in `required` but with no `default` key at all. Since kin-openapi only validates defaults that are present, this passes validation cleanly.

I confirmed it against v0.135.0 with the full integration suite.

The tradeoff is that we lose the documentation of `"the deserializer produces the zero value when this field is absent,"` but for fields where that zero value is never valid anyway, that information is arguably more confusing than helpful (as you mentioned).

Implementation it's a small check at the top of `generateDefault`:

```go
if gengo.ExtractCommentTags("+", comments)[tagNoDefault] != nil {
    return nil
}
```

Is this the kind of thing you had in mind, or were you thinking of something different?

liggitt:
Yeah, something like along those lines. Does that exist today or is that a speculative description of something we could add?

arthurbdiniz:
It's something I prototyped locally on top of this PR. I have a working implementation with unit tests and integration tests passing.

Happy to open a separate PR for it if you think it's worth adding. That way this PR can land as-is (bumping kin-openapi + structZeroDefault), and the +nodefault marker would be a follow-up that k/k types can adopt where the zero-value default is misleading.

Or if you'd prefer it bundled here, I can add it to this branch. Whatever works best.

```diff
diff --git a/pkg/generators/openapi.go b/pkg/generators/openapi.go
index 8338459..12d0f08 100644
--- a/pkg/generators/openapi.go
+++ b/pkg/generators/openapi.go
@@ -43,6 +43,7 @@ const markerPrefix = "+k8s:validation:"
 const tagOptional = "optional"
 const tagRequired = "required"
 const tagDefault = "default"
+const tagNoDefault = "nodefault"
 
 // Known values for the tag.
 const (
@@ -1003,6 +1004,10 @@ func structZeroDefault(t *types.Type) (interface{}, error) {
 }
 
 func (g openAPITypeWriter) generateDefault(comments []string, t *types.Type, omitEmpty bool, commentOwningType *types.Type) error {
+       // If +nodefault is specified, skip the default entirely.
+       if gengo.ExtractCommentTags("+", comments)[tagNoDefault] != nil {
+               return nil
+       }
        def, ref, err := defaultFromComments(comments, commentOwningType.Name.Package, t)
        if err != nil {
                return err
 ```

liggitt:
I'd think it is good to to be able to opt out of misleading zero-value defaults where those wouldn't actually be accepted by the server. Doing that in a separate PR, so that when we pulling this PR into k/k we can choose whether to accept the populated zero-value default or opt out sounds good good to me. @jpbetz @Jefftree, any thoughts on the idea in https://github.com/kubernetes/kube-openapi/pull/600#issuecomment-4492205343

liggitt:
(thanks for your patience with all the prereq / coreq things this PR is revealing we already had issues with ... unfortunately, dev in this space is sometimes like this)

Jefftree:
> I'd think it is good to to be able to opt out of misleading zero-value defaults where those wouldn't actually be accepted by the server. Doing that in a separate PR, so that when we pulling this PR into k/k we can choose whether to accept the populated zero-value default or opt out sounds good good to me. @jpbetz @Jefftree, any thoughts on the idea in [(earlier comment)](https://github.com/kubernetes/kube-openapi/pull/600#issuecomment-4492205343)

+1 to opt out of misleading zero value defaults, although I think relying on the developers/type authors to tag these on their types might be a bit excessive. We automatically generate defaults for structs that are not pointers, and in almost all cases generating an empty struct as a default is wrong and provides no real value.

How valuable would it be to have the option to skip defaulting for struct objects and only populate defaults at the leaf nodes? Something like `--skip-object-defaults` that drops all struct level defaulting (non leaf node). I haven't fully validated the implications of this, but seeing amount of `default: {}` suggests that if we go with `+nodefault`, we'll end up tagging a significant portion of the structs and add additional burden on API review and maintenance. In cases where having a default on a substruct does produce a valid kubernetes object, does the "the deserializer produces the zero value when this field is absent" provide additional value?

jpbetz:
Generating defaults into OpenAPI that match exactly what will be read back after the apiserver defaults the field makes sense to me.

I'm fine with `+nodefault` as a mechanism to tune it.

I tried out this against k/k and it myself (https://github.com/kubernetes/kubernetes/compare/master...jpbetz:kubernetes:test-bump-kin-openapi) before noticing https://github.com/kubernetes/kubernetes/pull/138932 in the comments.  I haven't had time to go through all the defaults and ensure they are correct so +1 to keeping the defaults in a separate PR so we can review the k/k default changes separate from other changes.

arthurbdiniz:
Thanks for the feedback everyone.

Sounds like we're aligned on:

- Land this PR as-is (kin-openapi bump + `structZeroDefault`)
- `+nodefault` (and possibly `--skip-object-defaults` per @Jefftree's suggestion) as a follow-up PR
- Review the k/k default changes separately in [#138932](https://github.com/kubernetes/kubernetes/pull/138932)

@liggitt, anything else needed here or good to `/lgtm`?

liggitt:
> Sounds like we're aligned on:
> 
> * Land this PR as-is (kin-openapi bump + `structZeroDefault`)
> * `+nodefault` (and possibly `--skip-object-defaults` per @Jefftree's suggestion) as a follow-up PR
> * Review the k/k default changes separately in [#138932](https://github.com/kubernetes/kubernetes/pull/138932)

I agree with those three items, though I'd prefer getting the default-opt-out and k/k opting out in place first so we can keep k/k at kube-openapi HEAD without large unwanted diffs

PR #603, opened 2026-05-07
test/integration: make valuevalidation fields with incompatible const…

arthurbdiniz:
`StringValue` and `NumberValue` in the valuevalidation test fixture have validation constraints (`minLength=1`, `exclusiveMinimum=true`) that their Go zero values (`""` and `0`) cannot satisfy. When these fields are required, the generated schema enforces defaults that immediately violate the constraints.

This PR marks them `// +optional` with `omitempty` JSON tags so the zero value is omitted rather than serialized into an invalid default.


Jefftree:
Thanks!

/lgtm
/approve

Jefftree:
test failure looks related, one of the fixtures probably needs to be updated.

arthurbdiniz:
> test failure looks related, one of the fixtures probably needs to be updated.

ack, checking it

PR #606, opened 2026-05-09
Do not emit defaults for slice items or map value schemas

arthurbdiniz:
Per JSON Schema, the `default` keyword means `"the value to assume if this property is absent."` Array items and map values are never absent in that sense, they are either explicitly provided or the entry does not exist.

Emitting `default` on these schemas is semantically meaningless and adds noise to the generated spec.

Started this at https://github.com/kubernetes/kube-openapi/pull/600#discussion_r3209661718

This PR removes the `generateDefault` calls from `generateSliceProperty` and `generateMapProperty` so defaults are only emitted at the object property level where they are meaningful.

This also significantly reduces the diff footprint when bumping kin-openapi in k/k, since the majority of generated changes were item-level defaults going from `{}` to populated zero values.

### DIFF
--- a/pkg/generators/openapi_test.go
+++ b/pkg/generators/openapi_test.go
@@ -342,7 +342,6 @@ Type: []string{"array"},
 Items: &spec.SchemaOrArray{
 Schema: &spec.Schema{
 SchemaProps: spec.SchemaProps{
-Default: "",
 Type: []string{"string"},
 Format: "",
 },
@@ -364,7 +363,6 @@ AdditionalProperties: &spec.SchemaOrBool{
 Allows: true,
 Schema: &spec.Schema{
 SchemaProps: spec.SchemaProps{
-Default: "",
 Type: []string{"string"},
 Format: "",
 },
@@ -918,7 +916,6 @@ AdditionalProperties: &spec.SchemaOrBool{
 Allows: true,
 Schema: &spec.Schema{
 SchemaProps: spec.SchemaProps{
-Default: "",
 Type: []string{"string"},
 Format: "",
 },
@@ -987,7 +984,6 @@ AdditionalProperties: &spec.SchemaOrBool{
 Allows: true,
 Schema: &spec.Schema{
 SchemaProps: spec.SchemaProps{
-Default: 0,
 Type: []string{"integer"},
 Format: "int32",
 },
@@ -1056,7 +1052,6 @@ AdditionalProperties: &spec.SchemaOrBool{
 Allows: true,
 Schema: &spec.Schema{
 SchemaProps: spec.SchemaProps{
-Default: false,
 Type: []string{"boolean"},
 Format: "",
 },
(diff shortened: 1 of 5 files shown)


Jefftree:
Let's also provide a diff of how the kubernetes/kubernetes repo will change with this PR. Can refer to https://github.com/kubernetes/kubernetes/pull/138895 for reference. Basically with kubernetes/kubernetes checked out, 

```
hack/pin-dependency.sh k8s.io/kube-openapi=github.com/arthurbdiniz/kube-openapi fix/remove-slice-map-item-defaults
hack/update-vendor.sh
hack/update-codegen.sh openapi
hack/update-openapi-spec.sh
```

 and isolate `hack/update-openapi-spec.sh` in a separate commit to see the diff.

pkg/generators/openapi_test.go
```diff
@@ -1158,30 +1153,6 @@ func TestFailingDefaultEnforced(t *testing.T) {
 				}
 			}`,
 		expectedError: `failed to generate default in example.com/base/foo.Blah: Struct: invalid default value (map[string]interface {}{"foo":5}) for non-pointer/non-omitempty. If specified, must be: {}`,
-	}, {
-		definition: `
```
liggitt, line 1162: Is there a spot to move these test cases that still exercises them but expects them to succeed now (without attempting to generate a default for the slice schema, etc)?

liggitt:
lgtm, just the one question on converting the two expect-error test cases to expect-success-without-default ones, and would like to see the dry run of updating kubernetes core repo with this

arthurbdiniz:
> Let's also provide a diff of how the kubernetes/kubernetes repo will change with this PR. Can refer to [kubernetes/kubernetes#138895](https://github.com/kubernetes/kubernetes/pull/138895) for reference. Basically with kubernetes/kubernetes checked out,
> 
> ```
> hack/pin-dependency.sh k8s.io/kube-openapi=github.com/arthurbdiniz/kube-openapi fix/remove-slice-map-item-defaults
> hack/update-vendor.sh
> hack/update-codegen.sh openapi
> hack/update-openapi-spec.sh
> ```
> 
> and isolate `hack/update-openapi-spec.sh` in a separate commit to see the diff.

https://github.com/kubernetes/kubernetes/pull/138932

pkg/generators/openapi_test.go
arthurbdiniz: Yes, I created another test routine to validate that

https://github.com/kubernetes/kube-openapi/pull/606/changes/c8725fb96bc71ad9d56ac26661b9bdf4311f6c99

liggitt:
/lgtm
/approve

kubernetes dry run PR looks good as well

liggitt:
huh, I guess I’m not a kube-openapi maintainer (to run CI) or approver 😆 

once @Jefftree or @jpbetz approve this and it merges, let’s update kubernetes for real to shrink the diff for when we update it with the zero value default PR

liggitt:
k/k update picked this up in https://github.com/kubernetes/kubernetes/pull/139001

PR #612, opened 2026-05-12
Configurable integration test timeout

arthurbdiniz:
Make integration test timeout configurable

On slower platforms (e.g. Debian's RISC-V build machines), the default timeout for Eventually assertions may not be enough to complete the BeforeSuite node.

Debian currently carries a patch to increase this value:

https://salsa.debian.org/go-team/packages/golang-k8s-kube-openapi/-/blob/debian/sid/debian/patches/0004-Increase-integration-tests-timeout-to-1min.patch

Rather than requiring a source patch each time the timeout needs adjusting, this PR makes it configurable at runtime via:

- `-eventually-timeout` test flag: `go test -v . -args -eventually-timeout=120`
- `KUBE_OPENAPI_INTEGRATION_TEST_TIMEOUT` env var: useful for CI and
  distro builds where `-args` scoping to a single package is inconvenient.

The flag takes precedence over the env var. The default remains `60s`.

### DIFF
--- a/test/integration/README.md
+++ b/test/integration/README.md
@@ -8,6 +8,22 @@ Within the current directory:
 $ go test -v .
 ```
 
+### Configuring the test timeout
+
+The integration tests use an `Eventually` timeout for subprocess execution (default: 60s).
+On slower platforms, you can override this with the `-eventually-timeout` flag or the
+`KUBE_OPENAPI_INTEGRATION_TEST_TIMEOUT` environment variable:
+
+```bash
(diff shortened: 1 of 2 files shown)


Jefftree:
This code is a bit weird, timeout isn't really a timeout on the test but rather the subprocess commands. I think a better way to fix this is just to remove the timeout and rely on the timeout set by `go test`

arthurbdiniz:
Make sense @Jefftree, thank for https://github.com/kubernetes/kube-openapi/pull/613, closing this PR.

PR #614, opened 2026-05-19
third_party/go-json-experiment: skip aliases absent from target package

dims:
The alias generator forwarded every exported declaration from the vendored go-json-experiment shim to the target stdlib package (encoding/json/v2). When stdlib v2 dropped symbols that still exist in the pinned upstream snapshot - SkipFunc and DiscardUnknownMembers after JSONv2 graduated to baseline in golang/go@e62d3e6e - the generated alias.go referenced undefined symbols and broke every consumer transitively importing kube-openapi.

- https://github.com/go-json-experiment/json/pull/201
- https://github.com/go-json-experiment/json/pull/203 

Teach alias_gen.go to load the target package's exported top-level declarations via go/build and go/parser, and emit aliases only for names present there. The check is applied to const, var, type, and non-method function declarations.

Regenerate alias.go for encoding/json/v2; SkipFunc and DiscardUnknownMembers are dropped. Builds clean under both GOEXPERIMENT=jsonv2 and GOEXPERIMENT=nojsonv2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,28 @@ jobs:
     - name: Run verify scripts
       run: |
         ./hack/verify-go-directive.sh 1.24.0
+  ci-go-tip:
+    # Informational lane that exercises the vendored encoding/json/v2 shim
+    # against Go tip so upstream changes (e.g. graduating JSONv2 to baseline)
+    # surface here rather than in kubernetes/kubernetes CI.
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.0
+    - name: Set up Go stable
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.0.0
+      with:
+        go-version: 'stable'
+    - name: Install gotip
+      run: |
+        go install golang.org/dl/gotip@latest
+        gotip download
+    - name: Build with gotip (default GOEXPERIMENT)
+      run: |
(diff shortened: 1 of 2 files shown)


Jefftree:
/cc @liggitt
ref: https://github.com/kubernetes/kubernetes/issues/139159

liggitt:
this subtree is imported as is from json-experimental ... any changes to it will be lost when we rerun the import script

I don't think we need to update the generator, just delete the SkipFunc and DiscardUnknownMembers bits. The next time we update json-experimental, we'll update to HEAD and it will be compatible with json/v2 stdlib

It would probably also be good to add a ci-pipline variant for go tip here as long as we have this json/v2 dep

pkg/internal/third_party/go-json-experiment/json/alias_gen.go
liggitt, line 1: I'd revert the changes here and just manually commit the alias.go change

dims: Done... i think :)

.github/workflows/ci.yml
```diff
@@ -38,6 +38,28 @@ jobs:
     - name: Run verify scripts
       run: |
         ./hack/verify-go-directive.sh 1.24.0
+  ci-go-tip:
```
Jefftree, line 41: Thanks! Making a note that we'll add a periodic test for this as well since it's only presubmit at the moment.

liggitt:
thanks, @Jefftree, can you open a bump PR in k/k once this merges?

Jefftree:
> thanks, @Jefftree, can you open a bump PR in k/k once this merges?

👍

PR #616, opened 2026-06-18
Insulate go-json-experiment from stdlib

liggitt:
Drop aliases to json/v1 and json/v2 stdlib packages, we want our import of go-json-experiment to be self-contained until we actually collapse onto json/v2 (we'll need to switch `inline` to `embed` at that point, for example)

Otherwise, the behavior automatically switches to calling into stdlib in Go 1.27, which we don't want

Resolves failures like https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-unit-golang-tip/2067669795943223296 due to changing json/v2 prerelease behavior

cc @Jefftree @dims 

```release-note
NONE
```


jpbetz (commented):
/approve
/lgtm

PR #619, opened 2026-06-26
validation: anchor duration regex to reject invalid tokens

bernot-dev:
I anchored the duration validation regex to ensure the entire string is validated.

Previously, the durationMatcher regex was unanchored. This allowed ParseDuration to match valid duration substrings embedded within invalid strings, ignoring the invalid text. For example, "1s invalid" would successfully validate and return a duration of 1s.

To fix this, I introduced a two-step validation:

1. Check the candidate string against durationValidator, an anchored regex that ensures the input consists only of valid duration components.
2. Sum the components using durationMatcher.

This allows composite durations like "1w2s" or "1w 2d 3h 4m" to be parsed while rejecting inputs with invalid tokens. I added positive and negative test cases to cover expected behavior.

### DIFF
--- a/pkg/validation/strfmt/duration.go
+++ b/pkg/validation/strfmt/duration.go
@@ -52,7 +52,8 @@ var (
 		"w":  7 * 24 * time.Hour,
 	}
 
-	durationMatcher = regexp.MustCompile(`((\d+)\s*([A-Za-zµ]+))`)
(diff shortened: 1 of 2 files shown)


bernot-dev:
Closing this in favor of #621, which is a much more direct backport of the upstream go-openapi library.

PR #620, opened 2026-06-27
Use AllowMultiple as signal for array query parameters

rithwik-01:
Use `AllowMultiple()` as the signal for emitting `type: array` in OpenAPI schemas instead of detecting a `[]` prefix on DataType. When `AllowMultiple()` is true, the `[]` prefix is only stripped to derive the element type.

Changes:
- `pkg/builder/openapi.go` (v2): array detection via `AllowMultiple()`, `CollectionFormat: "multi"` always set for array params
- `pkg/builder3/openapi.go` (v3): same, with `Style: "form"` + `Explode: true` for array params
- `[]byte` stays scalar naturally through `OpenAPITypeFormat`: no special guard needed
- `[]string` without `AllowMultiple` intentionally errors as a misconfiguration
- Tests: table-driven, uses `restfuladapter.ParamAdapter` (same adapter as runtime)

### Relationship to kubernetes/kubernetes#140050

This PR is a **necessary precondition** for fixing that issue, but does not close it alone. The Kubernetes apiserver's `typeToJSON` function (`staging/src/k8s.io/apiserver/pkg/endpoints/installer.go`) explicitly strips `[]string` → `"string"` and `[]int32` → `"integer"` with a TODO referencing the go-restful limitation this PR addresses. A companion PR to remove those cases from `typeToJSON` is needed to fully resolve the issue.

Refs: kubernetes/kubernetes#140050

### DIFF
--- a/pkg/builder/openapi.go
+++ b/pkg/builder/openapi.go
@@ -448,32 +448,34 @@ func (o *openAPI) buildParameter(restParam common.Parameter, bodySample interfac
 		return ret, fmt.Errorf("unknown restful operation kind : %v", restParam.Kind())
 	}
 	dataType := restParam.DataType()
-	var openAPIType, openAPIFormat string
-	openAPIType, openAPIFormat = common.OpenAPITypeFormat(dataType)
-	if openAPIType == "" && strings.HasPrefix(dataType, "[]") {
-		// Array type: element type is encoded as "[]string" or "[]integer"
-		itemsType := dataType[2:]
-		itemsAPIType, itemsAPIFormat := common.OpenAPITypeFormat(itemsType)
-		if itemsAPIType == "" {
-			return ret, fmt.Errorf("non-body Restful parameter array element type should be a simple type, but got: %v", itemsType)
-		}
-		openAPIType = "array"
-		openAPIFormat = ""
+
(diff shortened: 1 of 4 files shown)


rithwik-01:
@jpbetz / @sttts Can you please review this PR? Happy to provide more context or make changes if needed.

Jefftree (commented):
See if you can align to use `AllowMultiple` instead.

rithwik-01:
@Jefftree Thank you for the suggestion.
  Updated to use AllowMultiple() as the sole signal for array detection:

  - AllowMultiple()=true → emits type: array with items; [] prefix is only stripped to derive the element type, not for detection
  - []byte stays scalar naturally through OpenAPITypeFormat — no special guard needed
  - []string without AllowMultiple intentionally errors as a misconfiguration (returns to pre-PR behavior rather than silently masking it)
  - V2: CollectionFormat: "multi" always set for AllowMultiple params
  - V3: Style: "form" + Explode: true always set for AllowMultiple params

rithwik-01:
> See if you can align to use `AllowMultiple` instead.

@Jefftree I have made the changes, can you please review the PR? Happy to provide more context on the changes or make changes if needed.

PR #621, opened 2026-06-29
fix: update strfmt duration parsing

bernot-dev:
Update pkg/validation/strfmt/duration.go to backport the duration parsing improvements from upstream github.com/go-openapi/strfmt@v0.26.4.

Specifically, this makes validation more strict and performant. See: https://github.com/go-openapi/strfmt/pull/246

Unrecognized tokens that do not correspond to a known duration will cause the format to be deemed invalid.

### DIFF
--- a/pkg/validation/strfmt/duration.go
+++ b/pkg/validation/strfmt/duration.go
@@ -38,7 +38,7 @@ const (
 	hours      = uint64(time.Hour)
 	days       = uint64(hoursInDay * time.Hour)
 	weeks      = uint64(hoursInDay * daysInWeek * time.Hour)
-	maxUint64  = uint64(1 << 63)
+	maxDurationMagnitude = uint64(1 << 63)
 )
 
 // timeMultiplier holds all supported aliases for duration units, including their plural form.
@@ -52,7 +52,7 @@ var timeMultiplier = map[string]uint64{
 	"nanos":        nanos,
 	"us":           micros,
 	"µs":           micros, // U+00B5 = micro symbol
-	"μs":           micros, // U+03BC = Greek letter mu
+	// "μs":        micros, // U+03BC = Greek letter mu (rejected, new in upstream but not accepted here)
 	"micro":        micros,
 	"micros":       micros,
 	"microsecond":  micros,
@@ -74,7 +74,7 @@ var timeMultiplier = map[string]uint64{
 	"minutes":      minutes,
 	"h":            hours,
 	"hr":           hours,
-	"hrs":          hours,
+	// "hrs":       hours, (rejected, new in upstream but not accepted here)
 	"hour":         hours,
 	"hours":        hours,
 	"d":            days,
@@ -82,7 +82,7 @@ var timeMultiplier = map[string]uint64{
 	"days":         days,
 	"w":            weeks,
 	"wk":           weeks,
-	"wks":          weeks,
+	// "wks":       weeks, (rejected, new in upstream but not accepted here)
 	"week":         weeks,
 	"weeks":        weeks,
 }
@@ -118,8 +118,8 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
 
 // ParseDuration parses a duration from a string
 //
-// It is similar to [time.ParseDuration] but support additional units like days and weeks,
-// additional abreviations for units and is more tolerant on the presence of blank spaces.
+// It is similar to [time.ParseDuration] but supports additional units,
+// additional abbreviations for units and is more tolerant on the presence of blank spaces.
 //
 // A duration may be negative or fractional.
 //
@@ -134,15 +134,13 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
 // Units may be specified using aliases or a plural form.
 //
 //   - "ns", "nano", "nanosecond", "nanoseconds", "nanos"
-//   - "us", "µs" (U+00B5 = micro symbol), "μs" (U+03BC = Greek letter mu), "micro", "micros", "microsecond", "microseconds"
+//   - "us", "µs" (U+00B5 = micro symbol), "micro", "micros", "microsecond", "microseconds"
 //   - "ms", "milli", "millis", "millisecond", "milliseconds"
 //   - "s", "sec", "secs", "second", "seconds"
 //   - "m", "min", "mins", "minute", "minutes"
-//   - "h", "hr", "hrs", "hour", "hours"
+//   - "h", "hr", "hour", "hours"
 //   - "d", "day", "days"
-//   - "w", "wk", "wks", "week", "weeks"
-//
-// NOTE: inspired by scala duration syntax.
+//   - "w", "wk", "week", "weeks"
 //
 // # Examples
 //
@@ -154,6 +152,10 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
 func ParseDuration(s string) (time.Duration, error) {
 	// NOTE: this code is largely inspired by the standard library.
 	orig := s
+	s = strings.TrimSpace(s)
+	if dur, err := time.ParseDuration(s); err == nil {
+		return dur, nil
+	}
 	var d uint64
 	neg := false
 
@@ -236,7 +238,7 @@ func ParseDuration(s string) (time.Duration, error) {
 			return 0, parseDurationError(orig, fmt.Sprintf("unknown unit %q in duration", u))
 		}
 
-		if v > maxUint64/unit {
+		if v > maxDurationMagnitude/unit {
 			// overflow
 			return 0, parseDurationError(orig, "numerical overflow")
 		}
@@ -246,14 +248,14 @@ func ParseDuration(s string) (time.Duration, error) {
 			// float64 is needed to be nanosecond accurate for fractions of hours.
 			// v >= 0 && (f*unit/scale) <= 3.6e+12 (ns/h, h is the largest unit)
 			v += uint64(float64(f) * (float64(unit) / scale))
-			if v > maxUint64 {
+			if v > maxDurationMagnitude {
 				// overflow
 				return 0, parseDurationError(orig, "numerical overflow")
 			}
 		}
 
 		d += v
-		if d > maxUint64 {
+		if d > maxDurationMagnitude {
 			return 0, parseDurationError(orig, "numerical overflow")
 		}
 	}
@@ -262,7 +264,7 @@ func ParseDuration(s string) (time.Duration, error) {
 		return -time.Duration(d), nil
 	}
 
-	if d > maxUint64-1 {
+	if d > maxDurationMagnitude-1 {
 		return 0, parseDurationError(orig, "numerical overflow")
 	}
 
@@ -329,12 +331,12 @@ func leadingInt[bytes []byte | string](s bytes) (x uint64, rem bytes, ok bool) {
 			break
 		}
 
-		if x > maxUint64/10 { // overflow
+		if x > maxDurationMagnitude/10 { // overflow
 			return 0, rem, false
 		}
 
 		x = x*10 + uint64(c) - '0'
-		if x > maxUint64 { // overflow
+		if x > maxDurationMagnitude { // overflow
 			return 0, rem, false
 		}
 	}
@@ -360,14 +362,14 @@ func leadingFraction(s string) (x uint64, scale float64, rem string) {
 			continue
 		}
 
-		if x > (maxUint64-1)/10 {
+		if x > (maxDurationMagnitude-1)/10 {
 			// It's possible for overflow to give a positive number, so take care.
 			overflow = true
 			continue
 		}
 
 		y := x*10 + uint64(c) - '0'
-		if y > maxUint64 {
+		if y > maxDurationMagnitude {
 			overflow = true
 			continue
 		}
--- a/pkg/validation/strfmt/duration.go
+++ b/pkg/validation/strfmt/duration.go
@@ -17,45 +17,77 @@ package strfmt
 import (
 	"encoding/json"
 	"fmt"
-	"regexp"
-	"strconv"
 	"strings"
 	"time"
+	"unicode"
 )
 
-func init() {
+func init() { //nolint:gochecknoinits // registers duration format in the default registry
 	d := Duration(0)
-	// register this format in the default registry
 	Default.Add("duration", &d, IsDuration)
 }
 
-var (
-	timeUnits = [][]string{
-		{"ns", "nano"},
-		{"us", "µs", "micro"},
-		{"ms", "milli"},
-		{"s", "sec"},
-		{"m", "min"},
-		{"h", "hr", "hour"},
-		{"d", "day"},
-		{"w", "wk", "week"},
-	}
-
-	timeMultiplier = map[string]time.Duration{
-		"ns": time.Nanosecond,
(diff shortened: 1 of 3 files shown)


Jefftree:
LGTM

/cc @liggitt

Didn't get time to finish my previous attempt in #552, this should address the concerns you raised there. It backports go-openapi's parser, so the junk-skipping is gone. One note is that it still intentionally tolerates a space after the sign (`- 5s`), which I think should be fine?

pkg/validation/strfmt/duration.go
```diff
@@ -17,45 +17,77 @@ package strfmt
...
-// IsDuration returns true if the provided string is a valid duration
+// timeMultiplier holds all supported aliases for duration units, including their plural form.
+//
+//nolint:gochecknoglobals // package-level lookup tables for duration parsing
+var timeMultiplier = map[string]uint64{
+	"ns":           nanos,
+	"nano":         nanos,
+	"nanosecond":   nanos,
+	"nanoseconds":  nanos,
+	"nanos":        nanos,
```
liggitt, line 52: Anything *new* we accept is something that will be considered invalid by previous releases, so let's not expand the surface area of units we accept without further rationale.

(comment applies to all ... comment out the ones we don't currently accept in this list)

pkg/validation/strfmt/duration.go
```diff
@@ -17,45 +17,77 @@ package strfmt
...
+	daysInWeek = 7
+	nanos      = uint64(time.Nanosecond)
+	micros     = uint64(time.Microsecond)
+	millis     = uint64(time.Millisecond)
+	seconds    = uint64(time.Second)
+	minutes    = uint64(time.Minute)
+	hours      = uint64(time.Hour)
+	days       = uint64(hoursInDay * time.Hour)
+	weeks      = uint64(hoursInDay * daysInWeek * time.Hour)
+	maxUint64  = uint64(1 << 63)
```
liggitt: this is a confusing constant name ... 1 << 63 is *not* the max uint64 value ... what is this used for?

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
+// Units may be specified using aliases or a plural form.
+//
+//   - "ns", "nano", "nanosecond", "nanoseconds", "nanos"
+//   - "us", "µs" (U+00B5 = micro symbol), "μs" (U+03BC = Greek letter mu), "micro", "micros", "microsecond", "microseconds"
+//   - "ms", "milli", "millis", "millisecond", "milliseconds"
+//   - "s", "sec", "secs", "second", "seconds"
+//   - "m", "min", "mins", "minute", "minutes"
+//   - "h", "hr", "hrs", "hour", "hours"
+//   - "d", "day", "days"
+//   - "w", "wk", "wks", "week", "weeks"
```
liggitt: scope these down to the unit formats we already accepted

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
 	return nil
 }
 
-// ParseDuration parses a duration from a string, compatible with scala duration syntax
-func ParseDuration(cand string) (time.Duration, error) {
-	if dur, err := time.ParseDuration(cand); err == nil {
-		return dur, nil
+// ParseDuration parses a duration from a string
+//
+// It is similar to [time.ParseDuration] but support additional units like days and weeks,
```
liggitt:
```suggestion
// It is similar to [time.ParseDuration] but support additional units,
```

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
+//   - "ns", "nano", "nanosecond", "nanoseconds", "nanos"
+//   - "us", "µs" (U+00B5 = micro symbol), "μs" (U+03BC = Greek letter mu), "micro", "micros", "microsecond", "microseconds"
+//   - "ms", "milli", "millis", "millisecond", "milliseconds"
+//   - "s", "sec", "secs", "second", "seconds"
+//   - "m", "min", "mins", "minute", "minutes"
+//   - "h", "hr", "hrs", "hour", "hours"
+//   - "d", "day", "days"
+//   - "w", "wk", "wks", "week", "weeks"
+//
+// NOTE: inspired by scala duration syntax.
```
liggitt: can drop the scala reference, I don't think that's particularly relevant at this point

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
+
+	for s != "" {
+		var (
+			v, f  uint64      // integers before, after decimal point
+			scale float64 = 1 // value = v + f/scale
+		)
+		s = strings.TrimLeftFunc(s, unicode.IsSpace)
+
+		// The next character must be 0-9.]
+		if s[0] != '.' && ('0' > s[0] || s[0] > '9') {
```
liggitt, line 191: accessing `s[0]` after trimming will panic if s was entirely whitespace, right? I don't see this trimming trailing spaces in the loop, so I think that's a possibility if the value had trailing spaces. I think this is a bug upstream as well. I think `s = strings.TrimLeftFunc(s, unicode.IsSpace)` is required *after* consuming the unit (or trim leading *and* trailing space before entering the loop on line 170)

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
-// ParseDuration parses a duration from a string, compatible with scala duration syntax
-func ParseDuration(cand string) (time.Duration, error) {
-	if dur, err := time.ParseDuration(cand); err == nil {
-		return dur, nil
+// ParseDuration parses a duration from a string
+//
+// It is similar to [time.ParseDuration] but support additional units like days and weeks,
+// additional abreviations for units and is more tolerant on the presence of blank spaces.
+//
+// A duration may be negative or fractional.
```
liggitt, line 124: paging in context... `strfmt.ParseDuration` didn't correctly extract negative or fractional values on multi-unit values before, but it also didn't error on them? so now we're getting *stricter* about malformed multi-unit values but still accepting negative and fractional values, and now parsing them correctly?

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,170 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
 	return nil
 }
 
-// ParseDuration parses a duration from a string, compatible with scala duration syntax
-func ParseDuration(cand string) (time.Duration, error) {
-	if dur, err := time.ParseDuration(cand); err == nil {
```
liggitt, line 89: if there anything `time.ParseDuration` accepted that the new code no longer accepts?

is there a reason not to keep this as a first step?

pkg/validation/strfmt/duration.go
bernot-dev: I first noticed kube-openapi accepting non-duration junk in the duration string, which feels pretty incorrect to me. This is a more comprehensive breakdown of what is and is not accepted in each library:

| Rule / Feature | Example Duration | Go stdlib (`time.ParseDuration`) | Kube-OpenAPI (release-1.22) | Go-OpenAPI (v0.26.4) / Current PR |
| :--- | :--- | :---: | :---: | :---: |
| **Standard duration format** | `"1h30m"` | ✅ (`1h30m0s`) | ✅ (`1h30m0s`) | ✅ (`1h30m0s`) |
| **Custom units (`d`, `w`)** | `"2d"`, `"1w"` | ❌ | ✅ (`48h`, `168h`) | ✅ (`48h`, `168h`) |
| **Spaces between value & unit** | `"300 ms"` | ❌ | ✅ (`300ms`) | ✅ (`300ms`) |
| **Spaces between components** | `"1h 45m"` | ❌ | ✅ (`1h45m0s`) | ✅ (`1h45m0s`) |
| **Space after sign** | `"- 5s"` | ❌ | ✅ (`5s` ⚠️ **Positive**) | ✅ (`-5s` **Negative**) |
| **Negative custom units** | `"-1d"` | ❌ | ✅ (`24h` ⚠️ **Positive**) | ✅ (`-24h` **Negative**) |
| **Decimal custom units** | `"1.5d"` | ❌ | ✅ (`120h` ⚠️ **Parsed as 5d**) | ✅ (`36h` **Parsed as 1.5d**) |
| **Non-duration junk in string** | `"10s junk 5s"` | ❌ | ✅ (`15s` ⚠️ **Junk ignored**) | ❌ (Rejected) |
| **Suffix-matched typos** | `"45 weekz"` | ❌ | ✅ (`7560h` ⚠️ **Parsed as 45w**) | ❌ (Rejected) |
| **New unit abbreviations** | `"1 hrs"`, `"1 wks"`, `"1 μs"` | ❌ | ❌ | ❌ (Rejected) |

pkg/validation/strfmt/duration.go
bernot-dev: Accepted units are now aligned.

| Unit Family | Unit / Suffix | Go stdlib (`time.ParseDuration`) | Kube-OpenAPI (release-1.22) | Current PR |
| :--- | :--- | :---: | :---: | :---: |
| **Nanoseconds** | `ns` | ✅ | ✅ | ✅ |
| | `nano`, `nanos`, `nanosecond`, `nanoseconds` | ❌ | ✅ | ✅ |
| **Microseconds** | `us`, `µs` (micro symbol `U+00B5`) | ✅ | ✅ | ✅ |
| | `μs` (Greek letter mu `U+03BC`) | ✅ | ✅ (via stdlib fast-path) | ✅ (via stdlib fast-path) |
| | `micro`, `micros`, `microsecond`, `microseconds` | ❌ | ✅ | ✅ |
| **Milliseconds** | `ms` | ✅ | ✅ | ✅ |
| | `milli`, `millis`, `millisecond`, `milliseconds` | ❌ | ✅ | ✅ |
| **Seconds** | `s` | ✅ | ✅ | ✅ |
| | `sec`, `secs`, `second`, `seconds` | ❌ | ✅ | ✅ |
| **Minutes** | `m` | ✅ | ✅ | ✅ |
| | `min`, `mins`, `minute`, `minutes` | ❌ | ✅ | ✅ |
| **Hours** | `h` | ✅ | ✅ | ✅ |
| | `hr`, `hour`, `hours` | ❌ | ✅ | ✅ |
| | `hrs` | ❌ | ❌ | ❌ |
| **Days** | `d`, `day`, `days` | ❌ | ✅ | ✅ |
| **Weeks** | `w`, `wk`, `week`, `weeks` | ❌ | ✅ | ✅ |
| | `wks` | ❌ | ❌ | ❌ |

pkg/validation/strfmt/duration.go
liggitt: (want to contribute this fix upstream as well?)

pkg/validation/strfmt/duration.go
bernot-dev: This is how go-openapi is detecting overflow in the newer version of their library. I renamed it to `maxDurationMagnitude`, which is somewhat more descriptive.

pkg/validation/strfmt/duration.go
bernot-dev: Fixed.

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,172 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
 			}
 		}
+
+		if i == 0 {
+			return 0, parseDurationError(orig, "missing unit in duration")
+		}
+
+		u := s[:i]
+		s = s[i:]
+		unit, ok := timeMultiplier[u]
```
liggitt, line 236: should we look this up via lowercasing u to preserve accepting case-insensitive units?

pkg/validation/strfmt/duration.go
```diff
@@ -84,52 +116,172 @@ func (d *Duration) UnmarshalText(data []byte) error { // validation is performed
...
-		for _, variants := range timeUnits {
-			last := len(variants) - 1
-			multiplier := timeMultiplier[variants[0]]
+		if !pre && !post {
+			// no digits (e.g. ".s" or "-.s")
+			return 0, parseDurationError(orig, "expected digits")
+		}
 
-			for i, variant := range variants {
-				if (last == i && strings.HasPrefix(unit, variant)) || strings.EqualFold(variant, unit) {
```
liggitt, line 108: Thinking through this... previously we accepted any prefix or case-insensitive match of a variant

Prefix matching allowed all the pluralizations and full spellings (nano → nanos, nanosecond, nanoseconds, etc). Tightening to an explicit list of variants makes sense to me (so we allow `nanoseconds` but not `nanobananas`, etc).

Is there a good reason to drop the case-insensitive matching? `1H 30M` would previously have been accepted and parsed successfully, right?

pkg/validation/strfmt/duration.go
liggitt: I guess the case-insensitive matching was only for non-prefix matches. To preserve that behavior without expanding what we accept, we'd need a distinct `caseInsensitiveTimeMultiplier` map that only included what used to be in `timeUnits`.

Then we can get our multiplier either by looking up an exact match in `timeMultiplier`, then fall back to `caseInsensitiveTimeMultiplier` with the lowercased unit.

pkg/validation/strfmt/duration.go
bernot-dev: Actually, ISO-8601 was accidentally supported before. Even though it would not have been parsed correctly, it would pass validation.

I don't know if that's a good argument for true case-insensitive matching, but I think ISO 8601 should probably be supported.

bernot-dev: Opened https://github.com/go-openapi/strfmt/issues/278.

pkg/validation/strfmt/duration.go
bernot-dev: https://github.com/kubernetes/kube-openapi/pull/621#discussion_r3559861719

pkg/validation/strfmt/duration.go
bernot-dev: Upstream: https://github.com/go-openapi/strfmt/pull/279

pkg/validation/strfmt/duration.go
liggitt: I don't think ISO 8601 is *possible* to support and parse to concrete time.Durations with ambiguous things like year and month

bernot-dev: Yeah, I think there's basically 3 reasonable paths forward.
1. Don't support ISO 8601 at all.
2. Normalize years/months to 365 days and 30 days, respectively.
3. Support only the time portion of the standard.

Technically, days are also ambiguous when you consider daylight savings time. But hours, minutes, and seconds should work. Unless you deal with leap seconds...

I don't have a strong opinion here, but causing ISO 8601 to newly fail validation seems risky at scale.

bernot-dev: Go-openapi decided to go with option 2.

https://github.com/go-openapi/strfmt/blob/master/duration_iso8601.go#L18-L20

